Issue #95, opened 2026-03-28
Security: Broker catch-all proxies unregistered operations with credentials, bypassing RBAC policies


rodaddy:
## Summary

The broker catch-all route (`/{target}`) proxies requests to upstream APIs and injects credentials even when the requested operation is **not registered** in the imported OpenAPI spec. Since RBAC policy rules match against Capability IDs (`METHOD/host/path`), unregistered operations bypass policy evaluation entirely.

## Reproduction

1. Import a minimal spec for `api.github.com` with only `GET /user` and `GET /user/repos`
2. Create a toolkit with a write-deny policy:
   ```json
   [
     {"effect": "allow", "pattern": "GET/api.github.com/*"},
     {"effect": "deny", "pattern": "*/api.github.com/*"}
   ]
   ```
3. Bind a GitHub PAT credential to the toolkit
4. Using the toolkit's agent key, send `POST /api.github.com/user/repos` through the broker
5. **Expected:** Request blocked by the deny-all or system default write-deny policy
6. **Actual:** Request proxied with credentials -- successfully created a repo on GitHub

## Impact

An agent with a scoped toolkit key can perform **any** operation on the upstream API, as long as that operation wasn't explicitly registered in the spec. The system default write-deny policy (`_system: true`) didn't fire on the catch-all path for the unregistered operation.

This undermines the core trust model -- an admin who sets up read-only access for an agent can't guarantee the agent won't perform writes on unregistered endpoints.

## Suggested Fix

Option A: Default-deny on unregistered operations -- if the broker can't match the request to a registered Capability ID, reject it.

Option B: Apply policy rules against the raw `METHOD/host/path` even when no registered operation exists, so glob patterns still match.

Option C: Add a per-toolkit config flag (`deny_unregistered: true`) so admins can opt in to strict mode.

---

*Found during hands-on POC testing by [@rodaddy](https://github.com/rodaddy) (PAI Infrastructure). We forked jentic-mini, stood it up in Docker, imported a minimal GitHub spec, and tested RBAC boundaries -- this was discovered when a scoped toolkit key successfully created a repo on GitHub that should have been blocked by write-deny policies.*

PR #100, opened 2026-03-28
Security: fail closed on credential resolution and policy check errors

MichaelCordner:
## Summary

Closes the two exception-bypass paths identified in #95 (phase 1). All three `except Exception: pass` blocks on the credential resolution and policy evaluation paths now return denial responses instead of silently continuing.

**What changed:**
- **Credential resolution failure** (broker.py:580) — if `_resolve_credential_ids` throws, return 500 instead of skipping the policy gate and proceeding to credential injection
- **Policy check failure** (broker.py:610) — if `check_credential_policy` throws, return 403 instead of injecting credentials unchecked
- **Pipedream policy check failure** (broker.py:714) — same fix for the Pipedream credential path

**Principle: credentials are never injected when the policy gate couldn't run.**

This does not address the broader design gap (pattern-matching rules can match unregistered operations) — that's tracked as phase 2 in #95.

## Test plan
- [x] Normal broker flow still works (credential injection + policy check on registered operations)
- [x] Verify that a DB error during credential resolution returns 500 with `CREDENTIAL_RESOLUTION_FAILED`
- [x] Verify that a malformed policy rule returns 403 with `POLICY_CHECK_FAILED` instead of passing through
- [x] Pipedream OAuth credential path still works when policy check succeeds

### DIFF
--- a/src/routers/broker.py
+++ b/src/routers/broker.py
@@ -583,17 +583,19 @@ async def broker(request: Request, target: str):
             )
         except Exception:
             pass
-    else:
+    elif toolkit_id:
         try:
             _api_id_for_host, _resolved_cred_ids = await _resolve_credential_ids(
                 host=upstream_host, toolkit_id=toolkit_id
             )
-        except Exception as _resolve_exc:
+        except Exception:
             # Fail closed: if we can't resolve credentials for policy checking,
             # don't proceed to credential injection — deny the request.
-            log.error("Credential resolution failed for %s (toolkit=%s): %s",
-                              upstream_host, toolkit_id, _resolve_exc)
-            await _write_trace("error", 500, f"Credential resolution failed: {_resolve_exc}")
+            # Only applies to authenticated requests; anonymous passthrough
+            # skips credential resolution entirely.
+            log.exception("Credential resolution failed for %r (toolkit=%s)",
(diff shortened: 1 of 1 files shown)


char0n:
## Local test results

Set up a test credential (`github-test`, bearer token for `api.github.com`) bound to the default toolkit, then corrupted its policy rule (invalid JSON in `credential_policies.rules`) to trigger `check_credential_policy` exception.

| Test | Scenario | Expected | Actual |
|------|----------|----------|--------|
| A | Authenticated + valid credentials | Credential injected, upstream responds | **401** — GitHub says "Bad credentials" (fake token, but credential WAS injected) |
| B | Anonymous passthrough (no key) | Forward without credentials | **200** — GitHub rate_limit endpoint returns data |
| C | Authenticated + corrupted policy | Fail-closed, deny request | **403 POLICY_CHECK_FAILED** — credentials NOT injected |
| D | Anonymous + corrupted policy in DB | Passthrough unaffected | **200** — anonymous skips credential resolution entirely |

Trace for test C:
```json
{
  "status": "error",
  "http_status": 403,
  "error": "Policy check failed for GET /rate_limit (credential github-test)"
}
```

No raw exception messages in traces. Full stack trace visible in server logs only via `log.exception()`.

PR #101, opened 2026-03-28
fix: strip reverse-proxy headers before forwarding to upstream APIs

alfredo-xyz:
## Problem

When Jentic Mini runs behind a reverse proxy (nginx ingress, Traefik, etc.), inbound requests arrive with headers describing the proxy hop — `x-forwarded-host`, `x-forwarded-for`, `x-real-ip`, and similar. The broker was passing these headers through verbatim to upstream APIs.

This causes upstream failures. For example, TMDB's CloudFront CDN returns `403 Host not permitted` when it receives `x-forwarded-host: <jentic-hostname>` instead of `api.themoviedb.org`. Other APIs with similar host validation have the same failure mode.

## Fix

Adds the following headers to the existing `_HOP_BY_HOP` strip-set in `src/routers/broker.py`:

```
x-forwarded-for, x-forwarded-host, x-forwarded-port,
x-forwarded-proto, x-forwarded-scheme,
x-real-ip, x-request-id, x-scheme
```

These headers describe the inbound hop to Jentic Mini — they are meaningless (and harmful) when forwarded to an unrelated upstream API.

## Testing

Tested in a local network where Jentic Mini is deployed in a Kubernetes cluster behind an nginx ingress controller. The agent connecting to it runs outside the cluster on the same network and reaches Jentic via the reverse proxy. With this fix, broker requests to upstream APIs (including TMDB/CloudFront) succeed where they previously returned 403.

### DIFF
--- a/src/routers/broker.py
+++ b/src/routers/broker.py
@@ -70,7 +70,7 @@ _HOP_BY_HOP = {
     # hostname instead of the upstream API hostname.
     "x-forwarded-for", "x-forwarded-host", "x-forwarded-port",
     "x-forwarded-proto", "x-forwarded-scheme",
-    "x-real-ip", "x-request-id", "x-scheme",
+    "x-real-ip", "x-scheme",
 }
 
 # How we detect credentials for a given API host


char0n:
One consideration
                                                                                                               
  `x-request-id` stripping — Some upstream APIs use x-request-id for request correlation/deduplication. Stripping
   it means Jentic won't forward an inbound request ID to the upstream. This is probably fine (the broker      
  generates its own X-Jentic-Execution-Id for tracing), but worth noting that if an agent sets x-request-id
  intentionally for an upstream API, it will be silently dropped.

src/routers/broker.py
```diff
@@ -63,6 +63,14 @@
...
     # Host is set from the target URL
     "host",
+    # Reverse-proxy headers injected by nginx/traefik/etc. — these describe
+    # the inbound hop to Jentic, not the outbound hop to the upstream API.
+    # Forwarding them causes failures: e.g. CloudFront returns 403
+    # "Host not permitted" when it sees x-forwarded-host with the Jentic
+    # hostname instead of the upstream API hostname.
+    "x-forwarded-for", "x-forwarded-host", "x-forwarded-port",
+    "x-forwarded-proto", "x-forwarded-scheme",
+    "x-real-ip", "x-request-id", "x-scheme",
```
char0n:
```suggestion
    "x-real-ip", "x-scheme",
```

PR #103, opened 2026-03-29
feat: version display + one-click upgrades via Watchtower

MichaelCordner:
## Summary

Adds version visibility in the UI and a one-click upgrade path so users never have to SSH in to update.

### What changed

**UI sidebar footer** — always shows the current version (e.g. `v0.2.5`).

**Update banner** — when a newer release exists on GitHub:
- If Watchtower is configured: shows an **"Update now" button** that triggers the upgrade in one click
- If Watchtower is not configured: shows a **"View release notes"** link to the GitHub release page

**`POST /admin/upgrade` endpoint** — human session required. Calls Watchtower's HTTP API to trigger an image pull + container restart. Returns immediately; the app briefly goes offline and comes back on the new version. Agents cannot call this endpoint (403 for non-human sessions).

**Watchtower sidecar in `compose.yml`** — runs in **monitor-only mode** by default:
- `WATCHTOWER_MONITOR_ONLY=true` — checks for new images daily but never applies them automatically
- `WATCHTOWER_HTTP_API_UPDATE=true` — exposes an internal HTTP endpoint for on-demand updates
- `WATCHTOWER_SCOPE=jentic` — scoped to only the jentic-mini container (won't touch other containers on the host)
- No ports exposed to the host — only reachable from jentic-mini via the Docker network
- Token-authenticated (`WATCHTOWER_TOKEN` env var, defaults to `jentic-update`)

**`/version` endpoint** — now returns `watchtower_configured: true/false` so the UI knows whether to show the button or the fallback link.

**`jentic-update.sh`** — host-side fallback script for users who don't use the compose file with Watchtower: `docker compose pull && docker compose up -d`.

**README** — new "Updating" section documenting all three upgrade paths (UI button, script, manual commands).

### Files changed

| File | Change |
|------|--------|
| `ui/src/components/layout/Layout.tsx` | Version display in sidebar footer + update banner with button |
| `ui/src/hooks/useUpdateCheck.ts` | Added `upgradeAvailable` to `UpdateStatus` interface |
| `compose.yml` | Watchtower sidecar service + scope labels + env vars on jentic-mini |
| `src/main.py` | `POST /admin/upgrade` endpoint + `upgrade_available` field on `/version` |
| `jentic-update.sh` | New file — host-side manual upgrade script |
| `README.md` | New "Updating" section |

### Security notes

- `/admin/upgrade` requires a human session (`is_human_session` check) — agents get 403
- Watchtower's HTTP API is token-authenticated and only reachable on the internal Docker network
- Docker socket is mounted only into the Watchtower container, not into jentic-mini
- Watchtower is scoped — it only monitors/updates the jentic-mini container

### How to test

1. `docker compose up -d` — both jentic-mini and watchtower start
2. Log in to the UI — version number visible at bottom of sidebar
3. The update banner appears if a newer GitHub release exists (you can test by setting `APP_VERSION=0.0.1` as a build arg)
4. Click "Update now" — calls `/admin/upgrade`, Watchtower pulls the latest image and restarts the container
5. Without Watchtower: stop it with `docker compose stop watchtower`, verify the banner falls back to showing the release link instead of the button

### DIFF
--- a/scripts/jentic-update.sh
+++ b/scripts/jentic-update.sh
@@ -1,7 +1,7 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Manual upgrade script for Jentic Mini.
 # Run this on the host if Watchtower is not configured.
-set -e
+set -euo pipefail
 cd "$(dirname "$0")/.."
 echo "Pulling latest Jentic Mini image..."
 docker compose pull jentic-mini


char0n:
## Scope change: removing one-click upgrade, keeping version display

After review and testing, we're stripping the one-click upgrade mechanism from this PR and keeping only:
- **Version display** in the sidebar footer (always shows current version)
- **Update banner** with "View release notes" link when a newer GitHub release exists

### Why

**Watchtower is effectively EOL.** Last release Nov 2023, ships Docker client API v1.25 which is incompatible with Docker Engine v24+ (requires API v1.40+). No maintained drop-in alternative exists that combines image pull + container restart + HTTP trigger API.

**Pulling images locally doesn't make sense.** The `compose.yml` is a dev tool — it builds from source (`build: context: .`), bind-mounts `./src`, and tags a local image. `docker compose pull` has nothing to pull from (no registry configured). Developers rebuild from source, they don't pull.

**Production doesn't use compose.** The DigitalOcean deploy guide uses `docker run` directly with `ghcr.io/jentic/jentic-mini:latest`. A compose-based upgrader sidecar wouldn't work there either.

**A custom sidecar adds complexity without a clear deployment target.** We prototyped a minimal Python-based sidecar (scripts/upgrader.py) but it requires knowing the deployment method (compose vs docker run), project name, compose file path, etc. Too much coupling for marginal UX gain.

### What stays
- `GET /version` — current version + latest GitHub release (cached 6h) + release URL
- Sidebar footer showing `v{currentVersion}`
- Amber update banner with "View release notes" linking to the GitHub release page
- `scripts/jentic-update.sh` — CLI upgrade script for host-side use

### What gets removed
- `POST /admin/upgrade` endpoint
- Watchtower sidecar from compose.yml
- Upgrader sidecar (scripts/upgrader.py, scripts/upgrader.Dockerfile)
- `WATCHTOWER_*` / `UPGRADER_*` env vars
- "Update now" button in the UI
- `upgradeAvailable` / `upgrade_available` / `watchtower_configured` fields

PR #105, opened 2026-03-30
Fix hardcoded /app/data paths that break local dev without Docker

MichaelCordner:
## Summary

`startup.py` and `catalog.py` hardcoded `/app/data` for install ID files, catalog manifests, and the specs directory. This works inside Docker (where `/app/data` is the bind-mounted data volume) but breaks when running the server locally with just `uvicorn` and a `DB_PATH` env var.

**Changed files:**
- `src/routers/catalog.py` — `CATALOG_MANIFEST_PATH` and `WORKFLOW_MANIFEST_PATH` now use `DATA_DIR` from config instead of `Path("/app/data/...")`; removed unused `Path` import
- `src/startup.py` — `_INSTALL_ID_FILE`, `_INSTALL_REGISTERED_FILE` now use `DATA_DIR`; specs directory uses `SPECS_DIR` from config

**Not changed** (already correct):
- `src/vault.py` — derives `_KEY_FILE` from `DB_PATH` env var
- `src/routers/debug.py` — reads `DB_PATH` env var directly
- `src/config.py` — defines `DATA_DIR`, `SPECS_DIR`, `WORKFLOWS_DIR` from `DB_PATH`

After this fix, local dev works with:
```bash
DB_PATH=./data/jentic-mini.db .venv/bin/uvicorn src.main:app --reload --port 8900
```

## Test plan
- [x] `docker compose up -d` — verify existing Docker workflow still works
- [x] `DB_PATH=./data/jentic-mini.db uvicorn src.main:app --reload` — verify local dev starts without `/app` errors
- [x] Verify catalog manifest refresh, self-registration, and install ID creation all write to the correct data directory

### DIFF
--- a/src/startup.py
+++ b/src/startup.py
@@ -19,7 +19,7 @@ import uuid
 
 from src.db import get_db
 from src.brokers.pipedream import _API_ID_TO_PD_SLUG as _PIPEDREAM_APP_SEEDS
-from src.config import JENTIC_PUBLIC_HOSTNAME, DATA_DIR
+from src.config import JENTIC_PUBLIC_HOSTNAME, DATA_DIR, SPECS_DIR
 
 _REGISTER_INSTALL_URL = "https://api.jentic.com/api/v1/register-install"
 _INSTALL_ID_FILE = DATA_DIR / "install-id.txt"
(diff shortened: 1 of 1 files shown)


char0n (approved):
LGTM

PR #211, opened 2026-04-10
feat/credential-scheme-routes: self-hosted API support + broker decoupling

sean-at-jentic:
## Summary

This PR makes jentic-mini work properly with self-hosted APIs — Home Assistant, Discourse, Portainer, go2RTC and others like them. These APIs live on private networks, may use self-signed TLS, may have no authentication, and may run as multiple instances on different hosts. The hosted-API assumptions baked into the broker prevented all of this from working reliably.

The fix required addressing some architectural drift: the broker had accumulated direct dependencies on the `apis` table and on runtime spec lookups that tightly coupled it to OpenAPI. The broker is supposed to be a pure HTTP proxy — match a credential to the request host, inject auth, forward bytes, not touch the payload. This PR restores that contract.

The mechanism: a `credential_routes` table where each credential explicitly declares the route(s) it covers, and carries a pre-configured auth scheme blob. The broker resolves credentials by route alone — no catalog queries during request handling. A credential for a self-hosted service must define `server_variables` that specify where the service actually lives. For example, when setting up a credential for Discourse, the `defaultHost` server variable (defined in Discourse's OpenAPI description) must be provided. If an OpenAPI description doesn't provide a templated `server.url`, an overlay must be uploaded that adds one. The resulting server URL becomes the route that allows the broker to find the credential when handling a relevant request.

---

## New Features

**`credential_routes` table**

Each credential now explicitly declares the host(s) it covers (`credential_id`, `host`, `path_prefix`). The broker resolves credentials purely by matching the incoming request host against this table. No `api_id` intermediary; no `apis` table queries at request time.

**`auth_type=none`**

New credential type for no-auth local services. Carries `server_variables` to specify the actual network address. The broker skips credential injection entirely. The `value` field is optional for this type. Supported in both `POST /credentials` and `PATCH /credentials/{id}`.

**`server_variables` for host targeting**

When an OpenAPI spec uses server URL template variables (e.g. `http://{defaultHost}`), a credential's `server_variables` field fills those variables to derive the routing target. For specs with hardcoded server URLs, an overlay must be uploaded to introduce template variables before a credential can be configured — this is the correct and canonical path.

**`409 CREDENTIAL_AMBIGUOUS`**

When multiple credentials match the same host and no `X-Jentic-Credential` or `X-Jentic-Service` header is provided, the broker returns a `409` with a clear error listing the available credentials. Previously it silently picked the first match.

**Credential deep-link**

`/credentials/new?api_id=...&label=...&identity=...&server_vars[key]=val` pre-fills the credential form with all non-sensitive fields. An agent constructs the URL and sends it to the user; the user pastes only their secret. Documented in `docs/credential-deeplink.md`.

**Pipedream connect-link `api_id` override + auto-routes**

Connect-links now accept a caller-supplied `api_id` override that is stored on the synced credential. Routes are upserted into `credential_routes` on every sync. Previously routes were never written on sync, leaving every OAuth reconnect with empty routes that required a manual `PATCH`.

---

## Bugs Fixed

**Security: default-allow on unauthenticated passthrough**

Authenticated requests (carrying a `toolkit_id`) that couldn't resolve a credential for the target host fell through the injection logic and forwarded the request without credentials. Added an explicit fail-closed block: if no credential is found for an authenticated request, the broker returns `403 policy_denied` immediately.

**Compound auth scheme not injected**

Credentials with two-header auth schemes (e.g. Discourse's `Api-Key` + `Api-Username`) fell through the broker's injection path because it expected `"in"` at the top level of the scheme blob. Neither header was injected. Fixed by detecting and iterating the compound format.

**`GET /apis/{id}` always returned `500`**

A `NameError` in `_load_merged_spec` caused every call to the API detail endpoint to fail at runtime. Side effect: the credential edit page spinner never resolved, since the UI calls this endpoint to populate the form.

**`POST /credentials` silently dropped `routes`, `scheme`, `server_variables`**

These fields were accepted in the request model but never forwarded to the vault. Fixed.

**`DELETE /credentials` left orphan route rows**

SQLite `ON DELETE CASCADE` requires `PRAGMA foreign_keys=ON` per connection. Without it, deleting a credential left rows in `credential_routes` that contaminated credentials subsequently created with the same ID. Fixed with an explicit `DELETE FROM credential_routes`.

---

## Refactors

**`aiohttp` replaces `httpx` for upstream requests**

`httpx` automatically decoded `gzip`/`deflate` response bodies before the broker could forward them. This intermittently corrupted binary and compressed responses (e.g. git push/pull via the broker). `aiohttp` operates at a lower level and passes response bytes through untouched.

**Broker: zero `apis` table queries at request time**

`_resolve_credential_ids` and `_find_credential_for_host` now resolve purely via `credential_routes`. `get_merged_security_schemes` is removed from the broker entirely — all credentials carry pre-computed scheme blobs. Net: −94 lines in `broker.py`.

---

## Migration

Migration `0004` creates `credential_routes`, backfills routes for all existing credentials, and backfills `scheme` blobs for any credential that doesn't have one. No manual steps required on upgrade.

---

## Testing

Tested end-to-end through the live proxy with credential types: API key (header), API key (query param), Bearer token, Basic auth, compound two-header auth, OAuth2 (Pipedream-brokered), and `auth_type=none`. Multi-credential disambiguation, OAuth reconnect cycles, sync idempotency, and full credential CRUD were all verified.

---

## Docs

- `docs/credential-deeplink.md` — new: deep-link query parameter API and agent-assisted setup pattern
- `docs/server-variables.md` — new: overlay + server variable pattern for parameterised private hosts
- `AGENTS.md` — updated to document the bind-mount/hot-reload development model


char0n (commented):
_Superseded by updated review below._

char0n (commented):
## Review: PR #211 — Self-hosted API support + broker decoupling

_All review items addressed. 15 commits pushed to this branch._

### Summary of fixes applied

| # | Issue | Status |
|---|---|---|
| 1 | X-Jentic-Credential bypassed fail-closed check | **Fixed** — validates alias against route-matched credentials |
| 2 | api_id missing from route-resolved credentials | **Fixed** — added to SELECT; overlay confirmation works. +3 tests |
| 3 | auth_type=none blocked by security scheme gate | **Fixed** — skips gate when auth_type is "none". +2 tests |
| 4 | ssl=False for private hosts | **Won't fix** — TLS encryption still happens, only cert verification disabled. User controls private network. |
| 5 | cryptography downgraded to 46.0.6 | **Fixed** — restored to 46.0.7 |
| 6 | .local routing discards resolved scheme | **Fixed** — carries scheme through from resolved URL. +2 simulate-mode tests |
| 7 | apiKey in query param not handled | **Warning logged** — follow-up issue #224 |
| 8 | Auto-overlay overwrites existing template variables | **Fixed** — skips when base_url has template vars. +2 tests |
| 9 | aiohttp buffers full response body | **Not a regression** — httpx did the same. Follow-up issue #225 |
| 10 | 409 returned after upstream request | **Non-issue** — code already returns 409 before forwarding |
| 11 | server_variables_configured shows wrong credential | **Fixed** — filtered by api_id. +2 tests |
| 12 | Unused import in _backfill_credential_routes | **Fixed** |
| 13 | React useEffect deps missing | **Fixed** — added prefill?.serverVars to deps |
| 14 | (cred as any).scheme bypasses types | **Fixed** — typed as CredentialOut, removed all any casts |
| 15 | Self-registration credential api_id mismatch | **Fixed** — resolves actual API ID from DB (found during manual testing) |

### Test coverage added
- 85 backend tests passing (was 55 before this PR)
- 331 UI tests passing
- New test files: overlay confirmation, server URL resolution, self-hosted import, capability server vars

char0n:
All identified issues addressed. Manual testing still in progress...

char0n:
### Bug found during manual testing: credential edit UI stuck on loading spinner

**Root cause:** Self-registration creates the credential with `api_id = JENTIC_PUBLIC_HOSTNAME` (defaults to `localhost`), but `_register_openapi` derives the API ID as `jentic-mini.local` (private server URL → `.local` slug from title). The credential's `api_id` doesn't match the API's `id`.

The UI calls `GET /apis/localhost` → 404 → edit form spinner never resolves.

**Fix:** Resolve the actual registered API ID from the DB before creating the credential. Also fixed the "already imported" check to detect the `.local`-derived ID.

Commit: `fix: align self-registration credential api_id with derived API ID`

char0n (approved):

char0n:
## Full Review Analysis (for reference)

### What looks good
- Route-based credential lookup — clean separation, broker decoupled from apis table at request time
- Fail-closed security — authenticated requests without credentials return 403
- 409 CREDENTIAL_AMBIGUOUS — actionable error
- Migration backfill — defensive, handles existing data
- auto_decompress=False on aiohttp — correct fix for content-encoding corruption
- Compound auth scheme injection — Secret + Identity two-header pattern
- Self-hosted API overlay auto-generation — good DX

### Issues found and resolved

**1. SECURITY: X-Jentic-Credential bypassed fail-closed check** (broker.py:609)
Header set _resolved_cred_ids directly without route validation, letting agents proxy to arbitrary hosts. Fixed: resolves by route first, verifies alias is among matches.

**2. api_id missing from get_credentials_for_route()** (vault.py:526)
Broker expected creds[0].get("api_id") for overlay confirmation but query didn't SELECT it — always None. Overlay confirmation was silently broken. Fixed: added api_id to SELECT + return dict.

**3. auth_type=none blocked by security scheme gate** (credentials.py:211)
POST /credentials with auth_type=none hit the "no security scheme" 409. Prevented creating no-auth credentials for self-hosted APIs. Fixed: skip gate when auth_type is "none".

**4. ssl=False for private hosts** (broker.py) — **Won't fix.**
TLS encryption still happens — only certificate verification disabled. User controls private network. Requiring valid certs for every LAN service breaks the core self-hosted use case. Public hosts retain full TLS verification.

**5. cryptography downgraded** (requirements.txt)
46.0.6 reverted security patch from PR #193. Fixed: restored to 46.0.7.

**6. .local routing discards resolved scheme** (broker.py:874)
_resolve_server_url returned full URL with scheme but broker kept only netloc. Port heuristic defaulted HTTPS-on-443 private hosts to HTTP. Fixed: carries resolved scheme through.

**7. apiKey in query param not handled** (broker.py:247)
Injection logic only handled in:header. Now logs explicit warning instead of silent failure. Follow-up: #224.

**8. Auto-overlay overwrites existing template variables** (import_.py:184)
Ran for any self-hosted spec including those with existing {defaultHost} vars. Fixed: skips when base_url contains template vars.

**9. aiohttp buffers full response body** (broker.py)
Not a regression — httpx did the same. Follow-up: #225.

**10. 409 CREDENTIAL_AMBIGUOUS after upstream request**
Non-issue — code already returns 409 before forwarding.

**11. server_variables_configured shows unrelated credential** (capability.py:413)
Query fetched all toolkit credentials regardless of api_id. Fixed: filtered by api_id.

**12. Unused import** (startup.py:285) — Fixed.

**13. React useEffect deps missing** (CredentialFormPage.tsx:383)
Reads prefill?.serverVars but deps only had [serverVarDefs]. Fixed.

**14. (cred as any).scheme bypasses types** (CredentialsPage.tsx:567)
CredentialOut has typed scheme field. Fixed: removed all any casts, typed as CredentialOut.

**15. Self-registration credential api_id mismatch** (found during manual testing)
Credential created with api_id=localhost but API registered as jentic-mini.local. UI edit form showed perpetual loading spinner. Fixed: resolves actual API ID from DB.

### Test coverage gaps remaining
Missing simulate-mode injection tests for: bearer, basic, apiKey (header), compound two-header auth. These are the core broker features this PR changes.

PR #292, opened 2026-04-24
fix: use default_external_user_id from config in discover_accounts on broker creation (#158)

### LINKED ISSUE #158: OAuth broker creation ignores default_external_user_id from request body
## Problem

When creating an OAuth broker via `POST /oauth-brokers`, the `default_external_user_id` field from the request body is ignored. The backend hardcodes `"default"` in the INSERT statement (`oauth_brokers.py:220`), even though:

1. The UI has a field for it and sends it in the request body
2. `_extract_pipedream_config` parses other config fields but not `default_external_user_id`
3. The GET endpoints correctly return it from the DB (fixed in PR #107)

## Impact

Users who connect Pipedream with a non-default external user ID must manually update the DB after creation. The UI sync always uses `"default"` and finds 0 accounts.

## Fix

In `create_oauth_broker` (`src/routers/oauth_brokers.py:220`), read `default_external_user_id` from `body.config` and use it in the INSERT instead of hardcoding `"default"`.

Found during Pipedream Google Calendar testing.

nandanadileep:
## Summary

- `POST /oauth-brokers` accepts `default_external_user_id` inside the `config` dict and persists it correctly — but the `broker.discover_accounts(...)` call immediately after creation always passed the hardcoded string `"default"`, ignoring the caller-supplied value.
- In multi-user deployments this means the initial account sync runs under the wrong user identity, returning 0 discovered accounts even when the custom user has connected apps in Pipedream.

## Changes

- **`src/routers/oauth_brokers.py`**
  - `create_oauth_broker`: pass `default_external_user_id` instead of `"default"` to `broker.discover_accounts(...)`.
  - `_extract_pipedream_config`: fix return type annotation (`tuple[str, str, str, str|None, str]` → `tuple[str, str, str, str|None, str, str]`) and update docstring to include `default_external_user_id`.
  - `OAuthBrokerCreate.config` field description: document `default_external_user_id` as an accepted optional key.
- **`ui/openapi.json`**: regenerated to match updated `OAuthBrokerCreate.config` description (required by `test_ui_openapi_matches_served_spec`).
- **`tests/test_oauth_broker_default_user.py`**: two new tests — one asserts a custom `default_external_user_id` is persisted and returned by `GET /oauth-brokers/{id}`; one asserts omitting it falls back to `"default"`.

## Test plan

- [x] `test_custom_default_external_user_id_is_persisted` — passes
- [x] `test_omitted_default_external_user_id_falls_back_to_default` — passes
- [x] `test_ui_openapi_matches_served_spec` — passes with updated `ui/openapi.json`
- [x] All existing broker lifecycle tests pass

Closes #158

### DIFF
--- a/src/routers/oauth_brokers.py
+++ b/src/routers/oauth_brokers.py
@@ -132,7 +132,8 @@ class OAuthBrokerCreate(NormModel):
             "For `pipedream`: `client_id`, `client_secret`, `project_id` "
             "(all from Pipedream workspace → API settings → OAuth clients). "
             "Optional: `environment` (`production` or `development`, default `production`), "
-            "`support_email`."
+            "`support_email`, "
+            "`default_external_user_id` (user identity for initial account sync, default `\"default\"`)."
         ),
         examples=[_PIPEDREAM_CONFIG_EXAMPLE],
     )
@@ -201,10 +202,10 @@ def _make_broker_id(broker_type: str, existing_ids: list[str]) -> str:
     return f"{base}-{n}"
 
 
-def _extract_pipedream_config(config: dict) -> tuple[str, str, str, str | None, str]:
+def _extract_pipedream_config(config: dict) -> tuple[str, str, str, str | None, str, str]:
     """Extract and validate Pipedream config fields.
 
-    Returns (client_id, client_secret, project_id, support_email, environment).
+    Returns (client_id, client_secret, project_id, support_email, environment, default_external_user_id).
     app_name and logo_url are set by the backend — not exposed in the API.
     """
     missing = [f for f in ("client_id", "client_secret", "project_id") if not config.get(f)]
@@ -301,7 +302,7 @@ async def create_oauth_broker(body: OAuthBrokerCreate):
 
     accounts_discovered = 0
     try:
-        accounts_discovered = await broker.discover_accounts("default")
+        accounts_discovered = await broker.discover_accounts(default_external_user_id)
     except Exception as exc:
         log.warning("Account sync failed for broker %s: %s", broker_id, exc)
 
--- a/src/routers/oauth_brokers.py
+++ b/src/routers/oauth_brokers.py
@@ -133,7 +133,7 @@ class OAuthBrokerCreate(NormModel):
             "(all from Pipedream workspace → API settings → OAuth clients). "
             "Optional: `environment` (`production` or `development`, default `production`), "
             "`support_email`, "
-            "`default_external_user_id` (user identity for initial account sync, default `\"default\"`)."
+            '`default_external_user_id` (user identity for initial account sync, default `"default"`).'
         ),
         examples=[_PIPEDREAM_CONFIG_EXAMPLE],
     )
--- a/tests/test_oauth_broker_default_user.py
+++ b/tests/test_oauth_broker_default_user.py
@@ -1,0 +1,100 @@
+"""OAuth broker default_external_user_id tests.
+
+Verifies that POST /oauth-brokers correctly persists and returns the
+default_external_user_id from config, and that the value is passed
+to discover_accounts on creation rather than being silently overridden
+with the hardcoded string "default".
+"""
+
+import pytest
+
+from src.brokers.pipedream import PipedreamOAuthBroker
+
+
+BROKER_CUSTOM_USER = "test-oauth-broker-custom-user"
+BROKER_DEFAULT_USER = "test-oauth-broker-default-user"
+
+_BASE_CONFIG = {
+    "client_id": "test-client-id",
+    "client_secret": "test-client-secret",
+    "project_id": "proj_test123",
+}
+
+
+@pytest.fixture(scope="module")
+def broker_with_custom_user(admin_client):
+    resp = admin_client.post(
+        "/oauth-brokers",
+        json={
+            "id": BROKER_CUSTOM_USER,
+            "type": "pipedream",
+            "config": {**_BASE_CONFIG, "default_external_user_id": "alice"},
+        },
+    )
+    assert resp.status_code in (200, 201), f"Create failed: {resp.text}"
+    yield BROKER_CUSTOM_USER
+    admin_client.delete(f"/oauth-brokers/{BROKER_CUSTOM_USER}")
+
+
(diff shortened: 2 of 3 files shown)


char0n (changes requested):
Thanks @nandanadileep for digging into this — the root cause is correctly identified and the one-line fix at the `discover_accounts` call site is spot on. The stale return-type annotation, docstring, field description, and `ui/openapi.json` regeneration are all good catches. A few blocking issues before merge:

### 🔴 Must-fix

**1. Tests reference a deleted fixture — they won't run after rebase.**
`tests/test_oauth_broker_default_user.py` uses `admin_session` and `cookies=admin_session`. PR #312 (already merged to main after this branched from `04f6ea8`) replaced the `admin_session` cookie-dict fixture with an `admin_client` TestClient that persists cookies on the instance, and updated every call site. After rebase, pytest will fail with `fixture 'admin_session' not found`.

Follow the new pattern used throughout `tests/test_oauth_broker_lifecycle.py`:
```python
def broker_with_custom_user(admin_client):
    resp = admin_client.post("/oauth-brokers", json={...})
```
Drop the `cookies=...` kwarg everywhere.

**2. The tests don't exercise the bug the PR fixes.**
Both tests assert that `GET /oauth-brokers/{id}` returns the correct `default_external_user_id` — but persisting that value was already working on pre-fix `main` (`_extract_pipedream_config` already extracted it into the DB insert). Both tests would pass unchanged against the buggy code. The real fix is `broker.discover_accounts(default_external_user_id)` on line 304, and nothing in either test observes that call.

Please add a test that proves `discover_accounts` receives the caller-supplied value — e.g. monkeypatch `PipedreamOAuthBroker.discover_accounts` and assert it was called with `"alice"`, not `"default"`. Otherwise this exact regression can silently reappear.

**3. Commit message violates `.claude/rules/conventional-commits.md` and `git-workflow.md`.**
- **Missing scope** — should be `fix(oauth_brokers): …`
- **Title is 71 chars**, exceeds the 69-char header limit
- **`Fixes #158.` is in the commit body** — `git-workflow.md` forbids magic close-keywords in commit messages; they belong only in the PR body (which you already have as `Closes #158`). On squash merge this closes the issue twice and leaves stale auto-close text in permanent history.
- **Missing `Signed-off-by:` trailer** — every other recent commit carries a DCO sign-off (`git commit -s`).

Suggested header: `fix(oauth_brokers): pass configured default_external_user_id to initial sync` (68 chars). Drop `Fixes #158.` from the body and add the sign-off.

### 🟡 Nice-to-have

`_PIPEDREAM_CONFIG_EXAMPLE` (`src/routers/oauth_brokers.py:46`) doesn't include `default_external_user_id`. Since the description now advertises it, adding it to the Swagger example would make the option discoverable in `/docs`.

### ✅ What's good

- Minimal, surgical change — no incidental refactoring
- Correct regen of `ui/openapi.json` keeps the served spec and UI copy in sync
- Fixture teardown properly cleans up created brokers
- Scope check: `discover_accounts` at lines 678 (callback) and 851 (sync) already propagate the caller's value correctly — the fix is correctly localized to the create path
- PR body clearly explains both the user-visible symptom and the mechanical fix

Once the fixture rename, a meaningful regression test, and the commit-message cleanup land, this is ready to go. 🙌

char0n (changes requested):
Thank you! Please review my comment for this PR.

nandanadileep (commented):
Thanks @char0n, all three blocking points addressed in 2bfebd2:

1. **Fixture rename** -- switched both fixtures and all call sites off `admin_session`/`client` to `admin_client` (matching the pattern from #312).
2. **Regression test** -- added `test_discover_accounts_called_with_configured_user_id` which monkeypatches `PipedreamOAuthBroker.discover_accounts` and asserts it is called with `'alice'`, not `'default'`. This would have caught the original bug.
3. **Commit message** -- `fix(oauth_brokers): pass configured default_external_user_id to initial sync` (68 chars); removed `Fixes #158.` from the body (kept as `Refs #158.`); added `Signed-off-by` trailer.

Also rebased onto current main to pick up the #312 conftest changes. The nice-to-have (`_PIPEDREAM_CONFIG_EXAMPLE`) can follow in a separate cleanup if you'd like, or I can add it here.

char0n (approved):
Thanks for the quick turnaround @nandanadileep — all three must-fixes are addressed:

- ✅ **Fixture rename:** all three fixtures/tests now use `admin_client` and drop the `cookies=` kwarg, matching the post-#312 pattern.
- ✅ **Meaningful regression test:** `test_discover_accounts_called_with_configured_user_id` monkeypatches `PipedreamOAuthBroker.discover_accounts` and asserts it's called with `"alice"` — this would have failed against pre-fix `main`, which is exactly what we wanted.
- ✅ **Commit message:** scope (`oauth_brokers`), DCO sign-off, and `Refs #158` (instead of `Fixes #158.`) are all in place.

Approving — but one small thing for when you merge: since this repo squash-merges, the **PR title** (currently `fix: use default_external_user_id from config in discover_accounts on broker creation (#158)`, 92 chars) becomes the squash-commit headline, not the local commit message. Could you tighten the PR title to fit the 69-char Conventional Commits limit and add the scope? Something like:

> `fix(oauth_brokers): forward default_external_user_id to initial sync` (68 chars)

The local commit message you've written will be discarded by the squash, so it's just the title that needs to land. Up to you whether to fix it before merge or let a maintainer adjust at squash time.

Nice fix. 👍

char0n:
Pushed a lint fix directly to your branch (`1275b2b` — "style(oauth_brokers): apply ruff format and isort fixes"):

- `tests/test_oauth_broker_default_user.py`: removed the blank line between `import pytest` and `from src.brokers.pipedream...` (ruff I001) and let ruff format wrap the long test function signature.
- `src/routers/oauth_brokers.py`: ruff format preferred a single-quoted string literal for the new `default_external_user_id` description so the embedded `"default"` doesn't need escaping.

`pdm run lint` is clean locally. Once CI goes green this should be ready to merge — feel free to fix the PR title at squash time if you remember (the suggested `fix(oauth_brokers): forward default_external_user_id to initial sync` keeps it under the 69-char Conventional Commits limit), or a maintainer can adjust during the squash.

PR #310, opened 2026-05-04
fix(catalog): resolve phase 14 security advisories

char0n:
## Summary

Resolves all four open High-severity advisories from the Security tab per `specs/roadmap.md` Phase 14.

- **`py/path-injection`** at `src/routers/catalog.py` — determined to be a false positive. Traversal is already blocked by three independent layers (whitelist-sanitized `safe_id`, whitelist-sanitized `slug`, and a `resolve()` + `relative_to()` containment guard). CodeQL does not model `Path.relative_to()` as a sanitizer. Annotated the sink with `# codeql[py/path-injection]` and an inline comment explaining the defenses.
- **`wheel` `CVE-2026-24049`** (top-level + vendored inside setuptools) and **`jaraco.context` `CVE-2026-23949`** (vendored inside setuptools) — upgraded `pip`, `setuptools`, and `wheel` inside the final venv at Docker build time. Bootstrap tooling is not in `pdm.lock`, so the upgrade doesn't conflict with the frozen lockfile.

## Changes

- `src/routers/catalog.py` — inline sanitizer comment + `codeql[py/path-injection]` suppression on the `open()` sink
- `tests/test_catalog_path_traversal.py` — regression test: crafted Arazzo doc with traversal-laden `workflowId` must not produce any file outside `WORKFLOWS_DIR`
- `Dockerfile` — chain `pip install --upgrade pip setuptools wheel` to the `pdm install` step in the `py-deps` stage
- `specs/roadmap.md` — remove the shipped Phase 14 entry per the SDD lifecycle rule (numbers are stable; gaps preserved)

## Verification

- `pdm run lint` — clean
- `pdm run test tests` — 96/96 passing (including the new regression test)
- Post-merge: CodeQL and `docker-security.yml` workflows re-run against the new image; Security tab should show zero open High/Critical advisories

Refs: `specs/roadmap.md` Phase 14.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ COPY pyproject.toml pdm.lock ./
 # transitive CVEs in wheel and the setuptools-vendored copies of
 # wheel / jaraco.context reported by Trivy against the final image.
 RUN /root/.local/bin/pdm install --prod --no-editable --no-self --frozen-lockfile \
- && /app/.venv/bin/pip install --upgrade --no-cache-dir pip setuptools wheel
+ && /app/.venv/bin/python -m pip install --upgrade --no-cache-dir pip setuptools wheel
 
 # Stage 3: Runtime
 FROM python:3.11-slim
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,11 +17,13 @@ RUN curl -sSL https://pdm-project.org/install-pdm.py | python3 - --version 2.25.
 
 COPY pyproject.toml pdm.lock ./
 # Install locked project deps, then upgrade bootstrap tooling inside the venv.
+# PDM-managed venvs don't bundle pip, so bootstrap it with ensurepip first.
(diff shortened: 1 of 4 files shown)


char0n:
Copilot thread 3 (pin pip/setuptools/wheel) — **intentionally not addressed**.

Unpinned `--upgrade` is the right default here because:
- These are bootstrap-only packages, not imported at runtime. Runtime compatibility risk is near zero.
- They're not in `pdm.lock`, so Dependabot can't track them. Pinning means every new CVE needs a manual bump.
- Unpinned upgrade is self-healing: next image build clears whatever Trivy finds next.
- The image isn't reproducible today anyway (base images / apt packages aren't digest-pinned); pinning three venv packages wouldn't change that.

Trivy on every image build is the regression gate. Added an explanatory comment in the Dockerfile (`fea1a9b`) so future readers see the rationale.

Other two Copilot threads were already resolved by `24ac390`: slug assertion now enforces traversal-token absence, and the workflow manifest is snapshotted + restored in `finally`.

PR #344, opened 2026-05-07
fix(docker): dev compose venv binaries + vite node_modules permissions

Manuel-Jentic:
## Summary

Fixes local development when using `docker compose -f compose.yml -f compose.dev.yml up`.

### 1) `docker-entrypoint.sh` — always execute via venv binaries

Run Alembic, broker-app seed, and Uvicorn via `/app/.venv/bin/python` and `/app/.venv/bin/uvicorn` instead of relying on `python3` / `uvicorn` PATH resolution.

This removes ambiguity around interpreter resolution and prevents the class of "wrong Python binary" failures in environments where PATH resolution differs.

### 2) `compose.dev.yml` — fix `ui_node_modules` EACCES and keep Linux UID/GID mapping

The Vite service now:

- uses the same `JENTIC_UID` / `JENTIC_GID` pattern as `compose.yml` (defaults `1000:1000`)
- creates a writable npm HOME per uid/gid at `/tmp/jentic-mini-npm-home-<uid>-<gid>`
- `chown`s both `/app/ui/node_modules` and that npm HOME to `JENTIC_UID:JENTIC_GID`
- drops privileges with `setpriv` before running `npm install` and `npm run dev`

This fixes both permission paths:

- named-volume ownership issue on `/app/ui/node_modules`
- npm cache writes that would fail if HOME remained `/root`

Result: Vite/npm/esbuild run non-root with the configured UID/GID, including non-1000 Linux users.

## Docs updates

Updated:

- `README.md`
- `DEVELOPMENT.md`
- `specs/tech-stack.md`

to clarify that `compose.dev.yml` now honors `JENTIC_UID`/`JENTIC_GID` as well, and to recommend exporting them on Linux when host UID/GID differs from `1000:1000`.

## How to verify

```bash
# Optional cleanup if old root-owned volume state exists
docker compose -f compose.yml -f compose.dev.yml down -v

# Default uid/gid flow
docker compose -f compose.yml -f compose.dev.yml up --build

# Non-1000 Linux-style simulation
JENTIC_UID=1337 JENTIC_GID=1441 docker compose -f compose.yml -f compose.dev.yml up --build

# Vite reachable
curl -sSf -o /dev/null -w '%{http_code}\n' http://localhost:5173/

# Process uid/gid inside vite container
# (expect 1000:1000 by default, or provided JENTIC_UID:JENTIC_GID when overridden)
docker exec jentic-mini-vite-1 sh -c 'for p in /proc/[0-9]*; do n=$(cat "$p/comm" 2>/dev/null || true); case "$n" in MainThread|esbuild|npm|sh) awk "/^Uid:|^Gid:/{print}" "$p/status" 2>/dev/null; break;; esac; done'
```

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -46,6 +46,8 @@ docker compose up -d                                  # base only
 docker compose -f compose.yml -f compose.dev.yml up   # with UI dev server
 ```
 
+Export `JENTIC_UID=$(id -u)` and `JENTIC_GID=$(id -g)` (same shell line as `docker compose ...`) whenever your workstation UID/GID differs from Compose’s defaults (**1000:1000**) — notably on Linux, so `./ui`, `./src`, and `./data` stay owned by your user for both **jentic-mini** and **vite** (`compose.dev.yml` passes the same values through).
+
 The API is available at `http://localhost:8900` and the Swagger UI at `http://localhost:8900/docs`.
 
 ### 5. UI development (optional)
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ For UI development, use the dev compose override which runs Vite in a container 
 docker compose -f compose.yml -f compose.dev.yml up
 ```
 
-This starts a Vite dev server on `http://localhost:5173` with hot module replacement, proxying API calls to the backend on port 8900. Edit files in `ui/` and changes appear instantly.
+This starts a Vite dev server on `http://localhost:5173` with hot module replacement, proxying API calls to the backend on port 8900. Edit files in `ui/` and changes appear instantly. The Vite service uses the same `JENTIC_UID` / `JENTIC_GID` as `compose.yml` (defaults **1000:1000**), including for the bind-mounted `ui/` directory and for `npm`’s cache — **on Linux**, if your workstation user isn’t UID 1000, export matching values before spinning up compose (same snippet as below) so `./ui` is not littered with root- or mismatched-owner files from the container.
 
 Alternatively, run Vite directly on the host (requires Node.js 24):
 
@@ -206,9 +206,14 @@ See `ui/TESTING.md` for the full contributor guide.
 
 To rebuild the production UI bundle: `cd ui && npm run build`, then `docker compose up -d --build`.
 
-> **Note:** The container runs as a non-root user. `compose.yml` defaults to uid/gid 1000 for bind mount compatibility.
-> If your host user has a different uid, set `JENTIC_UID` and `JENTIC_GID` before starting:
-> `JENTIC_UID=$(id -u) JENTIC_GID=$(id -g) docker compose up -d`
+> **Note:** The backend container runs as a non-root user. `compose.yml` defaults to uid/gid 1000 for bind mount compatibility.
+> If your host user has a different UID/GID (common on Linux), set `JENTIC_UID` and `JENTIC_GID` before starting — applies to **`compose.yml`** and to **`compose.dev.yml`** so the bundled Vite service matches too:
+>
+> ```bash
+> JENTIC_UID=$(id -u) JENTIC_GID=$(id -g) docker compose up -d
+> # or with UI dev server:
+> JENTIC_UID=$(id -u) JENTIC_GID=$(id -g) docker compose -f compose.yml -f compose.dev.yml up
+> ```
 
 Swagger UI is available at `http://localhost:8900/docs` for interactive API exploration.
 
--- a/compose.dev.yml
+++ b/compose.dev.yml
@@ -13,7 +13,6 @@
 services:
   vite:
     image: node:24-slim
-    user: "${JENTIC_UID:-1000}:${JENTIC_GID:-1000}"
     restart: unless-stopped
     working_dir: /app/ui
     command: sh -c "npm install --ignore-scripts && npm run dev"
--- a/compose.dev.yml
+++ b/compose.dev.yml
@@ -35,11 +35,15 @@ services:
       - |
         set -eu
         mkdir -p "$$JENTIC_NPM_HOME"
-        chown -R "$$JENTIC_UID:$$JENTIC_GID" /app/ui/node_modules "$$JENTIC_NPM_HOME"
-        HOME="$$JENTIC_NPM_HOME"
-        export HOME
-        exec setpriv --reuid="$$JENTIC_UID" --regid="$$JENTIC_GID" --clear-groups \
-          sh -c 'cd /app/ui && npm install --ignore-scripts && npm run dev'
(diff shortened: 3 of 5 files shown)


char0n:
Thanks for chasing the non-1000 Linux UID case down — it was genuinely broken before, and the iteration history on the branch shows the trade-offs were thought through carefully.

The chown-then-drop-priv structure is the right pattern. Per-uid `HOME` under `/tmp` keeps the cache isolated when `JENTIC_UID` changes, and `--clear-groups` plus `--reuid` (rather than `--euid`) close off the easy escape paths from the dropped process. Verified locally that `setpriv`, `chown`, and `mkdir` all ship in `node:24-slim` (trixie-slim base).

## Worth discussing

**`compose.dev.yml:38` — `chown -R` on every restart.** Copilot raised the same concern on this line ([thread](https://github.com/jentic/jentic-mini/pull/344#discussion_r3200875011)) and I think they're right that this can become noticeable once `node_modules` is populated (~25k files in a typical Vite/React tree). A stat-guard would scope the recursive walk to first-init / UID-change cases and let warm restarts stay quick:

```sh
NM_DIR=/app/ui/node_modules
if [ "$$(stat -c %u "$$NM_DIR")" != "$$JENTIC_UID" ] \
   || [ "$$(stat -c %g "$$NM_DIR")" != "$$JENTIC_GID" ]; then
  chown -R "$$JENTIC_UID:$$JENTIC_GID" "$$NM_DIR"
fi
```

The cheap chown on `$JENTIC_NPM_HOME` can stay unconditional.

**`specs/tech-stack.md:101` — depth of the new note.** `.claude/rules/sdd-constitution.md` calls out tech-stack as the place for load-bearing invariants, with operational detail living in `docs/` or `DEVELOPMENT.md`. The new parenthetical (`HOME` under `/tmp`, brief root chown on the named volume) reads as implementation detail that's already captured better in the compose file's inline comments and the README. Something like:

> **User:** Non-root `jentic` system user; UID/GID configurable via `JENTIC_UID` / `JENTIC_GID`; `compose.dev.yml`'s Vite service honours the same values.

**`docker-entrypoint.sh` — defensive vs bug-fixing.** `Dockerfile` line 55 already sets `ENV PATH="/app/.venv/bin:$PATH"`, so `python3` and `uvicorn` were resolving into the venv before this change. Happy to keep the explicit form as a clarity improvement; the commit message just slightly overstates the necessity.

## Nits

- nit: `compose.dev.yml:42` — the `cd /app/ui` inside the inner `sh -c` is dead since `working_dir: /app/ui` already sets cwd and `setpriv` doesn't reset it.
- nit: `compose.dev.yml:39-40` — `HOME="..."; export HOME` could collapse to `export HOME="..."`.
- nit: `compose.dev.yml:41` — optional hardening: `--no-new-privs` on `setpriv`. Paranoia for npm + Vite, but it's one flag.

## Followup (out of scope)

`compose.dev.yml` isn't covered by CI today (`ci-docker.yml` exercises `compose.yml` only). Worth a small smoke test that brings up the dev override, curls `:5173`, and asserts the in-container Vite UID matches `JENTIC_UID` — separate issue.

Manuel-Jentic:
Thanks @char0n and @copilot-pull-request-reviewer — all in-scope items addressed in befe2b0. One nuance below.

## What changed in this revision (`compose.dev.yml`)

- **Stat-guard around `chown -R /app/ui/node_modules`** — only walks when the directory's uid/gid don't already match `JENTIC_UID:JENTIC_GID`. Same pattern as the official `postgres` image's `docker-entrypoint.sh`. The cheap chown on `JENTIC_NPM_HOME` stays unconditional and non-recursive (fresh dir, contents only ever written by the dropped user).
- **Removed the dead `cd /app/ui`** inside `sh -c` — `working_dir: /app/ui` plus `setpriv` preserving cwd makes it a no-op.
- **Collapsed** `HOME="…"; export HOME` to a single `export`.
- **Added `--no-new-privs`** to `setpriv`. `--ignore-scripts` already disables npm postinstall, so this is defense-in-depth against any setuid escape from inside the dev tree.

`specs/tech-stack.md`: trimmed to your suggested wording — the `HOME`/chown specifics live in `compose.dev.yml`'s inline comments and the README, not in the invariants doc.

## Smoke tests (run before push)

**Focused harness** — exercises the exact command body in isolation against three host-UID scenarios, with a second pass for each to verify the stat-guard:

| Scenario | UID:GID | First init | Warm restart |
|---|---|---|---|
| Linux/mac default | `1000:1000` | `chown -R` runs | skipped ✓ |
| macOS host | `501:20` | `chown -R` runs (UID change) | skipped ✓ |
| Linux non-1000 | `1234:1234` | `chown -R` runs (UID change) | skipped ✓ |

Each pass verifies: `setpriv` drops to the right uid/gid, `HOME` is writable by the dropped user, `node_modules` is writable by the dropped user, and the final ownership matches the requested UID.

**End-to-end** — real `docker compose -f compose.yml -f compose.dev.yml up`:

- backend `:8900/health` → `200`
- vite `:5173/` → `200`
- vite container processes (`pid 1 sh`, `pid 24 npm run dev`, `pid 36 node MainThread`, `pid 44 esbuild`) all run as **uid=1000** via setpriv
- warm restart of `vite`: first `200` in **~2s**, with `npm install` reporting `up to date, audited 880 packages in 765ms` — confirms the recursive chown was skipped and the named volume cache held

## On `docker-entrypoint.sh` — clarity, not a bug-fix

You're right that `Dockerfile:55` already sets `ENV PATH="/app/.venv/bin:$PATH"`, so the original `python3` / `uvicorn` invocations were resolving into the venv. Keeping the explicit `/app/.venv/bin/...` form as a clarity improvement (no behavior change), as you suggested.

## Out of scope (filing as follow-up)

Adding a CI smoke job for `compose.dev.yml` (asserting `:5173`, in-container UID matches `JENTIC_UID`) — tracked in #347.

char0n:
All four review points are addressed cleanly in `befe2b0`:

- The stat-guard on `/app/ui/node_modules` matches the suggestion verbatim, and dropping the `JENTIC_NPM_HOME` chown to non-recursive is a nice tightening — `mkdir -p` only touched the top-level dir, so the recursive form was overkill there too.
- `--no-new-privs`, the collapsed `export HOME=...`, and the dead `cd /app/ui` cleanup all land as expected.
- `specs/tech-stack.md` reads back as a load-bearing invariant now.

The three-UID smoke-test scenario coverage in the commit message (default 1000:1000, macOS 501:20, non-1000 Linux 1234:1234, plus warm restarts) is more thorough than I'd have asked for. LGTM.

One follow-up worth considering separately (still out of scope for this PR): bringing `compose.dev.yml` under CI so this configuration regresses loudly next time something shifts.

char0n (approved):
Verified by manual smoke test

Issue #325, opened 2026-05-05
modify_permissions access requests replace entire agent ruleset — should be incremental


sean-at-jentic:
## Problem

When a `modify_permissions` access request is approved, it **replaces** the entire agent ruleset rather than appending new rules. This means an agent requesting access to a new API inadvertently drops all existing permissions.

### Reproduction

1. Agent has `allow POST ^/repos/jentic/jentic-mini/issues$` in its rules
2. Agent files a `modify_permissions` request for `allow POST ^/repos/jentic/jentic-skills/issues$`
3. Human approves
4. Agent now has **only** the jentic-skills rule — jentic-mini access is gone

This was confirmed by testing: an agent with two jentic-mini rules requested two jentic-skills rules, and upon approval the jentic-mini rules were completely removed.

### Security concern

Best practice for agent permissions is **minimal incremental expansion** — agents should only request the narrowest additional access they need. The current API makes this extremely hard to use safely:

- To safely add one new rule, the agent must first read all existing rules, then re-submit the entire set plus the new rule
- If the agent forgets to include an existing rule, that permission is silently dropped on approval
- There is no way to incrementally add rules without knowing and re-stating the full current set
- This creates a constant risk of inadvertent privilege escalation (if the agent gives up and requests broad access) or loss of existing access (if the agent omits existing rules)

The principle of least privilege demands that requesting *only* the minimal new access should be the easy, natural path — not a multi-step dance that requires perfect recall of existing rules.

## Suggested Fix

Make `modify_permissions` access requests **incremental** — new rules are appended to existing agent rules, not replacing them.

- `POST /toolkits/{id}/access-requests` with `modify_permissions` → new rules are **added** to the existing agent rules
- `PUT /toolkits/{id}/credentials/{cred_id}/permissions` → full-replace semantics remain for when a human explicitly rewrites the policy (requires human session auth, not agent API key)

This aligns the API with the security principle of least privilege: agents can ask for the smallest possible expansion, and humans can review just that delta. The full-replace path remains available for human-initiated policy rewrites.

### Alternative: add a `type: "add_permissions"` request type

If changing `modify_permissions` semantics is too risky (existing behavior may be relied upon), add a new request type that is explicitly incremental:

```json
{
  "type": "add_permissions",
  "credential_id": "api.github.com-personal",
  "rules": [
    {"effect": "allow", "methods": ["POST"], "path": "^/repos/jentic/jentic-skills/issues$"}
  ],
[...]

Issue #324, opened 2026-05-05
Access requests should validate rules and provide guidance on rejection


sean-at-jentic:
## Problem

The API accepts `modify_permissions` access requests with:
- Empty or missing `rules` (defaults to `[]`) — effectively requesting unscoped access
- No `reason` — human reviewer has no context for the request
- Rules that duplicate system safety rules (deny-writes, allow-all) — redundant and confusing
- `allow` rules with no `path` or `operations` constraint — blanket access

There is no server-side validation or feedback. The agent gets no indication that its request is poorly scoped, and the human reviewer sees "Modify permissions: 0 rule(s)" with no idea what's being requested.

## Suggested Fix

Add validation to `POST /toolkits/{id}/access-requests`:

1. **`rules` must be non-empty** for `modify_permissions` — reject with 422 if omitted or `[]`
2. **`reason` must be non-empty and substantive** (minimum length, e.g. 20 chars) — forces agents to explain *why*
3. **Reject rules identical to system safety rules** (exact match on `effect` + `methods` + `path`) — they're redundant since system rules are always appended
4. **Warn or reject `allow` rules with no `path` or `operations` constraint** — this is effectively "allow everything" and almost never intended

### Error response format

When validation fails, return 422 with actionable guidance:

```json
{
  "detail": {
    "type": "invalid_access_request",
    "message": "Access request validation failed",
    "issues": [
      "rules must not be empty — specify the minimum permissions needed",
      "rule 3 duplicates a system safety rule (deny writes) — remove it, system rules are always appended",
      "rule 4 (allow with no path/operations constraint) is too broad — add a path or operations constraint"
    ],
    "best_practice": {
      "summary": "Access requests should specify minimal, well-scoped rules",
      "rules": [
        "Specify only the rules you need beyond system safety defaults",
        "Do not duplicate system safety rules — they are always appended",
        "Always anchor allow rules with a path regex (e.g., \"^/repos/myorg/myrepo/issues$\")",
        "Always check existing permissions before requesting — include rules you want to keep since modify_permissions replaces the entire ruleset",
        "Provide a clear, specific reason explaining what you'll use the access for"
      ],
      "example": {
        "type": "modify_permissions",
        "credential_id": "api.github.com-personal",
        "rules": [
          {"effect": "allow", "methods": ["POST"], "path": "^/repos/myorg/myrepo/issues$"},
[...]

Issue #133, opened 2026-04-01
Add 'pattern' field for regex matching in permission rules


char0n:
## Summary

Per @sean-at-jentic's proposal in #95:

- `path` should be evaluated as a **path prefix** only (not regex)
- A new optional `pattern` field should be evaluated as a **regex** against the full path+query
- This makes regex usage explicit — important when dealing with security policy

## Proposed behavior

| Field | Match type | Scope |
|-------|-----------|-------|
| `path` | Prefix match | Path only |
| `pattern` | Regex (`re.search`) | Full path + query string |

A rule must have either `path` or `pattern` (or both). Rules with neither are rejected.

## Example

```json
{"effect": "allow", "methods": ["POST"], "path": "/v1/messages"}
{"effect": "allow", "methods": ["GET"], "pattern": "/repos/[^/]+/[^/]+/issues\\?state=open"}
```

## Context

Currently `path` uses `re.search` (substring regex), which is broader than users expect. Splitting into prefix-only `path` and explicit `pattern` makes the security posture clearer.

Proposed by @sean-at-jentic in https://github.com/jentic/jentic-mini/issues/95#issuecomment-4169193761

Issue #122, opened 2026-03-31
Credential policies are global, not scoped per-toolkit


char0n:
## Summary

`credential_policies` is keyed by `credential_id` alone (UNIQUE constraint). When the same credential is bound to multiple toolkits, editing permissions via `/toolkits/{A}/credentials/{cred_id}/permissions` changes the policy for toolkit B too. There's no per-toolkit policy override.

## Current behavior

- `credential_policies` table: `credential_id TEXT NOT NULL UNIQUE`
- `INSERT ... ON CONFLICT(credential_id) DO UPDATE` — overwrites globally
- The toolkit_id in the API path (`/toolkits/{id}/credentials/{cred_id}/permissions`) is used for access control but the stored policy is credential-scoped, not (toolkit, credential)-scoped

## Impact

An admin managing two toolkits with the same credential (e.g., shared GitHub PAT) cannot set different permissions per toolkit. Editing rules for one toolkit silently changes the other.

## Expected behavior

Either:
- Policy should be scoped to (toolkit_id, credential_id) — different toolkits can have different rules for the same credential
- Or the behavior should be documented and the UI should make it clear that changes are global

Found during analysis of #95.

Manuel-Jentic:
Resolved. Permission rules now live in `toolkit_permission_rules` keyed by `(toolkit_id, credential_id)`, and both the write path and the broker query filter on both columns. Editing one toolkit's permissions can no longer affect another. It is fully per-binding now.

PR #348, opened 2026-05-07
docs(roadmap): add phases 26, 27 — access-request safety

Manuel-Jentic:
## Summary

Promotes two open issues into active phases on `specs/roadmap.md`:

- **Phase 26 — Incremental `modify_permissions` Access Requests** — `High (blocker)`, `Depends on: none (self-contained)`. Closes the silent-privilege-loss footgun where approving a `modify_permissions` request replaces the agent's entire ruleset. Refs #325.
- **Phase 27 — Validate Access Requests with Actionable Feedback** — `High`, `Depends on: none (self-contained — see context)`. Reject malformed requests (empty rules, no reason, redundant system rules, blanket `allow`) with structured 422 guidance the agent can self-correct against. Refs #324.

> Renumbered from 25/26 → 26/27 after [#357](https://github.com/jentic/jentic-mini/pull/357) (Phase 25 — Reverse-Proxy Path Prefix) merged to `main`. No existing phase is renumbered; the new phases take `max(active) + 1` per the lifecycle rule.

This PR is **roadmap-only** — no code changes. Once merged, each phase is materialized into a feature spec via `/sdd-new-spec <N>` (per `.claude/skills/sdd-new-spec/SKILL.md`), and implementation follows in a separate PR per phase.

## Why these priorities

`specs/roadmap.md`'s priority legend defines `(blocker)` as a trust/security gap that makes Mini unsafe for real agent usage today.

- **#325 → `High (blocker)`.** Silent replacement of an agent's ruleset on approval pushes agents toward over-broad asks and breaks the human reviewer's mental model of what was granted. Same bar as Phase 16 (hash toolkit keys), Phase 17 (scope traces), Phase 18 (control-plane audit), Phase 19 (legacy HTML XSS).
- **#324 → `High`.** Important hardening, but agents can recover today via trial-and-error and human guidance — does not independently gate early-access safety.

## Order and dependency (soft coupling)

The two phases are **independently shippable**. The four validation checks in Phase 27 all stand under either the current replace semantics or the post-Phase 26 append semantics — only the wording in the 422 `best_practice` payload (the example request and the prose around what "empty `rules`" means) revises when Phase 26 lands. If Phase 27 ships first, empty `rules` rejection actively prevents a silent wipe under today's replace behavior.

## Adjacent items already parked in `Later Phases`

Cross-checked for duplication; neither overlaps:

- **"Access request editing"** (#82) — patch/cancel **pending** requests. Distinct: lifecycle vs. approval semantics + validation.
- **"Operation scoping presets"** (#133, #122, #63) — preset policy templates and `pattern` grammar. Adjacent (rule shape) but distinct.

## Files

- `specs/roadmap.md`

## Review guidance

Reviewers should verify:
- Phase 26 and Phase 27 goals match the issue intent (#325, #324 respectively).
- The `(blocker)` tagging on #325 is appropriate per the legend in the file's preamble.
- Phase numbers continue from 25 (no renumber); the parking lot below the `---` is untouched.
- Bullets accurately capture "shippable" tasks without inventing scope.
- Phase 26 first bullet commits to changing `modify_permissions` semantics as the primary direction; `add_permissions` is an explicit fallback only if a concrete BC blocker surfaces during `/sdd-new-spec`.

Refs #324
Refs #325

### DIFF
--- a/specs/roadmap.md
+++ b/specs/roadmap.md
@@ -380,6 +380,38 @@ Large upstream responses are fully buffered today, which spikes memory and block
 - Add tests covering unmounted (default) and mounted (`/foo`) modes — SPA load, asset paths, `/docs`, `/openapi.json`, internal routes
 - Close or cross-reference issue #356
 
+## Phase 26 — Incremental `modify_permissions` Access Requests
+
+**Goal:** Make `modify_permissions` access requests append to the agent's existing ruleset on approval instead of replacing it, removing the silent-privilege-loss footgun.
+**Depends on:** none (self-contained)
+**Priority:** High (blocker)
+
+Today, approving a `modify_permissions` access request silently replaces the agent's entire ruleset with the request's `rules`. An agent that asks for one new rule loses every rule it relied on — silently, on approval. This pushes agents toward over-broad initial asks (because narrowing later requires perfect recall of the current ruleset) and breaks the human reviewer's mental model of what they are approving. The behavior is `bug`+`security`-labelled and matches the early-access trust bar that other `(blocker)` phases (16, 17, 18, 19) target.
+
+- Change `modify_permissions` access-request semantics: on approval, append the request's rules to the agent's existing rules; never replace. The feature spec may introduce a parallel `add_permissions` request type as a fallback if a concrete backwards-compatibility blocker surfaces during design, but the primary direction is the in-place semantics change
+- Keep full-replace policy editing on the human-only `PUT /toolkits/{id}/credentials/{cred_id}/permissions` path for explicit admin rewrites
+- Document the change as logic-only — no schema migration expected (rule storage shape is unchanged)
+- Add integration tests: starting rules `A,B` + request `C` → `A,B,C` on approval; two sequential requests stack correctly; full-replace path on `PUT` still works
+- Audit `AGENTS.md` and `docs/auth.md` for wording that implies replace semantics; correct it
+- Close #325; cross-reference #324 (validation hardens the new shape)
+
+## Phase 27 — Validate Access Requests with Actionable Feedback
+
+**Goal:** Reject malformed `modify_permissions` access requests with structured 422 guidance an agent can self-correct against.
+**Depends on:** none (self-contained — see context)
+**Priority:** High
+
+The API currently accepts access requests with empty `rules`, no `reason`, rules that duplicate the always-appended system safety rules, or unconstrained `allow` rules — silently. Agents have no signal that the request is malformed, and human reviewers see "Modify permissions: 0 rule(s)" with no context. The result is wasted human review and a soft pressure toward broad asks. The four checks below stand under either current replace or post-Phase 26 append semantics; only the wording in the 422 `best_practice` payload (the example request and the prose around what "empty `rules`" means) revises when Phase 26 lands. Either order is shippable.
+
+- Reject `modify_permissions` requests with empty or missing `rules` (HTTP 422 with `type: "invalid_access_request"`)
+- Reject requests with no substantive `reason` (minimum length, e.g. 20 chars)
+- Reject rules that exact-match a system safety rule (same `effect` + `methods` + `path`) — they are redundant since system rules are always appended
+- Reject `allow` rules with no `path` and no `operations` constraint (effectively "allow everything")
+- Return 422 with a structured `issues` list (per-rule indices) and a `best_practice` payload containing summary, rules, and a minimal example, per the shape proposed in #324
+- Cover each rejection path with integration tests; cover happy-path acceptance with one representative request
+- Update `AGENTS.md` and any agent-facing examples to show a well-formed request
+- Close #324; cross-reference #325
+
 ---
 
 ## Later Phases (Not Yet Planned)


char0n (approved):
Thanks for putting this up — the PR body is exceptionally well-grounded (issue cross-refs, blocker-bar comparison, parking-lot dedup, dependency rationale). The roadmap content itself slots in cleanly: phase numbers continue from 24 with no renumbering, the block format matches `/sdd-new-phase`'s template, and the `(blocker)` framing on Phase 25 lines up with the legend at `specs/roadmap.md:19-25` and the precedent set by Phases 16–19. Behavioural claims hold up against the code — `src/routers/access_requests.py:91` literally says "replaces the current agent rules entirely", and `:609-619` confirms that's not a doc-only mismatch.

Approving — nothing here blocks merge. Two observations worth weighing before this gets cut into a feature spec; your call on whether to address them in this PR or fold them into `/sdd-new-spec 25`.

## Worth discussing

**Phase 26 `Depends on:` Phase 25 — hard or soft?**

The PR body's rationale is that #324's validation rules need to "target the post-#325 semantics," but reading the four checks (empty rules, no reason, redundant system rules, blanket `allow`), each one stands under either replace or append semantics. The redundant-system-rule comparison is the same either way (system rules are always appended — `src/routers/toolkits.py:108`). Empty-rules rejection is arguably *more* useful pre-#325 because it prevents a silent wipe. What genuinely changes is the *wording* in the 422 best-practice payload — example request, prose around what "empty" means. That feels like a soft coupling: Phase 26 is shippable independently with a one-line note that error copy revises when #325 lands. Worth weighing against the cost of holding #324 behind #325 if #325 turns out to have a longer design loop (the `add_permissions` fork in Phase 25 hints it might). I might be missing context here.

**Phase 25 first bullet — can we commit to one direction?**

The bullet embeds a conditional fork: "...append ...; never replace; if backwards compatibility is required, introduce a parallel `add_permissions` request type." For a roadmap entry I think hedging is fine, but the PR body and goal both commit to changing `modify_permissions` semantics directly — the bullet's escape hatch undermines that commitment, and a feature-spec writer reading just the roadmap might read "decision deferred" rather than "primary path is in-place change, fallback only if BC forces our hand." Also, BC for `modify_permissions` is a soft constraint here (single-tenant self-host, agent-facing not a versioned public contract). Framing as "primary: change `modify_permissions` semantics; only introduce `add_permissions` if a concrete BC blocker surfaces during the spec" would carry the same hedge but signal the default direction.

## Nits

- PR body "Order and dependency" — "empty `rules` array means 'no-op' once requests are additive, not 'blanket access' as it does today" reads slightly off. Today, empty agent rules don't grant blanket access — they fall back to "Read-only (system safety rules apply)" per `src/routers/toolkits.py:161,288`. "...not a full ruleset reset as it does today" would be more precise. Roadmap content itself doesn't repeat the claim.
- Phase 25 `Depends on: none` — fine per template; neighbouring phases mostly use `none (self-contained)`. Minor consistency thing.
- Phase 25 third bullet (Alembic conditional) — based on the current apply path at `src/routers/access_requests.py:616`, this is logic-only and storage shape doesn't move, so the hedging reads slightly defensive. Could be tightened to "Document the change as logic-only — no schema migration expected."

## Out of scope follow-up

Once "Operation scoping presets" (parking lot — #133, #122, #63) lands, Phase 26's "blanket `allow`" detector will need to understand presets and tag-aware scoping. Worth a one-line cross-reference in the feature spec, not this PR.

char0n:
We need to renumber here because of https://github.com/jentic/jentic-mini/pull/357

Manuel-Jentic:
Thanks @char0n — both threads addressed in 1e0ec3c.

## Renumber (per your [comment](https://github.com/jentic/jentic-mini/pull/348#issuecomment-4406421102) on #357 landing)

Rebased onto `main` and renumbered to **Phase 26** (`modify_permissions` semantics) and **Phase 27** (validation), so Phase 25 — Reverse-Proxy Path Prefix stays where it landed. PR title + body updated. No existing phase renumbered.

## From the review

**"Worth discussing" #1 — Phase 27 dependency hard or soft.** Agreed, you talked me into it. Dependency softened to `none (self-contained — see context)`; context paragraph now states explicitly that the four checks stand under either replace or post-Phase 26 append semantics, with only the 422 `best_practice` copy revising depending on order. Either order is shippable. (Bonus: if Phase 27 ships first, empty-`rules` rejection actively prevents a silent wipe under today's replace behavior — which is a useful safety property to land independently.)

**"Worth discussing" #2 — Phase 26 first bullet hedging.** Reframed exactly as you suggested. Bullet now reads "Change `modify_permissions` access-request semantics: on approval, append…; never replace. The feature spec may introduce a parallel `add_permissions` request type as a fallback if a concrete backwards-compatibility blocker surfaces during design, but the primary direction is the in-place semantics change." Signals primary path + escape hatch, doesn't read as "decision deferred".

## Nits

- **PR body wording** — fixed. The "blanket access" line was wrong (empty agent rules fall back to read-only + system safety rules, not blanket access). The "Order and dependency" section now leads with soft-coupling instead of the misleading framing.
- **`Depends on: none` → `none (self-contained)`** — applied to both phases for consistency with neighbours.
- **Phase 26 Alembic bullet** — tightened to "Document the change as logic-only — no schema migration expected (rule storage shape is unchanged)." per the `src/routers/access_requests.py:616` apply path.

## Out of scope (noted for the feature spec)

The "Operation scoping presets" cross-reference (parking lot — #133, #122, #63) will be picked up in `/sdd-new-spec 27` for the "blanket `allow`" detector, as you suggested.

Issue #446, opened 2026-05-26
feat(discover): Add stats strip showing total APIs, workflows, and operations from catalog manifests


Manuel-Jentic:
## Summary

Add a `StatsStrip` to the Discover page (similar to the one on the Workspace page) that shows the **total number of APIs, workflows, and operations** available in the catalog. These numbers should come from the proper catalog manifests so they're always accurate.

## Motivation

The Workspace page already has a stats strip (`WorkspaceStatsStrip`) that gives users an at-a-glance overview of what they have. The Discover page is missing equivalent context — users landing there have no idea how large the catalog is or what's available to them.

Showing these totals sets expectations ("there are 1,200+ APIs with 150k+ operations to explore") and adds a sense of depth that encourages browsing and searching.

## Requirements

1. **Stats strip position** — Render just below the Discover page header, same visual pattern as the Workspace stats strip (compact horizontal ribbon with icon + label + value).
2. **Metrics to show**:
   - Total APIs (from the catalog manifest)
   - Total Workflows (from the catalog manifest)
   - Total Operations (aggregated from all API manifests/previews)
3. **Data source** — Pull numbers from the catalog manifests. This should work on a fresh workspace with zero imports. If the data isn't currently exposed by an endpoint, add a lightweight `/catalog/stats` or extend `/catalog/status` to include `total_apis`, `total_workflows`, `total_operations`.
4. **Consistency** — Use the same `Stat` component pattern (or a shared primitive) as the Workspace stats strip for visual consistency.

## Related

- #437 — Search overhaul (the catalog manifest indexing work in that issue will surface operation-level data that can power the operation count here)
- The Workspace `WorkspaceStatsStrip` component (`ui/src/components/workspace/WorkspaceStatsStrip.tsx`) is the reference implementation for layout and style

## Acceptance criteria

- [ ] Discover page shows a stats strip with total APIs, workflows, and operations
- [ ] Numbers are sourced from catalog manifests (not from imported/local data)
- [ ] Works on a fresh workspace with zero imports
- [ ] Visually consistent with the Workspace stats strip

Manuel-Jentic:
[...]

Issue #440, opened 2026-05-25
Persist catalog GitHub URL in apis table at import time


Manuel-Jentic:
## Problem

After importing a catalog API to workspace, the GitHub link to the public catalog source is lost. The `_links.github` field is only available on catalog entries (derived at runtime from the manifest's `path` field), but is never persisted to the `apis` SQLite table during import.

This means the Discover sheet's "GitHub" button disappears once the API is imported and viewed via deep link (where `initialEntity` is unavailable).

## Current state

- **Catalog entries** (`GET /apis?source=catalog`): include `_links.github` derived from `https://github.com/jentic/jentic-public-apis/tree/main/{entry['path']}`
- **Local entries** (`GET /apis?source=local`): only store `id`, `name`, `description`, `spec_path`, `base_url`, `created_at` — no GitHub URL or catalog path

## Proposed solution

1. Add a `catalog_github_url` (or `catalog_path`) column to the `apis` table
2. At import time (`POST /import`), when the source is a catalog spec URL, look up the manifest entry and persist the GitHub URL
3. Include `_links.github` in the local API response so the UI can always render the link

## Affected code

- `src/routers/import_.py` — persist the URL during import
- `src/routers/apis.py` — include in local entry response (line ~505-519)
- Schema migration for the `apis` table

## UI impact

Once this is in place, `SheetBody.tsx` can show the GitHub button for workspace APIs too, not just catalog ones.

Issue #438, opened 2026-05-22
Validate credentials after adding them (health check)


Manuel-Jentic:
## Problem

Currently when a user adds a credential for an API, we store it and assume it's valid. There's no verification step that actually tests the credential against the API to confirm it works.

This means:
- A user could paste an expired or malformed key and only discover it later when an agent tries to use it
- The `has_credentials` boolean doesn't reflect whether the credential is *actually usable*
- We can't distinguish between "credential missing" (never added) and "credential expired" (was valid, now isn't) in the UI

## Proposed solution

1. **On-add validation**: After a credential is saved, make a lightweight test request to the API (e.g. a low-cost read endpoint, or an OAuth token introspection) to verify it works. Surface success/failure to the user immediately.

2. **Periodic health check**: Optionally run a background check (e.g. daily) to detect credentials that have expired or been revoked. Update a `credential_status` field on the API record.

3. **Expose status in the API response**: Add a `credential_status: 'valid' | 'expired' | 'invalid' | null` field to `/apis` so the UI can render distinct "Credential expired" vs "Credential missing" states.

## UI impact

Once the backend exposes `credential_status`, the `CardStatusPill` component already has the type structure to support a split between "missing" and "expired" — just needs the signal.

Issue #437, opened 2026-05-22
Search overhaul: ranking, pagination envelope, catalog-operation discovery, and agent surface


Manuel-Jentic:
# Search overhaul: ranking, pagination envelope, catalog-operation discovery, and agent surface

## Summary

`/search` is the primary entry point for discovery for both agents and the UI. It mostly works, but on closer inspection it has a cluster of issues that are tightly coupled — fixing any one of them in isolation forces a wire-format rev that the next fix would also need. This issue tracks the whole cluster as a single piece of work.

The headline gap (which subsumes everything else): **on a fresh workspace, agents have no real way to search for operations or workflows that aren't already imported.** They get vendor names back and have to guess. Everything below is either contributing to that, or worsening it.

Found while polishing the Discover page UX (chat thread May 21–22, 2026).

## Background — what `/search` does today

- BM25 over locally registered operations + workflows (in-memory index, `src/bm25.py`).
- Plus a "catalog blender" that appends per-vendor `{type: "catalog_api"}` rows for unimported vendors whose `api_id` substring-matches the query.
- Plus a `has_workflows` boolean on catalog rows so the UI can render a `+ workflows` chip.
- Returns a bare list (`response_model=list[SearchResult]`), no envelope.
- Capped server-side at `n ≤ 100`. Each lane (BM25 + catalog) is independently capped at `n`, so the response can carry up to `~2n` rows.

## Problems

Numbered by priority. (0) is the strategic gap; (1)–(3) are the wire-format breakers that should land together; (4)–(11) are smaller corrections that ride along.

### 0. Catalog operations and workflows are not searchable across vendors *(highest leverage)*

An agent asks `"send email"`. The catalog has hundreds of vendors, dozens of which expose a "send email" operation. Today:

- BM25 indexes only **registered** operations/workflows. Unimported catalog content is invisible to it.
- The catalog blender only matches `api_id` substrings, so the agent gets vendor tiles like `mailgun.com` and `sendgrid.com` — but no signal about *which operations* match.
- `GET /catalog/{api_id}/preview-operations` and `…/preview-workflows` exist, but they're per-vendor and unindexed. An agent must:
  1. Search → get vendors.
  2. Guess which vendors are relevant.
  3. Call `preview_catalog_operations` per vendor.
  4. Filter client-side.
- Same gap on the UI: `ApiDetailSheet` only fetches operations *after* a vendor is picked.

[...]

PR #447, opened 2026-05-26
feat(workspace+discover): unified Discover surface and Workspace management

### LINKED ISSUE #161: Catalog browse section in UI lacks pagination
## Problem

The Catalog page has pagination for registered (local) APIs, but the catalog browse section loads all entries without pagination. With 1079+ catalog APIs, this creates a long unpaginated list.

## Expected behavior

Catalog browse should have the same pagination controls as the registered APIs section.

Found during catalog manifest fix (#157).

Manuel-Jentic:
## Summary

This PR replaces the old Search, Catalog, and Workflows pages with two new unified surfaces: **Discover** (browse + search the public catalog) and **Workspace** (manage imported APIs, workflows, credentials, and toolkits).

### Discover page (`/discover`)
- Unified browse + search replacing the separate Search, Catalog, and Workflows pages
- API detail sheet (slide-out panel) with operations list, workflow list, and import CTA
- Operation and workflow drill-down views within the sheet
- Paginated operations with full-dataset filtering (auto-fetches all pages in 200-item batches when a filter/tag is active)
- Import-to-workspace flow with proper source detection and post-import state refresh
- Keyboard shortcuts (`/` to focus search, arrow keys, `Enter`, `Esc`, `Cmd+/` for help)
- Recently inspected APIs strip for quick navigation

### Workspace page (`/workspace`)
- API and workflow cards with imported date, operation count, and workflow count
- Stats strip (total APIs, workflows, credentials, last activity)
- Search/filter with keyboard shortcuts matching Discover
- API detail view with full operation list, credentials, toolkit bindings, and workflow associations
- Workflow detail page with metadata, steps, and input schema
- Delete with cascade info dialog showing affected workflows, credentials (with specific labels), and toolkits (with linked credential names)
- Cascade delete option that removes credentials and toolkit bindings
- Import from URL, file upload, or catalog with `ImportSourceDialog`

### Legacy route preservation
Old bookmarks keep working — every removed route redirects to its new home, preserving query strings:
- `/catalog` → `/discover` (preserves `?q=…`)
- `/search` → `/discover` (preserves `?q=…`)
- `/workflows` → `/workspace`
- `/workflows/:slug` → `/workspace/workflows/:slug`

E2E coverage in `e2e/catalog.spec.ts`, `e2e/search.spec.ts`, and `e2e/workflows.spec.ts` asserts the redirect URL, the heading on the destination page, and that query strings round-trip.

### Shared UI additions
- `ConfirmDeleteDialog` — reusable delete confirmation with cascade impact visualization
- `KeyboardShortcutsBar` + `PageHelp` — platform-aware shortcut display (⌘ vs Ctrl)
- `SheetPrimitive` — animated slide-out panel primitive
- `Markdown` — safe markdown renderer for API descriptions
- `SearchInput`, `Checkbox`, `Kbd`, `WorkflowRow`, `Skeleton`, `Toaster`
- `BackButton` with history-aware navigation
- Nav tab indicator fix (removed `layout="position"` to prevent scroll-offset animation drift)

### Test infrastructure
A handful of cross-cutting test plumbing changes were needed to land the visual + a11y work without disabling assertions:
- App-wide `<MotionConfig reducedMotion="user">` wraps `<RouterProvider>` so framer-motion respects `prefers-reduced-motion` in production.
- Test wrapper sets `<MotionConfig reducedMotion="always">` so axe colour-contrast checks see post-animation pixel values rather than mid-fade translucent ones.
- Vitest browser config passes `context: { reducedMotion: 'reduce' }` to Chromium for the same reason at the browser level.
- `src/__tests__/setup.ts` now imports `@/index.css` so the design-system tokens (background, foreground, primary, muted, …) are available in the JSDOM-equivalent computed-style passes that axe relies on.
- `mockWorkflows` in `e2e/fixtures.ts` now matches both `**/workflows` and parameterised forms (`**/workflows?source=local` etc.) so the workspace stats strip's queries don't leak through to the network in E2E.

### New endpoints
- `GET /catalog/{api_id}/operations` — paginated operation preview for catalog APIs (limit capped at 200)
- `GET /catalog/{api_id}/workflows` — workflow preview for catalog APIs
- Enhanced `GET /apis` with `has_workflows` field and search/filter support

### Behavioral changes
- **Credential preservation on soft delete**: `DELETE /apis/{id}` (non-cascade) no longer clears `credentials.api_id` — credentials keep their reference so they auto-relink when the same API is re-imported
- **Cascade delete**: `DELETE /apis/{id}?cascade=true` deletes credentials AND their toolkit bindings
- **Path-based API IDs**: `DELETE /apis/{api_id:path}` supports IDs with slashes (e.g. `/api/v3`)
- Improved search with BM25 index rebuild after imports

### Tests added
- `test_reimport_links.py` — 4 scenarios: soft delete+reimport, cascade+reimport, toolkit bindings across both
- `test_cascade_delete.py` — soft and cascade delete behavior for credentials and toolkits
- `test_workspace_lifecycle.py` — full import → credential → delete → idempotent lifecycle
- `test_multi_source_lifecycle.py` — catalog, URL, and file upload import+delete
- `test_operations_pagination.py`, `test_search_filters.py`, `test_preview_catalog_operations_parser.py`

## Related issues (follow-up work)

- #451 — Soft-delete leaves toolkits in inconsistent state (broker bypass + misleading UX). Directly adjacent to the credential-preservation behaviour added here: the broker is orphan-tolerant, but workflow dispatch and `/inspect` aren't — needs a dedicated fix.
- #452 — Split `components/ui/` into primitives vs composed widgets. Surfaced by the detail-view refactors above; tracks the third step of the cleanup.
- #450 — Add favicon, meta title, and meta description. UI polish that's out of scope for this PR but wanted for the new surfaces.
- #446 — Add stats strip to Discover page (total APIs, workflows, operations from manifests).
- #438 — Validate credentials after adding them (health check). Pairs with the cascade-delete dialog work — once we have `credential_status` we can show distinct "expired" vs "missing" states.
- #437 — Search overhaul: ranking improvements, pagination envelope, catalog-operation discovery.
- #440 — Persist catalog GitHub URL in apis table at import time.
- #253 — Inspect links inconsistently URL-encode capability IDs.

## Test plan

- [x] 503 non-network tests pass
- [x] All 443 UI unit tests pass
- [x] All 33 Playwright E2E specs pass (including the new legacy-route redirect coverage)
- [ ] Discover → search → open sheet → import → "Open in Workspace" navigates correctly and workspace shows fresh data
- [ ] Workspace → API detail → delete (with/without cascade) → workspace refreshes, correct impact shown in dialog
- [ ] Workspace → filter operations → all ops searchable regardless of pagination
- [ ] Nav tab indicator animates horizontally only (no vertical drift when scrolled)
- [ ] Delete dialog shows only toolkits that actually bind credentials for the target API
- [ ] Visit `/catalog?q=stripe`, `/search?q=stripe`, `/workflows`, `/workflows/some-slug` — confirm redirects land on the new routes with the query string preserved

---

> **Merge strategy**: Squash + merge. Replace auto-generated commit list with this summary.

---

## Closes

- Closes #161 — Catalog browse now uses paginated `/apis` with Load More in both the Discover grid and the API detail sheet operations panel
- Closes #186 — `ui/openapi.json` regenerated and kept in sync via `test_ui_openapi_matches_served_spec`


Manuel-Jentic:
## Multi-perspective review

I ran six parallel reviews of this PR (architecture & IA, backend, frontend, tests, a11y/motion/keyboard, security). Each was scoped to one lane to avoid overlap. Below is the consolidated, de-duplicated map. **TL;DR: ship it after five small fixes.**

CI is fully green (10/10). No data-corruption risks. No security blockers. The test-infrastructure plumbing (MotionConfig layering, `index.css` in setup, reduced-motion across app/tests/Playwright) is the kind of work that pays back on every future UI PR.

---

### Convergence — issues flagged independently by 2+ perspectives

| # | Issue | Severity |
|---|-------|----------|
| **A** | `credentials.api_id` is a bare TEXT column with no FK; soft-delete preserves it. **Backend**: orphan window if re-import derives a different id. **Tests**: no assertion that operations are recreated identically after re-import — credential survives, references could silently break. | medium |
| **B** | Synchronous `urllib.request.urlopen` blocks the event loop in `preview_catalog_operations`, `preview_catalog_workflows`, `lazy_import_catalog_workflows` (`src/routers/catalog.py:814-816, 973-975, 275-276`). Tests mock `urllib.request.urlopen` directly — if implementation switches to `httpx`, tests pass without the mock firing. | high (operational) |
| **C** | Stale-after-event invalidation gaps. `useCredentialImportedSync` does NOT invalidate `['credentials']` despite `WorkflowDetailView`/`ApiDetailView` chips depending on it. `['catalog-entry', id]` (hyphen) is orphaned from `['catalog']` invalidator. | high |
| **D** | MSW mock drift. `/inspect/:id` shape already drifted once mid-PR; no schema-driven generation. `mockWorkflows` (`ui/e2e/fixtures.ts:282-289`) returns bare array even when consumer asks for envelope — latent runtime crash on the workspace redirect spec. | medium |
| **E** | Heading hierarchy + missing nav semantics. Discover jumps `h1 → h3` (`DiscoveryCard.tsx:308,410,496`); active nav tab uses colour only — no `aria-current=\"page\"` on `NavTabs` or `BottomNavbar`. Deleted page tests removed axe sweeps for those exact surfaces; new pages have no axe coverage. | medium |
| **F** | Dead code that contradicts the IA story. `DiscoveryView` `mode='sectioned'` branch unreachable in production; `DirectoryBody` function unused; `RecentlyUsedStrip` never rendered. Plus oversized files: `WorkspaceView` 695 LOC, `DiscoveryCard` 615 LOC, `ImportSourceDialog` 584 LOC. | medium |
| **G** | Order-dependent backend tests (`test_workspace_lifecycle.py`, `test_cascade_delete.py`, `test_reimport_links.py`, `test_multi_source_lifecycle.py`) — each top-level `def test_*` assumes the previous ran. Will explode if `pytest-randomly` lands. The session-scoped DB also leaks BM25 noise across modules. | medium-high |

---

### Singleton findings worth addressing

**Backend**
- Fragile JSON-string equality in workflow delete (`src/routers/apis.py:1136-1139`) — `f'[\"{api_id}\"]'` second clause breaks if `api_id` ever contains a `\"`.
- `delete_api` doesn't unlink the spec file from disk after the DB commit.
- `extract_vendor` mishandles multi-part TLDs (`api.example.co.uk` → `co.uk`); pre-existing but the new dedup logic amplifies it.
- `register_openapi` (`src/routers/import_.py:193-209`) does two separate transactions — operations insert failure leaves an orphaned API row.
- `list_api_operations` page-from-offset translation (`src/routers/apis.py:907-909`) has off-by-one drift: `offset=50, limit=20` → `page=3, start=50` but page 3 starts at 40, so `_links.self` is non-reproducible.

**Frontend**
- `ApiDetailSheet.tsx:117` exact-id pin uses `limit=1` → siblings (`slack.com/openai`) page out the real `slack.com` match. **This regresses 44dbad5** — the original false-positive is fixed but a new false-negative is introduced.
- `ui/src/api/client.ts` paths skip URL encoding entirely — fine for current ids, foot-gun if any id contains `?`, `#`, whitespace, `+`, unicode. The generated client uses `encodeURI` (preserves `/`); ad-hoc paths should match.
- In-render `setState` for `apiId` reset in `SheetBody.tsx:56-63`, `useApiOperations.ts:36-44`, `DirectoryBody.tsx:47-54`. Use `key={apiId}` on the parent instead.
- `useScrollRestore` writes per scroll event with no throttle and never prunes the module-scoped map.

**Architecture**
- Legacy `/catalog` link still in `DashboardPage` quick actions.
- `components/workspace` and `components/workflows` import from `components/discovery/*` for shared primitives (`VendorIcon`, `SectionTitle`, `ApiSummary`). Primitives belong in `components/ui` or a `components/shared`.
- Hard `window.location.assign` after credential import (sledgehammer where invalidation would do).

**Tests**
- Backend correlate of the workflow-survives-soft-delete chip test is missing — `WorkflowDetailPage.test.tsx` cache-invalidation regression is protected on the frontend, not on the backend.
- `@pytest.mark.network` against Petstore + GitHub raw URLs will rot.
- Three deleted tests (`CatalogPage`, `SearchPage`, `WorkflowsPage`) removed: 500-error UX coverage, zero-results UX, axe sweeps for those surfaces. Largest accessibility-coverage regression in the PR.
- `DiscoveryView.test.tsx:166-167, :1078-1079` assert against `window.location.search` under `MemoryRouter` — tautology, false-pass.
- 1741-LOC `DiscoveryView.test.tsx` is a god-test-file. ~30 tests with 30-50 lines of `worker.use` boilerplate each.

**A11y / motion**
- `ConfirmDeleteDialog` has `aria-labelledby` but no `aria-describedby` — destructive action, cascade impact silent to screen readers.
- 12+ inline `Loader2` spinners with visible \"Loading…/Fetching spec…\" labels lack `role=\"status\"`/`aria-live`. SR users get silent visual changes.
- `SheetPrimitive` exit timer is a fixed 300ms even under reduced-motion — focus return delayed.
- `text-primary/60 → /75` migration covered 5 page-level eyebrows but missed `text-muted-foreground/80` at `text-[10px]` in `KeyboardShortcutsBar`, `RecentlyUsedStrip`, `ApiDetailSheetHeader`, `CredentialFormPage`. Inconsistent eyebrow audit.
- Three duplicate copies of `isTypingTarget` (`lib/keyboard.ts`, `PageHelp.tsx`, `DiscoveryView.tsx`) — one will drift.

**Security** — no blockers
- Path-based IDs safe by construction (DB lookups are exact-match string equality; file paths derived from sanitized regex output, not raw `api_id`).
- `rehype-sanitize` extends `defaultSchema` correctly — `javascript:`/`data:`/`script`/`iframe`/`svg` stripped. Untrusted third-party OpenAPI descriptions are safe.
- Auth posture preserved on all new endpoints; no new state-changing GETs; no new SQL injection vectors; no PII logging regressions; `_derive_involved_apis` regexes are linear.
- Two medium tightenings: `force_api_id` accepts unsanitized strings (Pydantic `pattern=` would lock it); cascade delete relies on aiosqlite's implicit deferred transaction — explicit `BEGIN`/`COMMIT` would be immune to future `isolation_level` flips.

---

### What I'd fix before merging (5 small changes, ~2 hours)

1. **`useCredentialImportedSync`** — add `'credentials'` to the invalidate list. One line.
2. **`ApiDetailSheet.tsx:117`** — bump `limit` from 1 to 20 (or add `exact=true` server-side). One line.
3. **`ConfirmDeleteDialog`** — wire `aria-describedby` to the impact summary. ~5 lines.
4. **`mockWorkflows` in `ui/e2e/fixtures.ts`** — return envelope when `page`/`limit` present, mirroring `__tests__/mocks/handlers.ts:27-37`. ~10 lines.
5. **Wrap `delete_api` in explicit `BEGIN`/`COMMIT`/`ROLLBACK`** — make atomicity contract explicit, immune to future `isolation_level` flip. ~10 lines.

### Safe to defer to follow-up issues

- Blocking-IO → `httpx.AsyncClient` migration in `catalog.py` (real perf concern; separate PR)
- FK from `credentials.api_id → apis.id` (schema migration; separate PR)
- Splitting `WorkspaceView`, `DiscoveryCard`, `ImportSourceDialog` (already tracked in #452)
- `aria-current=\"page\"` + per-page axe smoke tests on new surfaces
- Pruning unreachable `DiscoveryView mode='sectioned'` branch + unused `DirectoryBody`
- Per-`api_id` `asyncio.Lock` around concurrent catalog imports
- `extract_vendor` Public Suffix List handling
- De-dup the three copies of `isTypingTarget`

---

I'll start landing the five immediate fixes on a follow-up branch (pre-merge) and double-check each one against the live code before touching it.

PR #536, opened 2026-07-01
fix(security): resolve token scopes live from actor grants

### LINKED ISSUE #531: [Bug]: Changing scopes and agent token bad
### What happened?

jentic register
go and change scopes, add apis:write (add catalog:write or something scope for this, apis:write is to wie)
jentic catalog
import something
still get 403 scopes no apis:write

### Steps to reproduce

above

### Version / commit

main

### Deployment method

Local (venv / from source)

### OS / environment

mac

### Relevant logs

```shell

```

### Before submitting

- [x] I searched existing issues and this is not a duplicate.
- [x] I have removed any secrets, credentials, or tokens from this report.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/jentic/.github/blob/main/CODE_OF_CONDUCT.md).

ren-jentic:
## Summary

Fixes #531 — changing an agent's scopes (e.g. adding `apis:write`) had no effect on an already-minted token, which kept returning 403 until a full re-mint.

**Root cause:** Opaque agent/service-account access tokens stored a *frozen snapshot* of scopes in `access_tokens.scopes` at mint time, and the request path (`resolve_access_token`) read permissions from that snapshot rather than the live `actor_scope_grants` table. Token refresh even copied the stale snapshot forward, so a scope edit wasn't picked up for up to the refresh-token lifetime.

**Fix:** For long-lived agent/SA tokens (those with a refresh sibling), scopes are now resolved *live* from `actor_scope_grants` at request time, so grant/revoke/replace and approved `scope:grant` access requests take effect immediately — no re-mint required. The broker's raw-SQL resolver mirrors the same logic for the data-plane path.

## Security guardrails preserved

- **Ephemeral minted tokens** (`mint_task_token` → `issue_access_only`, no refresh sibling) keep their deliberately-downscoped snapshot and are **never** re-broadened to the actor's full grants (least-privilege).
- **User tokens** are untouched — their permissions do not come from `actor_scope_grants`.

## Changes

- `token_service.resolve_access_token`: live-resolve scopes from `actor_scope_grants` for long-lived agent/SA tokens.
- `broker/repos/token_resolver.py`: same logic in the broker's raw-SQL resolver.
- `RefreshTokenRepository.family_exists`: distinguishes a long-lived access+refresh pair from an ephemeral standalone mint.

## Test plan

- [x] Unit tests for the three resolver branches (long-lived → live grants; ephemeral minted → snapshot; user → snapshot).
- [x] Integration test: adding a scope grant takes effect on the existing token pair without re-mint (direct #531 regression).
- [x] Integration test: revoking a scope also takes effect immediately.
- [x] Broker integration test: long-lived token sees live grants; ephemeral minted token keeps its downscoped snapshot.
- [x] `ruff check` + `mypy` clean on changed files; full unit + arch suite green (one pre-existing arch failure in `admin/services/auth_service.py`, unrelated to this change).

> Note: the broker's `CachedTokenValidator` caches resolved identities in-memory for a short TTL, so broker-side scope changes propagate after that TTL rather than instantly. Happy to invalidate on scope change in a follow-up if desired.

Please use **squash + merge**.

### DIFF
--- a/src/jentic_one/broker/repos/token_resolver.py
+++ b/src/jentic_one/broker/repos/token_resolver.py
@@ -6,7 +6,7 @@ import hashlib
 import json
 from datetime import UTC, datetime
 
-from sqlalchemy import text
+from sqlalchemy import Boolean, text
 
 from jentic_one.shared.auth.identity import Identity
 from jentic_one.shared.db import DatabaseSession
@@ -36,7 +36,7 @@ class InProcessTokenResolver:
             "SELECT actor_id, actor_type, scopes, is_ephemeral, expires_at, revoked_at"
             " FROM access_tokens"
             " WHERE token_hash = :token_hash"
-        )
+        ).columns(is_ephemeral=Boolean)
         async with self._admin_db.session() as session:
             result = await session.execute(stmt, {"token_hash": token_hash})
             row = result.one_or_none()
@@ -50,7 +50,7 @@ class InProcessTokenResolver:
             # from actor_scope_grants so scope edits take effect immediately.
             # Ephemeral minted tokens keep their downscoped snapshot; user tokens
             # do not draw scopes from actor_scope_grants.
-            if not _as_bool(row.is_ephemeral) and row.actor_type in (
+            if not row.is_ephemeral and row.actor_type in (
                 ActorType.AGENT.value,
                 ActorType.SERVICE_ACCOUNT.value,
             ):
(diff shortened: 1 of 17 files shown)


src/jentic_one/broker/repos/token_resolver.py
```diff
@@ -67,6 +86,17 @@ def _as_aware_datetime(value: datetime | str) -> datetime:
...
 
+def _as_bool(value: object) -> bool:
+    """Coerce a DB boolean column to ``bool`` (SQLite returns 0/1 integers)."""
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, int):
+        return value != 0
+    if isinstance(value, str):
+        return value.strip().lower() in {"1", "true", "t"}
+    return bool(value)
```
ren-jentic: dont want this here, sqllite type processor

PR #542, opened 2026-07-01
fix(search): include active IMPORTED revisions in lexical search

ren-jentic:
## Summary

Catalog imports were invisible to search. Both the Postgres and SQLite lexical
search strategies hard-filtered to `state == PUBLISHED`, but `jentic catalog
import` creates revisions in the `IMPORTED` state (`origin` set) and nothing
auto-promotes them. Everywhere else in the system (the
`ix_api_revisions_one_active` unique index, `is_current`) treats `IMPORTED` as
active — search was the lone exception, so a `jentic catalog import` followed by
a search returned empty results.

The `tsvector`/FTS wiring itself was correct; this is purely a semantic
state-filter fix.

- Widen both strategies' state filter from `PUBLISHED` to `{PUBLISHED, IMPORTED}`
  in the pinned and unpinned branches (`postgres_lexical.py` uses a shared
  `_ACTIVE_STATES` tuple with `.in_()`; `sqlite_lexical.py` uses `ar.state IN (…)`
  and drops the single `published_state` bind).
- Update both strategy docstrings from "published-state" to "active-state
  (published or imported)".
- Add a real-DB (SQLite) integration test that drives the actual `Ingestor` +
  `SearchService` and asserts an `IMPORTED` catalog revision is searchable while
  `DRAFT` and `ARCHIVED` are excluded.

## Test plan

- [x] `make lint` (ruff + format + mypy) clean
- [x] `make test-unit` — 1789 passed
- [x] `make test-arch` — 128 passed
- [x] `tests/integration/registry` on SQLite — 93 passed
- [x] search unit tests — 33 passed
- [ ] Postgres integration path (covered by CI; not run locally)

Please **squash + merge**.

### DIFF
--- a/src/jentic_one/registry/repos/search/postgres_lexical.py
+++ b/src/jentic_one/registry/repos/search/postgres_lexical.py
@@ -2,8 +2,9 @@
 
 Uses PostgreSQL's native full-text search: ``to_tsvector`` over the operation's
 ``search_text`` matched against ``websearch_to_tsquery`` and ranked with
-``ts_rank_cd``. Applies the same revision-pin and published-state filters as the
-other strategies and paginates with a keyset cursor on ``(distance, id)``.
+``ts_rank_cd``. Applies the same revision-pin and active-state (published or
+imported) filters as the other strategies and paginates with a keyset cursor on
+``(distance, id)``.
 
 ``ts_rank_cd`` is unbounded and always non-negative. We squash it into ``[0, 1)``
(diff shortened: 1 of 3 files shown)


ren-jentic (commented):
I have reviewed the changes. Widening the state filter to include `IMPORTED` revisions is necessary and completely correct for the catalog workflow, as these revisions don't get auto-promoted but should still be discoverable. The implementation across both PostgreSQL and SQLite lexical strategies is consistent, and including an end-to-end integration test to verify the state filter handles `IMPORTED`, `DRAFT`, and `ARCHIVED` correctly gives high confidence. LGTM!

PR #557, opened 2026-07-04
Add Slack feedback notifications + structured issue templates

sophie-jentic:
## What

Enables community feedback on the public repo: newly opened issues are posted to Slack, and contributors file them through structured, pre-labelled templates.

### Slack workflow
- `.github/workflows/notify-new-feedback.yml` — posts every newly opened issue to Slack
- Block Kit message with title, summary, per-label emoji, and author
- Header emoji reflects severity (`severity:blocker` 🚨 / `severity:major` ⚠️)
- No-ops when the webhook is not configured, so forks / dry runs don't fail

### Issue templates
Replaces the generic `bug_report` / `feature_request` templates with a structured set:

| Template | Emoji | Title prefix | Prefilled labels |
|---|---|---|---|
| Pain point | 🤕 | `[pain-point]` | `pain-point`, `needs-triage` |
| Bug | 🐛 | `[bug]` | `bug`, `needs-triage` |
| Feature | ✨ | `[feature]` | `enhancement`, `needs-triage` |
| Idea / Suggestion | 💡 | `[idea]` | `idea`, `needs-triage` |
| Docs | 📖 | `[docs]` | `documentation`, `area:docs`, `needs-triage` |

### Labels
Uses the repo's existing default type labels as the **main** labels, with `area:*` and `severity:*` as **sub-labels**:

- **Type (main):** `bug`, `enhancement`, `documentation` (existing defaults) plus new `pain-point` and `idea` (no default equivalent)
- **`area:*`** (cli, broker, install, control, registry, admin, auth, ui, api, deploy, docs) — added; renamed from "surface" for clarity, with `area:auth` covering `src/jentic_one/auth`
- **`severity:*`** (blocker, major, minor) — added
- `needs-triage`, `confirmed`, `agent-filed` — added

The interim `type:*` labels were **removed** in favour of the default `bug` / `enhancement` / `documentation` labels, so there's one coherent taxonomy rather than duplicated type labels. None of the `type:*` labels were in use, so no existing issues were affected.

## Notes

No secrets or credentials are hardcoded in the workflow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #560, opened 2026-07-08
ci(github): harness-driven issue intake with lean templates and labels-as-code

Manuel-Jentic:
## Summary

Reshapes issue intake around the **external issue-intake harness** so that **no human has to triage**: contributors file freely through two lean forms, and an automated intake agent classifies, scores, prioritizes, de-dupes, and loops the author back in — in a single pass.

Builds directly on **@sophie-jentic's** work in #557 (feedback Slack notifications + structured issue templates + contributor docs). That PR is the foundation here; this evolves it toward the harness-driven model rather than replacing the idea. Sophie is credited as co-author on the commit, and her Slack workflow is carried forward (with the review fixes from #557 applied).

## How it works (the 30-second version)

A contributor files an issue → it's automatically labelled `ai-intake` → the harness does **one pass** (classify, score, prioritize, de-dupe, ask the author for anything missing) → the issue ends up either **ready for a maintainer** or **waiting on the author**. Nobody hand-triages, and **the bot never closes anything** — that stays human.

```mermaid
flowchart TD
    A["👤 Contributor opens an issue<br/>(bug.yml or request.yml)"] --> B{{"auto-label: ai-intake<br/>+ needs-triage"}}
    B --> C["🤖 Harness intake — one pass"]
    C --> D["Classify: type + area"]
    C --> E["Score: fit + feasibility"]
    C --> F["Prioritize: severity"]
    C --> G["De-dupe: search existing"]
    D --> H{"Confident &<br/>info complete?"}
    E --> H
    F --> H
    G --> H
    H -- "yes" --> I["✅ Apply labels, remove needs-triage<br/>Post intake summary comment"]
    H -- "missing info" --> J["💬 Comment @author + label needs-info"]
    H -- "ambiguous / risky" --> K["🙋 Label needs-human (escalate)"]
    I --> L["🧑‍🔧 Maintainer picks it up<br/>(only humans close)"]
    J --> M["👤 Author replies"]
    M -->|"auto-clear needs-info"| C
    K --> L
```

### The single intake pass (what the harness decides)

Each step applies **labels only** — it never closes or locks an issue. When the harness isn't confident, it downgrades to a `:med` label or escalates `needs-human` instead of guessing.

```mermaid
flowchart LR
    subgraph IN["Issue text"]
        T["title + body"]
    end
    subgraph PASS["Harness pass (labels are advisory)"]
        direction TB
        C1["1 · Classify<br/>type + area:*"]
        C2["2 · Score<br/>fit:* + feasibility:*"]
        C3["3 · Prioritize<br/>severity:*"]
        C4["4 · De-dupe<br/>duplicate-candidate (suggest)"]
        C5["5 · Author loop<br/>needs-info (non-blocking)"]
        C6["6 · Hand-off<br/>remove needs-triage / needs-human"]
        C1 --> C2 --> C3 --> C4 --> C5 --> C6
    end
    subgraph RUBRIC["Sources of truth"]
        R1["docs/product-scope.md<br/>(fit rubric)"]
        R2[".github/labels.yml<br/>(allowed labels)"]
    end
    T --> C1
    R1 -.-> C2
    R2 -.-> C1
    C4 -.->|"human reply vetoes"| C4
```

### Labels are four independent layers

An issue gets **at most one label per layer**. This is the layering you asked for — type, area, and priority are separate axes, and the harness adds its own advisory score layer.

```mermaid
flowchart TB
    ISSUE["📄 One issue"] --> TYPE["**Type** (what)<br/>bug · pain-point · enhancement · idea · documentation · question"]
    ISSUE --> AREA["**Area** (where)<br/>area:broker · area:cli · area:ui · area:auth · …"]
    ISSUE --> SEV["**Severity = priority** (how bad)<br/>severity:blocker · major · minor"]
    ISSUE --> SCORE["**Harness score** (advisory)<br/>fit:high/med/low · feasibility:high/med/low"]
```

## What's here

- **Two lean issue forms** — `bug.yml` + `request.yml` (feature / idea / pain-point / docs all funnel through `request`; the harness sub-classifies). Only the first field is required. Restores the **Code of Conduct** + **secrets** + **duplicate-search** acknowledgements and the **SECURITY.md** disclosure pointer.
- **Labels-as-code** — `.github/labels.yml` (existing taxonomy + new `fit:*`, `feasibility:*`, `needs-info`, `duplicate-candidate`, `ai-intake`), synced by `labels-sync.yml`. Fixes the earlier concern that the taxonomy lived only in repo settings.
- **Harness brain** — `.harness/ISSUE_INTAKE_STANDARDS.md` (same layering pattern as `GARBAGE_COL_STANDARDS.md`): one pass to **classify** (type + `area:*`), **score** product-fit + feasibility, **prioritize** (`severity:*` == priority), **de-dupe** (suggest `duplicate-candidate`, never auto-close), **@-mention the author** for missing info via `needs-info` (non-blocking), and escalate `needs-human` when unsure. Starts at a conservative authority tier.
- **Trigger** — `issue-intake-label.yml` applies `ai-intake` on every opened issue (the label the harness watches; the forms also request it).
- **Author-loop auto-clear** — `issue-intake-followup.yml` removes `needs-info` / `duplicate-candidate` when the original author replies.
- **Product-fit rubric** — `docs/product-scope.md` **(DRAFT)** gives the harness something authoritative to score fit against: in-scope surfaces, **non-goals**, and principles. `TODO(product)` marks every spot a human must confirm.
- **Docs** — `CONTRIBUTING.md` / `AGENTS.md` updated for the two-template, automated-intake flow.

## Design decisions baked in

- **Fit vs feasibility vs severity are independent axes.**
- **Author stays in the loop** — the harness asks for missing detail rather than hard-requiring fields.
- **Triage, don't close** — the harness never closes or locks an issue at any tier; duplicates are *suggested* (`duplicate-candidate`) and the author can veto. (Informed by the OpenHands / VS Code / Kubernetes Prow prior art and the documented claude-code triage backlash.)
- **Confidence gates action** — decisive labels only at high confidence with a criteria/disqualifiers rationale; otherwise `:med` or `needs-human`.
- **Conservative rollout** — the steering note defines `comment-only` → `assist` → `auto` authority tiers; a human ratifies until the classifier earns trust.

## Open questions for reviewers

1. `docs/product-scope.md` is a DRAFT inferred from README/docs — the `TODO(product)` items (especially **non-goals**) need a product owner's confirmation before fit scores should be trusted.
2. Which authority tier should the harness start at?
3. Confirm the `ai-intake` trigger label matches what the harness platform actually watches for.
4. Confirm `@jentic/staff` is the right team to authorize for restricted labels (see Setup below).

## ⚙️ Setup required before this fully enforces

- **`ORG_READ_TOKEN` secret (required)** — the label-guard workflow (`label-guard.yml`) restricts harness-managed labels to the `jentic-harness` bot + `@jentic/staff`. Checking team membership needs a token with `read:org`, which the default `GITHUB_TOKEN` does **not** have. Add a repo secret **`ORG_READ_TOKEN`** (a PAT or GitHub App token with `read:org`). The guard **fails closed**: if the secret is missing/expired it removes the unverifiable restricted label and fails the run (red CI), so label security can't silently lapse. Trusted automation (harness bot, `github-actions`) is exempted before the token check, so this never blocks our own workflows — but the secret must be set for member label application to work.
- **Labels sync** — merge triggers `labels-sync.yml` to provision the taxonomy in `.github/labels.yml` (incl. `duplicate-candidate`, `fit:*`, `feasibility:*`, `needs-info`). Pruning is off by default.
- **Harness wiring** — the harness platform must actually consume the `ai-intake` label + `.harness/ISSUE_INTAKE_STANDARDS.md` (repo-side contract is complete here).

Supersedes the templates in #557 while preserving its Slack workflow and contributor-docs intent.

## 🔒 Security model & follow-up

**Intake is safe to run for all authors — including external/anonymous ones — because it is tightly constrained**, not because it's gated. The steering note pins this down: issue content is treated as untrusted data (prompt-injection guard), the harness may only apply **allow-listed labels** + post its single comment + search for duplicates, it never runs code / fetches URLs / reads secrets / touches PRs, and it runs least-privilege (`issues: write`, single repo). All its actions are reversible and low-blast-radius.

**Deliberately out of scope here (separate follow-up PR):** the **privileged** harness jobs — `ai-implement`, `ai-review`, and other code-executing agents — are a different risk class (they run code and open PRs). Those must additionally **gate on author association** (members/collaborators only, via GitHub's unspoofable `author_association`) and harden against prompt injection before acting. That belongs with those workflows, not with intake.

## ✅ Before this works — merge checklist

**Merging this PR does not, by itself, make automated intake function.** The classifier (the harness) runs outside this repo and must be wired up, and two secrets are prerequisites. What happens at each stage:

**Works immediately on merge (no setup):**
- [x] The two issue forms (Bug / Request) go live.
- [x] `issue-intake-label` tags every new issue `ai-intake` + `needs-triage`.
- [x] The guard / notify / followup workflows deploy (valid, but inert until the pieces below).

**Required before intake actually functions:**
- [ ] **Wire the external harness to the `ai-intake` label** using `.harness/ISSUE_INTAKE_STANDARDS.md`. Until this exists, issues just get `ai-intake` + `needs-triage` and **sit there** — no classification, scoring, or Slack. *(This is the core engine and lives outside this repo.)*
- [ ] **Add the `ORG_READ_TOKEN` repo secret** (`read:org`). `label-guard` **fails closed** — without this secret it removes any restricted label a non-member applies **and red-fails the run**. So this is a prerequisite, not optional: absent it, the guard is disruptive, not merely inert.
- [ ] **Confirm the harness's bot login is `jentic-harness` (or `jentic-harness[bot]`)**. If it authenticates as any other identity, `intake-output-guard` silently never runs and its enforcement disappears. Keep the login in sync with the guards.
- [ ] **Run `labels-sync` once via `workflow_dispatch`** right after merge, to provision the taxonomy immediately (avoids the first-merge window where template labels may not exist yet).

**Optional:**
- [x] **Add the `SLACK_WEBHOOK_URL` secret** to enable Slack notifications. Without it, `notify-new-feedback` safely no-ops (no Slack, no failure).

**Still owned by the harness platform (not in this repo), recommended follow-ups:**
- [ ] Run the harness under a **least-privilege identity** (GitHub App, this repo only, `Issues: read & write`, no secrets/private-repo reach).
- [ ] **Sanitize + datamark** issue text before the model; optionally add an injection classifier that routes suspicious issues to `needs-human`.


Manuel-Jentic:
### Why does `bug` get a label but `request` doesn't?

Good question — it's intentional, but it's a compromise worth explaining.

**The short version:** `bug` is a *type*; `request` is a *funnel*.

- **`bug.yml`** maps 1:1 to a single type, so it pre-applies `labels: ["bug", "needs-triage", "ai-intake"]`. A bug is self-evidently a bug → label it immediately (the "fast path").
- **`request.yml`** is **not** a type — it's the entry point for four different types (`enhancement`, `idea`, `pain-point`, `documentation`). The form can't know which one it is at filing time; that's precisely what we want the harness to decide. So it deliberately applies **no type label** — just `needs-triage` + `ai-intake` — and the harness reads the body and picks the right one (the "classify-later path").

So the rule is: **obvious type → label now; ambiguous type → let the harness sub-classify.**

**Honest caveats (so nobody thinks it's an accident):**
1. There is intentionally **no `request` label** — "request" is a UX word for the form, not a taxonomy label. That's why you won't find it in `labels.yml`.
2. There's a minor asymmetry: before the harness runs, a bug shows a type and a request shows none. It self-corrects within a minute once intake runs.
3. Even the `bug` fast-path label isn't final — the harness may reclassify (e.g. a "bug" that's really a `pain-point`).

**Alternative we considered:** make it fully symmetric by letting the harness assign *all* type labels (both forms seed only `needs-triage`/`ai-intake`). That's cleaner conceptually and removes the special case — the tradeoff is losing the "bugs are obvious, label them instantly" fast path. Happy to switch to that if reviewers prefer the consistency over the fast path.

Manuel-Jentic:
### Restricting labels to certain Jentic members

GitHub can't do per-label permissions natively — the built-in **Triage** role is all-or-nothing (any member with it can apply *any* label, and outside contributors can't apply labels at all). So restricting *specific* labels to *specific* people is done at the workflow layer, the same way Kubernetes Prow does its "restricted labels."

**What `label-guard.yml` does:** on `issues.labeled`, if the actor is **neither the `jentic-harness` bot nor an active member of `@jentic/staff`**, it removes the label and posts a short note. Restricted set = all harness-managed labels — type (`bug`, `enhancement`, `pain-point`, `idea`, `documentation`, `question`), `area:*`, `severity:*`, `fit:*`, `feasibility:*` — plus the human-authority status labels (`confirmed`, `needs-human`, `duplicate`, `wontfix`, `invalid`). Left **open** so automation keeps working: `needs-triage`, `ai-intake`, `needs-info`, `duplicate-candidate`.

**Two honest limitations:**
1. **Needs a token.** Reading `@jentic/staff` membership requires `read:org`, which the default `GITHUB_TOKEN` lacks. Add a repo secret **`ORG_READ_TOKEN`** (PAT or GitHub App token with `read:org`). **Until it's set, the guard no-ops with a warning** — it won't strip labels it can't authorize, so it never misfires before it's configured, but it also isn't enforcing yet.
2. **It's reactive, not preventive.** GitHub can't block the initial apply, so an unauthorized label briefly appears and is then removed within seconds. Same limitation Prow has.

**Side effect (a good one):** to avoid the guard stripping a template-applied label, `bug.yml` no longer pre-applies the `bug` label. Both forms now seed only `needs-triage` + `ai-intake`, and the harness assigns *every* type — which also resolves the earlier `bug`-vs-`request` asymmetry (they're now symmetric).

ren-jentic:
The symmetric shift to let the harness assign all type labels (including bugs) is the right move. The templates look great.

However, I have two pieces of feedback before we merge:

1. **`label-guard.yml` must fail closed.** If `ORG_READ_TOKEN` is missing, expired, or revoked, the script currently logs a warning and exits `0`. This means if our token expires in 6 months, our label security silently disappears and CI stays green. Please change the `[ -z "${ORG_READ_TOKEN:-}" ]` check to `exit 1` so we get a loud, red CI failure if the infrastructure breaks.

2. **Product Scope Dependency.** Just calling out that this entire intake system relies heavily on `docs/product-scope.md` being accurate. Since it's currently a draft with `TODO`s, we need to ensure Product ratifies it quickly, as the AI's `fit:*` scoring will only be as good as the boundaries defined in that document.

Manuel-Jentic:
### 🔒 Prompt-injection hardening (defense-in-depth)

Ran a multi-agent audit + red-team of the intake surface (workflow script-injection audit, steering-note hardening, defense-in-depth design, an adversarial red-team, and a 2026 best-practices pass). Consensus: **the steering note's prose is a mitigation, not a boundary** — real safety comes from mechanical controls that hold even if the model is fully injected. This commit adds those and closes the gaps the red-team found.

**Mechanical boundaries (outside the model):**
- **`intake-output-guard.yml` (new)** — deterministically reverts any harness-bot action outside the allow-list: a label not in `labels.yml`, a forbidden label (`security`/`confirmed`/`duplicate`/`wontfix`/`invalid`/`ai-*`), closing/locking the issue, or an unmarked/duplicate intake comment. Turns "the model must behave" into "the model can't leave damage behind." Reactive (reverts within seconds), same model as Prow restricted-labels.
- **`label-guard.yml`** — fixed a **fail-open** bug (unquoted `$RESTRICTED` glob could be filename-expanded → patterns silently dropped; wrapped in `set -f`). Added the **privileged AI trigger labels** (`ai-implement`/`ai-review`/`ai-review-and-trigger-implement`) + `security` to `RESTRICTED`, closing the worst chain (anonymous issue → label → code-executing agent). Exempted `github-actions`, and validate the actor login before the API path.
- **`notify-new-feedback.yml`** — render untrusted title/body as `plain_text`, not `mrkdwn`, so an attacker can't inject a clickable link/markup into the Slack channel.

**Steering-note hardening (`ISSUE_INTAKE_STANDARDS.md`):**
- "untrusted payload" framing + a concrete jailbreak-refusal list (instruction override, **authority-spoofing**, fake tool calls, homoglyph/zero-width/HTML-comment obfuscation, mention/label/close manipulation, exfiltration);
- a **provenance self-check** before any action + **fail-closed to `needs-human`** on detected attacks;
- **never echo the issue verbatim** into the comment (it's read by downstream `ai-implement`/`ai-review` agents — this breaks the second-order injection chain);
- only ever @-mention the **verified event author**, never a body-supplied handle.

**Platform-side (documented, must be enforced by the harness runtime — tracked, not in this repo):**
- least-privilege identity (ideally a GitHub App on this repo only, `Issues: read & write`, short-lived token, **no secrets/private-repo reach** — the "Agents Rule of Two");
- input **sanitization + datamarking** (strip zero-width/bidi/Unicode-Tag/variation-selectors, NFKC, homoglyph-fold, length-cap) before the model;
- optional **injection classifier** routing suspicious issues to `needs-human`.

Real-world exploits that motivated this (all documented): the Gemini-CLI `gh issue edit` data-leak (CVSS 10), GitLost (GitHub AI agent leaking private repos), and the Claude-Code-Action `[bot]`-trust / `/proc/self/environ` bugs. The throughline: keep the agent at *untrusted-input + state-change* and give it **no** access to anything worth stealing.

> Note: `intake-output-guard.yml` and `label-guard.yml` are reactive (the bad action exists briefly, then is reverted) — inherent to GitHub's model, same as Prow. `label-guard` also needs the `ORG_READ_TOKEN` secret to enforce (documented in the PR body).

Manuel-Jentic:
Thanks @ren-jentic — both good calls.

**1. label-guard now fails closed.** ✅ Fixed in d78593f0. A missing/expired/revoked `ORG_READ_TOKEN` no longer warns-and-`exit 0`; it now **removes the unverifiable restricted label and `exit 1`s** (loud red CI), so label security can't silently vanish if the token lapses. Trusted automation (the harness bot, `github-actions`) is exempted *before* the token check, so a lapsed token never blocks our own workflows — it only affects member label application, which is exactly what should break loudly. Updated the workflow header and the PR's setup section to match.

**2. product-scope dependency.** Agreed, and flagged as open question #1 in the PR — `docs/product-scope.md` is a DRAFT with `TODO(product)` markers, and the steering note already caps `fit:*` at `fit:med` when a decision depends on an unresolved TODO, so the AI won't assert confident fit scores against undefined boundaries until Product ratifies it. I'll leave the ratification itself to Product; the harness degrades gracefully until then.

Manuel-Jentic:
### Full review pass (7 lenses) + fixes — `3d14b3d1`

Reviewed the final state across security, YAML/Actions correctness, harness steering-note logic, docs/DX, product-scope quality, end-to-end flow coherence, and a red-team. **No security blockers** — the mechanics (env-var injection discipline, least-privilege perms, fail-closed label-guard, allow-list parsing, Slack plain_text) all held up. The real findings were lifecycle + a load-bearing identity assumption. All fixed in `3d14b3d1`:

**Critical (lifecycle — the flow didn't complete):**
- **Author-loop dead-end** → `issue-intake-followup` now re-adds `ai-intake` when it clears `needs-info`, so the harness re-runs after the author answers instead of the issue sitting stranded.
- **Slack never fired at the default `assist` tier** (it keyed on `needs-triage` removal, which `assist` leaves for a human) → `notify` now fires on `ai-intake` removal, the harness's universal "pass complete" signal at every tier.

**High (guard identity):**
- Both guards assumed the harness login is exactly `jentic-harness`; a GitHub App is `jentic-harness[bot]`, which silently disabled enforcement. Both guards now match either form; `label-guard`'s actor `case` quoted the `[bot]` glob (was a char-class), strips `[bot]` for the org lookup, and no longer flags bot logins as malformed; `intake-output-guard` dedups by comment marker alone (identity-agnostic). Documented the invoker is out-of-repo and the login must stay in sync.

**Consistency/polish:**
- Steering note: removed stale "product-scope is a DRAFT with TODO(product)"; added `needs-human` to the assist tier + escalation-at-every-tier note; made the never-apply label set exhaustive and matched to the guards; paired `fit:med` with `needs-human`; added non-human-author handling.
- `notify`: `$NUMBER` via `--arg` (was spliced into the jq program).
- `labels-sync`: documented the first-merge label-provisioning race + manual run.
- CONTRIBUTING: mention the submit checkboxes; scoped "prioritizes (severity)" to bugs/pain-points.

Validated: all YAML parses; all four workflow run-blocks pass `bash -n`; lint/mypy clean.

**Still owned by the platform (not fixable in this repo), tracked for follow-up:** the least-privilege GitHub App identity the harness authenticates as, input sanitization/datamarking before the model, and wiring the external invoker to the `ai-intake` label.

sophie-jentic (approved):
LGTM. Thank you!

PR #561, opened 2026-07-08
docs(release): propose release & distribution procedure

Manuel-Jentic:
## Summary

A **proposal for review** (doc only, no automation wired) for jentic-one's release procedure — versioning, tagging, release notes, and PR-to-publish. Adds `docs/plans/release-procedure.md`.

### Why now
jentic-one **inherited jentic-mini's full semantic-release → tag → GitHub Release → GHCR publish pipeline** (it reached `v0.13.2`), but the **2026-07-01 OSS scrub deleted all of it** and reset the version to `0.1.0`. Today there is **no versioning / tagging / release / changelog automation** — only CI. Conventional Commits + squash-merge are still enforced, but nothing consumes them, and the version has **drifted** (`0.1.0` in files vs the `v0.13.2` tag line). The git-conventions rule still references an "auto-generated changelog" that no longer exists.

### What the doc contains
- **Current state** (facts): version sources, the `0.1.0`↔`v0.13.2` drift, `backup/*` tag pollution, commented-out GHCR publish, no `CHANGELOG.md`.
- **What we did before** (jentic-mini's semantic-release stack, still live in the sibling checkout).
- **OSS benchmark**: release-please vs semantic-release vs changesets vs git-cliff; GoReleaser; Docker + Helm-to-GHCR OCI; pre-1.0 conventions (Vector's `VERSIONING.md`).
- **Recommended procedure**: **release-please** (manifest, single lockstep version, `bump-minor-pre-major`) → **GoReleaser + Docker + Helm-to-GHCR on the tag**, bridged by a **GitHub App token** (like mini's `ARAZZO_BUILDER_APP_ID`); categorized release notes + `CHANGELOG.md`; a `VERSIONING.md`.
- **End-to-end flow diagram**, **open decisions**, and a **suggested implementation order**.

### Key recommendation (needs ratification)
**Continue the 0.x line — first OSS release `v0.14.0`** — rather than reset to `0.1.0` or jump to `1.0.0`:
- `1.0.0` contradicts the README's "public beta, breaking changes without a major bump, not production-recommended."
- `0.1.0` misrepresents the mature codebase **and collides with the existing `v0.1.0`…`v0.13.2` tags** (release-please could compute an already-used tag).
- Continuing `0.x` is honest, collision-free, and matches the beta breaking-change policy.

### Open decisions for reviewers (§6 in the doc)
1. Version baseline: continue `0.x` (recommended) / clean `v1.0.0` / fresh `0.1.0` lineage.
2. Lockstep vs independent versioning for the Go CLI.
3. GitHub App token: reuse mini's or provision new.
4. Chart publish: GHCR OCI (recommended) vs `gh-pages`.
5. `backup/*` tags: delete vs relocate.

No code/automation changes — this is the plan to agree on before implementing.


Manuel-Jentic:
## Revised after a multi-lens review (facts re-verified against the repo)

Pushed `83210909` amending the proposal. Every finding below was **verified directly** against the tree before amending.

### Factual corrections
- **Version is `0.1.1`, not `0.1.0`.** Drift is three-way: `pyproject.toml 0.1.1` vs all `Chart.yaml 0.1.0` vs the `v0.13.2` tag line.
- **No tag-collision constraint.** The `v0.1..v0.13` + `backup/*` tags exist **only locally — `git ls-remote --tags origin` (and opensource/internal) returns 0 tags.** The earlier "0.1.0 would collide" argument was wrong; it's a one-time *local* cleanup that applies to all baseline options.
- **`2654c92c` added `pyproject.toml` + all charts fresh at `0.1.0`** on rewritten history — it was not a `0.13.2 → 0.1.0` edit.
- **Helm facts:** 9 charts, only the umbrella + observability have `appVersion`; the image tag resolves from the umbrella `appVersion`/`global.image.tag`, not subchart appVersions.

### Technical must-fixes now called out
- release-please won't rewrite the umbrella `dependencies:` version pins or the `file://` `common` pins → **breaks `helm dependency build`** unless loosened / handled via `extra-files`.
- The Go CLI version is a build-time ldflag placeholder → **GoReleaser stamps it from the tag**; it doesn't belong in the manifest.
- **Publish only the umbrella + observability**, not the subcharts / library chart.
- **Seed the manifest + `bootstrap-sha`** so the first changelog doesn't replay pre-scrub history; add a **`uv.lock`** updater.
- `on: release` does **not** bypass the `GITHUB_TOKEN` downstream-trigger limit → use the App token.

### New sections
- **Distribution disconnect (§4.7) — gates the whole plan:** `install.sh`/`jenticctl` build from source at a git ref and never pull `ghcr.io/...:X.Y.Z`, so a released image/chart is currently orphaned.
- **Upgrade/migrations (§4.8):** ~65 forward-only migrations → one-minor-at-a-time policy, mandatory pre-upgrade backup, rollback = restore + pin, a hotfix branch flow.
- **Supply chain (§4.4/§4.9):** image cosign signing + SBOM + provenance + a **CI-on-tags** build/migrate/health smoke gate (tags get no CI today), release authority + branch protection on the Release PR.
- **CLI-vs-server changelog**, `deploy/README.md` versioning cross-reference, and a decision on where the ratified procedure lives.

### Baseline
With the collision constraint gone, **clean `v0.1.0`** (most honest for a new public repo) and **continuing `0.x` → `v0.14.0`** (continuity, needs an explanatory note) are both viable. `1.0.0` stays excluded while the README allows breaking changes without a major bump. Left as open decision #1.

Manuel-Jentic:
## Reframed after pushback on distribution strategy

Digging into how the product actually installs revealed the earlier "publish images + Helm to GHCR" framing was solving a problem no install path has yet — **everything builds from source today; nothing pulls a registry artifact.** Also surfaced: the `jentic` CLI is **already a standalone remote client** (separate Go module, targets a remote via `--base-url`/broker flags, documented in `hardening.md`). Reworked the plan accordingly (backed by research on Sentry/PostHog/Supabase/Meilisearch + a repo audit).

### Distribution: prebuilt-first, not source-first
Nobody's happy path is compile-from-source. At beta we ship the high-leverage prebuilt artifacts and defer the heavy ones:
- **Ship:** Python **wheel** (PyPI) — makes the no-Docker "venv + SQLite" install a clean `pipx`/`uvx`; **GoReleaser CLI binaries** (+ brew) — agents shouldn't need a Go toolchain; tag + notes + checksums.
- **Defer honestly:** the **Docker server image** stays build-locally for beta (documented as such, not dressed up), with a tripwire to add a signed multi-arch GHCR image + CVE-rebuild cron once a non-author asks "how do I run this without building?"
- **Defer:** **Helm OCI** publish (repo-referenced chart until real k8s users).

### CLI as a remote client (the VPC / different-host case)
The CLI already drives a remote jentic-one. So it's a *client*, kubectl-style. For beta: **keep lockstep versioning** (simple) **but publish the CLI as standalone binaries**, and add now: a **CLI↔server version-skew warning** and **remote-client UX** (single instance URL that derives control-plane + broker, plus a `JENTIC_BASE_URL` env var). Full release-train decoupling is deferred behind a tripwire.

### Update notifications (firm requirement)
One-way GitHub Releases check surfaced in **CLI + UI**, covering both "newer release available" **and** "CLI vs remote-server skew." Pull-only (sends no data), default-on/one-flag-off, **separate from opt-in telemetry**, `--offline` silences all. Builds on `/health`'s version field; adds a Go semver lib (none today — current `update` compares git SHAs, not tags).

Decisions 2 & 3 in the doc are now settled per the above; **version baseline (#1)** and **app token (#4)** remain open. Doc kept skimmable (diagram + tables; detail in foldouts).

ren-jentic (commented):
This proposal is incredibly thorough and accurately captures the complexities of the distribution matrix. The CLI-as-a-remote-client framing is spot on, and using `release-please` + `GoReleaser` natively solves the manual tagging drift.

However, viewing this through the lens of the recently merged **Open-Core Split Architecture (`jentic-one-internal#704`)**, there is one critical contradiction that needs addressing before ratification:

### The Conflict: Deferred Docker Publish vs Enterprise Composition
In the distribution table (Section 4.4), the proposal states:
> **Docker server image (GHCR): Defer — honestly.** Keep "build locally" for beta... Tripwire: the first "how do I run this without building?" from a non-author

However, the Open-Core spec explicitly relies on a published OSS image for the Enterprise tier to function:
> **Enterprise Docker Composition:** Enterprise does not dynamically mutate the OSS image at runtime. It maintains its own Dockerfile **using the released OSS image** (the output of `deploy/docker/app.Dockerfile`) **as its `FROM` base**. On top of this base, it installs the Enterprise Python wheel and copies in the compiled `jenticctl-enterprise` binaries.

If the OSS pipeline does not publish the Docker image to GHCR upon tagging, the private `jentic-one-enterprise` repository has no base image to pull for its `x.y.z` pin. The Enterprise CI would be forced to build the OSS image from source every run, defeating the "Dependency Overlay" boundary.

**Recommendation:**
The Docker GHCR publish for the server image cannot wait for an external user tripwire; it is an immediate requirement for the internal Enterprise CI. It should be moved from **Deferred** to **Ship**.

(The Helm OCI deferral is completely fine and doesn't conflict with the architecture).

Otherwise, the baseline choice (`v0.14.0`) and the remote-client UX are perfect.

Manuel-Jentic:
## Revised after a six-lens review (fact-check, maintainer, security, implementability, operator UX, clarity)

Applied the consolidated feedback. Summary of what changed:

### Factual corrections (fact-check + implementability agents)
- **Version drift is FOUR-way, not three** — `src/jentic_one/__init__.py = 0.1.0` is what `/health` actually serves, and release-please won't touch it without an `extra-files` updater. Highest-value catch (would've shipped a lying `/health`); fix = `__version__` reads package metadata.
- 7 version-only subcharts (not 6); `jentic-mini` GHCR tags reach **0.13.2** (not 0.9.x); softened "no `jentic-one` image" → "no *public* image" (unverifiable without `read:packages`); stale `broker.jentic.ai` also in `execute.go`/`install.go`; GoReleaser produces a **formula**, not a cask.

### Decisions made (were left open)
- **Version baseline → clean `v0.1.0`** (recommended independently by the maintainer + UX reviewers: 0 public releases, `v0.14.0` implies history that doesn't exist, no collision).
- **Release-trigger token → provision a NEW narrowly-scoped App** (`contents`+`pull-requests` on this repo only), not inherit mini's broader install.
- Both surfaced in a **decisions table at the very top**.

### Security (blockers reframed)
- The **live public `jentic-mini` image exposure** is pulled out of a foldout into a **P0 security doc**: rotate the enumerated secret classes + **full git-history scan across all four repos** + delete (rotation is the only real fix). → `docs/plans/security-orphaned-mini-package.md`
- **Self-update must verify cosign/checksums against a release tag** (today it rebuilds from a mutable branch with no verification — RCE risk); source-build fallback pinned to a tag + verified Go-toolchain download; CI-gate runs with **ephemeral secrets only**; least-privilege release workflow.

### Structure (maintainer + clarity agents)
- **Split the kitchen-sink doc:** `jenticctl update` rework, version-notifications/remote-UX, and the orphaned-package cleanup are now their own focused docs (linked inline).
- **441 → 170 lines**, decisions at the top, rationale folded, **beta-blocking (PR1–3) vs deferred** split, fixed the step-4/step-7 (brew-vs-`update`) inversion.
- **Moved out of `docs/plans/`** (per-issue plans) → `docs/release-procedure.md`; **retitled** `docs(plans)` → `docs(release)`.

### Operator UX
- **Upgrade safety promoted to must-ship:** backup-by-default, one-minor-at-a-time hard stop, `--dry-run`, `UPGRADING.md` with a concrete rollback recipe + support policy (data-unrecoverable migrations on a credential broker demand this).

**Agreed minimal value slice:** PR-1 fix version drift → PR-2 release-please → PR-3 GoReleaser signed CLI binaries. Everything else is deferrable and now tracked separately. The P0 `jentic-mini` exposure is independent and urgent.

Manuel-Jentic:
## Re: @ren-jentic — deferred Docker publish vs Enterprise composition (`jentic-one-internal#704`)

Thanks — this is the right lens, and it caught a real coupling the OSS-only view missed. I re-verified against **both** repos before replying. Three things, one of which narrows your objection and one of which is a genuine gap worth kicking back to the open-core spec.

### 1. You're right about the *image* — and I'll move it. But it's the image, not the wheel.
The Enterprise `FROM base` = "the released OSS image (output of `deploy/docker/app.Dockerfile`)" pinned at `x.y.z`. For `FROM ghcr.io/jentic/jentic-one/app:x.y.z` to resolve, that image must be published where Enterprise CI can pull it. Agreed — **Docker image publish moves from Defer → Ship** as an *internal* requirement (Enterprise CI), independent of the external "how do I run this without building?" tripwire.

One clarification that matters for the plan: this does **not** re-open the dropped **PyPI wheel** decision. Inside the base image the OSS app is *already* a pip-installed wheel (`python-base.Dockerfile` runs `uv build --wheel`; `app.Dockerfile` `pip install`s it). The overlay pins the OSS *package* via git tag in `go.mod`/`uv.lock`, and Enterprise step 1 installs the *Enterprise* wheel on top. So the overlay never needs a **published PyPI wheel** — only the **published image**. "No PyPI wheel install path" (§4.4) stands; only the image line changes.

### 2. One open question before we hard-commit "publish": pull vs build-from-pinned-tag.
The spec says "the output of `app.Dockerfile`" but **never names a registry**. Since the OSS CLI itself *builds the image locally* today (nothing in `cli/` ever pulls `ghcr.io`), there are two ways Enterprise CI can satisfy the `FROM`:
- **(a)** Pull a **published** `ghcr.io/jentic/jentic-one/app:x.y.z` (your reading), or
- **(b)** Build the OSS image locally from the **pinned git tag** in Enterprise CI (mirrors how the OSS CLI already works).

(a) is cleaner and I'm inclined to it — but it also means the new image's **public visibility is a deliberate, reviewed choice** (this is exactly how the `jentic-mini` orphan happened — tracked P0). Can you confirm (a) is the intended contract, i.e. Enterprise CI *pulls* rather than *builds*? If so, "Ship" it is.

### 3. Genuine gap to raise back to #704: is Enterprise deliberately Docker-only?
OSS has a first-class **no-Docker** install — `jenticctl install` → "Run locally" → `uv pip install -e` into a venv + SQLite, zero Docker, zero registry (verified: preflight requires only `uv`/`git`; nothing pulls an image). The open-core spec §5 defines Enterprise production as **Docker-only** (`FROM oss-image` + enterprise wheel + binaries + SPA); the only venv-style mechanism is the `[tool.uv.sources] editable` dev path, explicitly *"stripped/ignored in prod."*

So a no-Docker **Enterprise** customer — the paid analogue of the OSS venv+SQLite operator — has **no documented deployment path**. That's an open-core *architecture* decision (does Enterprise want a venv overlay: enterprise wheel + OSS wheel into a venv, to mirror OSS?), not something this OSS release doc should decide. Flagging it here so it lands back on #704 rather than silently becoming "Enterprise = Docker-only" by omission.

### Net
- **Image → Ship** (Enterprise-CI requirement) — updating the doc now.
- Confirm **pull vs local-build** so we scope the publish correctly (and treat public visibility as a reviewed decision).
- **No-Docker Enterprise path** is a #704 question — raising it there.
- Wheel/Helm decisions unchanged.

PR #562, opened 2026-07-08
refactor(seams): add pluggable extension points across backend, CLI, and UI

ren-jentic:
## Summary

Introduces backward-compatible **seams** so downstream consumers can inject alternate implementations and mount extra components without editing core code. When no consumer wiring is supplied, behavior is identical to today.

- **Backend**
  - Neutral `Broker` protocol + transport-neutral value objects in `shared/broker` (so the broker surface and any consumer depend on the protocol, not the concrete pipeline).
  - DI `AppContainer` for the app factories (inject a `Broker`, mount extra routers/installers after the built-in surfaces). The async worker honors an injected `broker_factory` too, so injected brokers reach both the sync and async paths.
  - Registrable config sub-models (`register_config`, with collision guards) and telemetry events (`register_telemetry_event`) without editing the core schema/enum.
  - Ordered `MigrationTarget` registry (`register_target`) as the single source of truth for migration sequencing, with reversible up/down.
- **Testing**: ships `jentic_one.testing` compliance bases (`isinstance` + `inspect.signature` checks) so implementations can prove they honor the seam contracts.
- **CLI**: exportable `pkg/core` container + root builder; `internal/cmd -> pkg/core` stays one-directional so alternate binaries can compose their own command tree.
- **UI**: renames the `src` alias `@/` -> `@oss-internal/` across the SPA and exposes the route/handler registries as append-only composition points.
- **Arch tests**: sources ORM-convention facts from `jentic-one-rules` with a committed vendored fallback, so a standalone clone still self-enforces.
- **Docs**: adds `docs/development/extending-jentic-one.md` — a unified composition guide for downstream integrators (build an `AppContainer`; `register_config` / `register_target` / `register_telemetry_event`; subclass the `jentic_one.testing` compliance bases).

## ⚠️ Breaking Change: unknown top-level config keys now fail loudly

`AppConfig` now sets `model_config = ConfigDict(extra="forbid")`.

**Behavior change:** an **unrecognized top-level config key** — one that is neither a known core field nor a *registered* extension section — now causes a **loud failure at startup** (a `ConfigError`) instead of being silently ignored.

This is defensively correct: it ensures extensions are formally registered via `register_config` (registered sections are extracted by `load_config()` before validation, so they are accepted), and it surfaces misconfiguration instead of dropping it. **Downstream users must be warned:** any config with legacy or typo'd top-level keys that was previously tolerated will now fail on boot until the offending key is removed or migrated to a registered extension section.

## Test plan

- [x] Backend unit tests for the new extension seams:
  - `tests/unit/test_config_extensions.py` — `register_config`, `registered_config_models`, `AppConfig.extension()`, collision guards (core fields + reserved `extensions`), and `load_config` extraction / `extra="forbid"` rejection of unregistered keys.
  - `tests/unit/shared/telemetry/test_event_registry.py` — `register_telemetry_event` + `resolve_wire_name` (wire-name validation, built-in enum/map collision rejection, precedence, idempotency/conflict).
  - `tests/unit/broker/test_injected_broker_seam.py` — an injected `Broker` is honored by **both** the sync path (`execute.py` selection) and the async path (`PipelineExecutor` via `run_execution`), plus `AppContainer` wiring onto `app.state.broker`.
- [x] `make test-fast` — unit + arch.
- [x] `ruff check` + `ruff format --check` clean on touched files.
- [ ] CI: lint + typecheck + score + UI lint/tests.
- Note: `test_origin_propagation[admin/services/auth_service.py]` fails on this branch **and on `main`** — pre-existing, unrelated (that file is untouched here).

## Merge

Squash + merge.


ren-jentic (commented):
This looks solid and executes perfectly on the 'Dependency Overlay with Strict Seams' directive from the architecture spec (jentic-one-internal#704).

Key takeaways:
- `AppContainer` in the Go CLI (`cli/pkg/core/container.go`) successfully models the bi-directional boundary without pulling `internal/` outward.
- Moving to `DB_TARGETS` with `register_target` instead of a static dict in Alembic resolves the Postgres schema collision danger natively.
- The `jentic_one.testing.compliance` harnesses are an elegant fix for Python's structural typing gaps (`isinstance` not validating method signatures).
- The neutral `Broker` interface decouples the execution layer properly.

A thorough follow-through on the open-core strategy.

Manuel-Jentic:
## Alternative worth considering: keep the public repo on plain `@`, push the rename into enterprise

@ren-jentic — the `@/ → @oss-internal/` rename is mechanically correct and I verified it works. But I want to float an alternative, because renaming the public SPA's own alias to `@oss-internal` bakes the private-tier vocabulary into every import line of the *public* repo (335 files reading `@oss-internal/` — a name that only makes sense if you know about the private enterprise repo, which OSS users don't have).

I tested this empirically. The key fact is that **Vite/Rollup has one flat, global alias table for the combined build**, so exactly *one* of the two repos must give up `@` — but it doesn't have to be OSS.

### Test results (reproducible script below)
```
1) OSS '@' + host also claims '@' (naive keep-@)           BUILD-FAILS (collision)
2) OSS '@oss-internal' + host keeps '@' (CURRENT PR)       PASS
3) OSS neutral '@app' + host keeps '@'                     PASS
4) OSS '@' + enterprise uses '@ent' for self, '@'→OSS      PASS   ← "keep public as @"
```

### The "keep public as `@`" option (variant 4), stress-tested
OSS stays exactly as it is today (`@ → src`, zero churn). The **enterprise** build is the one that disambiguates:
```ts
// enterprise vite.config.ts (private repo)
resolve: { alias: {
  '@oss': resolve('../jentic-one/ui/src'),  // consumer-facing OSS mount
  '@ent': resolve(__dirname, 'src'),        // enterprise's own code
  '@':    resolve('../jentic-one/ui/src'),  // OSS's own '@/' imports resolve to OSS
}}
```
I hardened this against the cases that actually bite in a real SPA, and all pass:
- **Deep transitive OSS chains** (`foo → deep/a/b/leaf → shared/marker`, all `@/`) resolve to OSS.
- **Shared singleton is NOT duplicated across the boundary**: enterprise importing an OSS store both via a re-export and directly via `@oss/shared/store` gets the *same module instance* (`sameSingleton=true`, mutable count accumulates across all call sites — no double-instantiation).
- **Standalone public OSS build** (plain `@ → src`, unchanged) still resolves the deep chain — the public repo needs **no change at all**.

The trade-off is symmetric — someone must cede `@`:
- **Current PR:** OSS cedes `@` (renames to `@oss-internal`) so enterprise keeps `@`. Cost: churn + private-tier naming in the public tree.
- **Variant 4:** enterprise cedes `@` (uses `@ent/` for itself, maps `@ → OSS`) so **OSS stays pristine `@`**. Cost: the enterprise repo has an `@ → OSS` mapping (mildly unusual, but it lives in the private repo where the split is understood).

My argument for variant 4: a cost that exists only because of the private split should be paid by the private repo, not stamped across every public import line. If you'd prefer OSS to own a distinct prefix regardless, I'd at least suggest a **neutral name (`@app/` / `@ui/`)** over `@oss-internal` — variant 3 proves it's behaviorally identical without leaking the tier vocabulary.

Not blocking — the current approach works. Just think variant 4 (or a neutral prefix) keeps the OSS repo cleaner for the thousands who only ever see the public side. What do you think?

<details><summary>Reproducible test (bash, needs node + npx vite)</summary>

```bash
# Builds a tiny OSS package + enterprise consumer and executes the bundle,
# asserting OSS internals resolve to OSS (no cross-binding) across alias schemes.
# Full script: https://github.com/jentic (attach on request) — core loop:
#   OSS: shared/marker.ts, modules/foo/index.ts (imports via a switchable prefix),
#        deep/a/b/leaf.ts (transitive), shared/store.ts (mutable singleton)
#   ENT: main.ts imports @oss/modules/foo + @oss/shared/store + @ent/... ,
#        builds with the alias block under test, then node executes dist/out.js.
# PASS = output shows OSS_MARKER, deep leaf resolved to OSS, sameSingleton=true.
```
</details>

Manuel-Jentic:
## Small hardening ask: `BaseSearchStrategyComplianceTest` should assert the strategy actually *resolves*, not just that `dialect` is a string

The `inspect.signature` checks in `jentic_one.testing.compliance` are a real upgrade over bare `runtime_checkable` — nice. One gap I hit while reviewing the enterprise overlay against this PR:

`test_has_required_attrs` only asserts `isinstance(instance.dialect, str)`. That passes for **any** string, including a wrong one. Concretely: the enterprise strategies declared `dialect = "postgresql"`, but `PostgresBackend.dialect_name` returns `"postgres"`, so `resolve_strategy((dialect_name, mode))` never finds them and raises `SearchUnsupportedError` **at runtime** — while the compliance suite stays green. A downstream implementer gets a passing test and a broken search path.

Suggestion: add a compliance check that the registered `(dialect, name)` is actually resolvable — e.g. assert `strategy_cls.dialect` is one of the known backend `dialect_name` values, or (stronger) that `resolve_strategy` returns the strategy for a backend whose `dialect_name == strategy_cls.dialect`. That turns the most likely real-world drift (a plausible-but-wrong dialect string) from a silent prod failure into a caught test failure — which is the whole point of the compliance bases.

Not blocking; the signature checks already cover the harder case. This just closes the one class of drift the current attr check waves through.

PR #566, opened 2026-07-10
docs(auth): document Ed25519 key requirement in RegisterRequest schema

juliahooper:
## Summary

- Adds a description to the `jwks` field on `RegisterRequest` specifying that keys must be Ed25519 (kty=OKP, crv=Ed25519)
- Agents previously discovered this requirement only through the error message after a failed registration attempt
- Regenerated OpenAPI spec and UI client to include the new field description

## Test plan

- [x] `ruff check` + `mypy` clean
- [x] All unit + arch tests pass (1929 passed)
- [x] OpenAPI spec and UI client regenerated


src/jentic_one/auth/web/schemas/registration.py
```diff
@@ -14,7 +14,10 @@ class RegisterRequest(BaseModel):
     """POST /register request body."""
 
     client_name: str
-    jwks: dict[str, Any]
+    jwks: dict[str, Any] = Field(
+        description="A JSON Web Key Set containing at least one Ed25519 public key"
+        " (kty=OKP, crv=Ed25519). RSA and other key types are not accepted."
+    )
```
ren-jentic, line 20: can you add an example here, something like:
```
 jwks: dict[str, Any] = Field(
            description="A JSON Web Key Set containing at least one Ed25519 public key (kty=OKP, crv=Ed25519). RSA and other key types are not accepted.",
            json_schema_extra={
                "example": {
                    "keys": [
                        {
                            "kty": "OKP",
                            "crv": "Ed25519",
                            "x": "11qYAYKxCrfVS_7TyWQHOg7hcvPapiMlrwIaaPcHURo",
                            "kid": "agent-key-1"
                        }
                    ]
                }
            }
        )
```

ren-jentic (changes requested):

src/jentic_one/auth/web/schemas/registration.py
ren-jentic: Done in 2899c604 — added the suggested `json_schema_extra` example (RFC 8037 appendix A.2 public-key test vector) and regenerated the spec-derived artifacts (`control.openapi.yaml`, `ui/openapi.json`, and the Go CLI generated client, which also fixes the `check-cli-gen` CI failure). The example value is allowlisted in `.secrets.baseline` since detect-secrets flags it as a high-entropy string.

PR #568, opened 2026-07-10
docs(access-requests): describe resource_reference as alternative to resource_id in schema

juliahooper:
## Summary

- Adds descriptions to `resource_id` and `resource_reference` fields on `AccessRequestItemRequest`
- Makes it clear that `resource_reference` with `{vendor, name, version}` can be used instead of a toolkit ID for `toolkit:bind` requests
- Agents previously only discovered this flow after hitting the broker's `no_toolkit_binding` error

## Test plan

- [x] `ruff check` + `mypy` clean
- [x] All unit + arch tests pass (1929 passed)
- [x] OpenAPI spec and UI client regenerated


src/jentic_one/control/web/schemas/access_requests.py
```diff
@@ -45,8 +45,17 @@ class AccessRequestItemRequest(BaseModel):
...
+        default=None,
+        description="Explicit ID of the resource (e.g. a toolkit tk_… or credential cred_… ID)."
+        " For toolkit:bind, you can omit this and use resource_reference instead.",
+    )
+    resource_reference: dict[str, Any] | None = Field(
+        default=None,
+        description="Look up the resource by API identity instead of by ID."
+        " For toolkit:bind, provide {vendor, name, version} to resolve the toolkit"
+        " that serves the given API. Use this when you don't know the toolkit ID.",
+    )
```
ren-jentic, line 67: Could you add an example here, something like:
```
        resource_reference: dict[str, Any] | None = Field(
            default=None,
            description="Look up the resource by API identity instead of by ID."
            " For toolkit:bind, provide {vendor, name, version} to resolve the toolkit"
            " that serves the given API. Use this when you don't know the toolkit ID.",
            json_schema_extra={
                "example": {
                    "vendor": "github",
                    "name": "issues",
                    "version": "v3"
                }
            }
        )
```

ren-jentic (changes requested):

Issue #641, opened 2026-07-15
[ui] Server-variable (region) selection on credential create is not persisted/applied — broker always routes to the spec default


ash-jentic:
### What happened

Testing an API whose spec declares a region server variable (`https://{region}.posthog.com`, enum `[us, eu]`, default `us`):

1. The credential-create form correctly renders the **Server configuration → Region** dropdown (from the spec).
2. Selected **eu**, created the credential, bound it to the toolkit (as the only credential for that API).
3. Agent executes → upstream call goes to **us.posthog.com** (verified at TLS level in broker logs: `server_hostname='us.posthog.com'`; also visible in the response's `Www-Authenticate` pointing at us.posthog.com). EU-region key → 401 "invalid key".
4. Deleted and re-created the credential, carefully re-selecting **eu** (verified selected before Create). Same result: TLS to `us.posthog.com`, 401.

### Evidence it's the form→record link, not the broker

- The string `server_variables` appears **zero times** in a DEBUG-level broker log across all attempts — the resolver appears never to have loaded any.
- The broker's execute path does apply credential `server_variables` when present (this mechanism works for other APIs, e.g. Atlassian's `{your-domain}`).
- The spec side is proven: the same variable declaration is what renders the dropdown and (with `default`) steers the broker's fallback.

So the selected value seems to be dropped between the create form and the stored credential (not sent, or ignored by the create request for this credential type).

### Impact

The region feature chain (spec variable → UI dropdown → credential → broker routing) breaks at the third link, silently. The user does everything right and still gets routed to the wrong regional host with a misleading upstream "invalid key" (feedback#76 (internal tracker)). Because the credential view/edit hides server configuration entirely (feedback#78 (internal tracker)), the user cannot even see that their selection was lost.

### Environment
Jentic One self-hosted, 2026-07-14. Credential type: Bearer token. API: PostHog with region-parameterised spec. Exec ids for the failing calls: exec_2f5064d6ae4544379cdb6974, exec_946a918e11f44d1ea658b165.

---
*Migrated from the internal dogfooding tracker (feedback#79, originally filed 2026-07-14). Screenshots/attachments referenced in the original will be re-attached.*

Issue #640, opened 2026-07-15
[ui] Credential view/edit shows only Name + Token — auth method and server configuration (e.g. region) are invisible and uneditable after creation


ash-jentic:
### What happened

After creating a credential for an API whose spec declares a server variable (region), I needed to verify/correct the region value. Opening the credential's view/edit modal shows **only the Name and the Token field** — the authentication method and the entire Server configuration section (the Region dropdown that appeared during creation) are absent. There is no way to see what a stored credential is configured to do, and no way to change it.

### Why it matters
- **Can't verify:** when a call fails with a region-shaped error (401 "invalid key" from the wrong regional host), the one thing you need to check — "what region is this credential set to?" — is unviewable.
- **Can't fix:** the only remedy is deleting and re-creating the credential from scratch, blind, hoping the selection sticks this time. Combined with the type-the-slug delete friction (feedback#26 (internal tracker)) and no validity feedback (feedback#43 (internal tracker)/#68), a simple "flip region to eu" becomes a multi-step, unverifiable ritual.
- Same class as feedback#35 (internal tracker) (OAuth scopes not viewable/editable after setup): **everything decided at credential-creation time becomes invisible afterwards.** Create-time state must remain viewable and editable.

### Suggested fix
The edit view should show the full credential configuration: auth method, server-variable values (region), scopes where applicable — read-only where immutable, editable where safe — plus stored-secret presence (masked) per feedback#71 (internal tracker).

### Environment
Jentic One self-hosted, 2026-07-14. Surface: Credentials → edit. Found while setting up PostHog with a region-parameterised spec.

---
*Migrated from the internal dogfooding tracker (feedback#78, originally filed 2026-07-14). Screenshots/attachments referenced in the original will be re-attached.*

Issue #638, opened 2026-07-15
[broker] Region-split APIs fail with a misleading 'invalid key' — a host/region mismatch is surfaced as a credential error


ash-jentic:
Setting up PostHog (an EU-region account), execute returned upstream `401 "Personal API key is invalid."`. The key was fine — the catalogue spec targets the **US** host (`us.posthog.com`) while the account is **EU** (`eu.posthog.com`). Region-split SaaS (PostHog, Mixpanel — which *silently drops data* on the wrong region — Amplitude, Segment, Braze, etc.) route by hostname, and sending a valid key to the wrong regional host yields 401/403/404 or a silent drop.

The broker already knows the real cause (it sees the upstream status and `Www-Authenticate`), but relays a raw 'invalid key', which sends the user to debug their credential for a long time when the problem is the host. Two asks: (1) when a spec exposes a region/server variable, let the user choose region at credential setup (the credential form already renders server-variable inputs); (2) when an upstream auth failure comes from a region-split API, surface a 'possible region mismatch' hint rather than a bare 'invalid key'. Cost us ~20+ minutes chasing the wrong cause; every region-mismatched user hits the same trap.

---
*Migrated from the internal dogfooding tracker (feedback#76, originally filed 2026-07-14). Screenshots/attachments referenced in the original will be re-attached.*

Manuel-Jentic:
Closing as completed — fixed by [PR #717](https://github.com/jentic/jentic-one/pull/717), verified present on current `main`: the broker attaches a `Jentic-Hint` header on upstream 401/403 when the API's spec uses a templated host (`src/jentic_one/broker/core/headers.py`, `broker/web/routers/execute.py`, covered by `tests/unit/broker/test_region_mismatch_hint.py`). The companion server-variable persistence fix (#641) is already closed.

This should have auto-closed on merge — the PR body said `Resolves **#638**`, and the bold markers appear to have broken GitHub's closing-keyword parsing.

PR #666, opened 2026-07-16
docs(skills): add contribute-spec-fix skill (overlay fix → PR → optional local apply)

### LINKED ISSUE #644: End-to-end experience: getting ONE API (PostHog) from import to first successful call — an afternoon-long gauntlet through 8 distinct product failures
## What I was trying to do

Connect **one API I actually use** — PostHog (EU cloud account) — so my agent could read my analytics. Toolbox-first approach: I'd already imported the APIs I use; this was "add the credential and make it work." It became an afternoon-long diagnostic odyssey involving a domain expert + an AI agent working together, and it surfaced eight distinct product failures before the first successful call. It ended in success — the agent now reads my real PostHog dashboards through the broker — but nothing about the journey was survivable for a normal user.

### Phase 1 — the credential
1. **No provider guidance** (feedback#24 (internal tracker)): the credential screen says "detected authentication method: bearer token" with no link to PostHog, no hint where keys live. Getting a Personal API key took **~20 minutes** of self-directed hunting (I'm non-technical).
2. **Vocabulary trap** (feedback#77 (internal tracker)): PostHog calls it a *"Personal API key"*; the form offers "API key" and "Bearer token" as options. I nearly picked the lookalike "API key" (wrong — it's sent as a Bearer header). Nothing bridges vendor vocabulary to transport type.
3. **No feedback on save** (feedback#68 (internal tracker), feedback#71 (internal tracker)): after saving, no dots in the field, blank on reveal (eye icon does nothing), no validity check, no way to confirm anything was stored.
4. **Built the whole chain on a dead end:** bound the API to my toolkit, added permission rules, asked the agent to call it → upstream `401 "Personal API key found in request Authorization header is invalid."` — the FIRST feedback of any kind, several steps and ~an hour after key entry.

### Phase 2 — the false trail
5. The 401 says *credential*, so we debugged the credential: re-checked the key, recreated it, suspected the field was storing the wrong value (can't tell — feedback#71 (internal tracker)/#78). All wasted effort: **the key was valid the entire time.**
6. **Real cause: region.** My PostHog account is EU (`eu.posthog.com`); the catalogue spec hardcodes the US host. A valid EU key sent to US fails with "invalid key" — a **host problem misreported as a credential problem** (feedback#76 (internal tracker)). Region-split SaaS (Mixpanel, Amplitude, Segment, Braze…) all share this failure class; Mixpanel's version silently drops data.

### Phase 3 — fixing the spec (where each fix exposed the next bug)
7. Fixed spec prepared: `"servers": [{"url": "https://{region}.posthog.com", "variables": {"region": {"default": "us", "enum": ["us","eu"], "description": "…must match your project's region; the wrong region returns 401 invalid key."}}]`.
8. **Import demands an identity the modal never asks for**: first two imports failed after a ~40s wait with `cannot resolve api_identifier: missing vendor` — the paste/upload form has no vendor/name/version fields even though the API accepts overrides (filed separately, see the import-identity issue).
9. **Import "completed" into an unusable state**: the successful import landed as a **draft with no live revision**; the credential form then said *"We couldn't read the API spec — pick the type manually below"* and showed all four auth types (two of which can never work for PostHog). Nothing pointed to the needed **Promote** step (feedback#47 (internal tracker) family; also the auth-fallback design question in feedback#77 (internal tracker)).
10. **Promote found and clicked** (the API detail page explains it well once you find it) → spec became readable → the form transformed: correct "Detected: Bearer token", and a **Region dropdown (us/eu)** rendered from the server variable, with the description text as inline guidance. (This part of the product — ServerVariablesSection — is excellent: pure spec-driven UI.)
11. **The region value doesn't persist** (feedback#79 (internal tracker)): selected `eu`, created the credential, executed → TLS shows the request still went to `us.posthog.com`. Recreated carefully; same. The dropdown renders but its value never reaches the stored credential; `server_variables` appears zero times in a DEBUG-level broker log. And the credential edit view hides server configuration entirely (feedback#78 (internal tracker)), so this is invisible to the user.
12. **Workaround: set the spec's default to `eu` and re-import** → **promote-over-live 500s** (feedback#80 (internal tracker)): `sqlalchemy.exc.MissingGreenlet` in `revision_service.promote`; first click failed silently, second showed "Promote failed — Internal Server Error." You can publish an API once but cannot ship a corrected spec over it.
13. **Workaround for the workaround: remove the API and import fresh** (first-promote works) → **removing the API stranded its credential bindings** (feedback#81 (internal tracker)): execute now hit `409 "Ambiguous: 2 active credentials match"`, with both candidates displaying the identical machine name (feedback#69 (internal tracker)). Unbinding the stray from the toolkit didn't help — **the resolver matches credentials across the whole store, not the toolkit's bindings** (feedback#81 (internal tracker) update). Only deleting all-but-one credential (via the type-the-full-slug ritual, feedback#26 (internal tracker), twice) cleared it.

### Phase 4 — success
14. Execute → request TLS-confirmed at **`eu.posthog.com`** → PostHog **accepted the key** and returned a *clear, actionable* scope error (org-level endpoint not allowed for project-scoped keys) → switched to a project-scoped endpoint → **200 OK, real dashboards returned through the broker.**

## What this cost
A product-owner-level user plus an AI agent, several hours of continuous active work in a single afternoon, four platform bugs requiring workarounds, two spec re-imports, one API deletion, and five credentials created to end up with one. **Every individual obstacle is filed; this issue exists to show they compound into a single user story.**

## What worked (credit)
- The broker's diagnostics are excellent when surfaced: `Jentic-Error-Origin`, `Jentic-Upstream-Status`, TLS host in logs told the truth throughout.
- The 409 ambiguity refusal is loud rather than silently picking a key — correct behaviour.
- Spec-driven UI (auth detection + server-variable dropdown + description-as-help-text) is genuinely elegant — one spec edit produced a guided region picker with inline docs.
- PostHog's own scoped-key 403 message was the clearest error of the entire journey.

## The one-sentence takeaway
The platform's *architecture* did everything right at the end; the *journey* to reach it was survivable only because this was a dogfooding exercise with unlimited patience — a real user would have churned at step 4, believing their key was broken.

## Related issues
feedback#24 (internal tracker), feedback#26 (internal tracker), feedback#28 (internal tracker), feedback#35 (internal tracker), feedback#37 (internal tracker), feedback#43 (internal tracker), feedback#47 (internal tracker), feedback#49 (internal tracker), feedback#51 (internal tracker), feedback#68 (internal tracker), feedback#69 (internal tracker), feedback#70 (internal tracker), feedback#71 (internal tracker), feedback#76 (internal tracker), feedback#77 (internal tracker), feedback#78 (internal tracker), feedback#79 (internal tracker), feedback#80 (internal tracker), feedback#81 (internal tracker) + the import-identity and import-status issues filed alongside this one.

*(Screenshots to be attached: credential form "detected bearer" with sk_live placeholder; Choose-an-API picker with duplicate PostHog entries; "couldn't read the API spec" four-option auth grid; API-key/Location wrong branch; delete-credential type-the-slug modal; the Region dropdown with eu selected and description text; the us (default) variant; "Promote failed — Internal Server Error" toast over the revisions panel; final toolkit state with single PostHog credential.)*

---
*Migrated from the internal dogfooding tracker (feedback#82, originally filed 2026-07-14). Screenshots/attachments referenced in the original will be re-attached.*

sophie-jentic:
## Summary

Adds a new agent skill, **`contribute-spec-fix`**, that turns an OpenAPI spec fix into a community contribution instead of a local patch. It:

1. Captures the fix as an **OpenAPI Overlay 1.0.0** (`meta/overlay.json`).
2. **Validates** it before anything ships:
   - **Idempotency** — applies the overlay twice via `bump-cli` and asserts it's a no-op on the committed spec (semantic JSON compare, so formatting/key-order differences don't matter).
   - **Lint** — `spectral lint` against `spectral:oas`, asserting zero errors and no new finding codes vs. the pre-change baseline.
3. **Applies** the overlay to `openapi.json` with a small **format-preserving Python applier** (keeps indentation and original key position → minimal, reviewable diff). No CLI applier preserves formatting — `spectral` is lint-only, `bump-cli`/Redocly both re-serialise — so the applier is in-repo and the CLI tools are used only as the cross-check/lint.
4. **DEFAULT: opens a PR to `jentic/jentic-public-apis`** so the whole community gets the fix.
5. **FALLBACK (opt-in): applies the same overlay to the local Jentic registry** (via the registry overlay API — `POST /apis/{vendor}/{name}/{version}/overlays` + `:confirm`) only when the user is blocked on maintainer approval time. The PR stays open as the source of truth — nothing forks.

This is the **improvement flywheel**: fixes discovered by AI agents flow back into the shared catalogue rather than being siloed on one machine.

## Motivation

Comes directly out of the PostHog EU/US region gauntlet documented in #644 — where the spec hardcoded the US host and an EU account got a misleading `401 invalid key`. The fix in that story (steps 6–7) was exactly a **region server-variable enum** (`https://{region}.posthog.com`, `enum: [eu, us]`). I've since landed that spec fix upstream in **jentic/jentic-public-apis#21889**; this skill generalises and automates that whole overlay → validate → PR path so the next agent doesn't repeat the afternoon-long manual slog.

Related platform issues this touches: #641, #638, #640.

## ⚠️ Before #644 can be closed — platform must honour the server-URL enum

This skill (and the upstream spec fix #21889) fixes the **spec side**: the server URL becomes a region enum. But that only helps if **jentic-one actually applies the chosen server-variable value end-to-end**. Per #641, the region selected at credential creation is currently **dropped between the form and the stored credential**, and the broker always routes to the spec `default` (`server_variables` never reaches the resolver).

So closing #644 still requires confirming — or adding — that **jentic-one supports a server-URL enum choice** and routes to the selected region:

- [ ] Confirm the credential-create region selection is persisted to the stored credential (#641).
- [ ] Confirm the broker applies the credential's `server_variables` for region-split APIs so execution hits the chosen host, not the spec default (#641, #638).
- [ ] Until then, the upstream enum spec + this skill mitigate but **do not** fully resolve #644.

**This PR does not by itself close #644.** It delivers the contribution mechanism and references the region-enum spec fix; the platform-side region routing (#641) must be verified/implemented first.

## Files
- `skills/contribute-spec-fix/SKILL.md` — the skill (self-contained; no machine-specific values).

### DIFF
--- a/skills/contribute-spec-fix/SKILL.md
+++ b/skills/contribute-spec-fix/SKILL.md
@@ -1,0 +1,326 @@
+---
+name: contribute-spec-fix
+description: Fix a broken OpenAPI spec in jentic-public-apis with an OpenAPI Overlay, validate it (spectral lint + idempotency check), and contribute it back via a PR to the community catalog. Falls back to applying the same overlay to the local Jentic registry if the user can't wait for maintainer approval. Use when an API spec is wrong/incomplete (bad server URL, missing enum, wrong content-type, missing param, etc.) and the fix should reach the community project, not just the local box.
+argument-hint: [vendor/api or path to the broken spec]
+---
+
+# Contribute a Spec Fix (Overlay → PR → optional local apply)
+
+Fix a spec in `jentic-public-apis` the right way: capture the fix as an **OpenAPI Overlay**,
+validate it, and **open a PR so the whole community gets the fix**. This is the improvement
+flywheel — fixes made by AI agents are contributed back, not siloed on one machine.
+
+Target: `$ARGUMENTS` (a `vendor/api` id like `posthog.com/posthog-api`, or a path to a spec).
+
+## The flywheel — read this first
+
+```
+        broken spec found
+                │
+                ▼
+     write overlay  ──►  validate (lint + idempotency)  ──►  apply to openapi.json
+                │
+                ▼
+     ┌─────────────────────────────────────────────┐
+     │  DEFAULT: PR to jentic/jentic-public-apis     │  ← community gets the fix
+     └─────────────────────────────────────────────┘
+                │
+                ▼  (only if the user is blocked on maintainer approval time)
+     ┌─────────────────────────────────────────────┐
+     │  FALLBACK: apply the SAME overlay to the      │  ← unblocks the user now;
+     │  local Jentic registry — PR stays open        │    PR is still the source of truth
+     └─────────────────────────────────────────────┘
+```
+
+**Always open the PR.** The local apply is an impatience valve, not a substitute — it uses the
+*same overlay document*, so nothing forks. Never skip the PR just because a local apply worked.
+
+## Prerequisites
+
+- A local checkout of `jentic-public-apis`. Find it (don't assume the path):
+  `find ~ -maxdepth 4 -type d -name jentic-public-apis 2>/dev/null`. The repo has `origin`
+  pointing at `jentic/jentic-public-apis`.
+- `python3` — the format-preserving overlay applier (step 4) and the validation checks are Python.
+- `spectral` on PATH (`spectral --version`) — used **only for linting** (step 6). Spectral is a
+  linter/validator; it has no command to apply an overlay and emit a modified spec.
+- `npx` available — `bump-cli` (fetched on demand via `npx --yes bump-cli`) is used **only as the
+  idempotency cross-check** (step 5), never as the writer: it re-serializes and reorders keys, so
+  it would reformat the whole file. Redocly 2.x can apply overlays too but only via a
+  `redocly.yaml` config and it also reorders — same problem. No CLI applier preserves source
+  formatting, which is why step 4 uses a small in-repo Python applier instead.
+- `gh` authenticated (`gh auth status`) for the PR.
+- For the local fallback only: a running Jentic control plane (default `http://127.0.0.1:8000`)
+  and a registered agent (`jentic profile list` shows a valid token).
+
+## Steps
+
+### 1. Locate the spec and understand the fix
+
+Resolve `$ARGUMENTS` to a spec at
+`apis/openapi/<vendor>/<api>/<version>/openapi.json` in the jentic-public-apis checkout.
+Read the relevant section and confirm with the user exactly what's wrong and what the fixed
+value should be (e.g. "servers only has the US host; add an EU host"). Note the `<vendor>`,
+`<api>`, and `<version>` path segments — you need them for both the PR and the local apply.
+
+### 2. Branch off the latest main
+
+Never work on a dirty tree or a stale base (see the no-direct-push-to-main rule).
+
+```
+git -C <repo> fetch origin main
+git -C <repo> switch -c fix/<vendor>-<short-slug> origin/main
+```
+
+If the working tree has an unrelated in-progress state that blocks the checkout, prefer an
+isolated `git worktree add -b fix/... <path> origin/main` so you don't disturb it. If you must
+clear it in place, record the current branch tip first and confirm with the user.
+
+Confirm the branch is at the tip of origin/main: `git -C <repo> log --oneline -1`.
+
+### 3. Write the overlay
+
+Create `apis/openapi/<vendor>/<api>/<version>/meta/overlay.json` (JSON for JSON specs). Use the
+**OpenAPI Overlay 1.0.0** format. Prefer **remove-then-set** action pairs — they are
+deterministic regardless of the starting state, which is what makes the overlay idempotent:
+
+```json
+{
+  "overlay": "1.0.0",
+  "info": { "title": "Overlay for the <Vendor> API", "version": "1.0.0" },
+  "actions": [
+    {
+      "description": "Why the current value is wrong.",
+      "target": "$.servers",
+      "remove": true
+    },
+    {
+      "description": "The corrected value.",
+      "target": "$",
+      "update": { "servers": [ /* corrected servers */ ] }
+    }
+  ]
+}
+```
+
+`target` is JSONPath. For "a field that should offer a choice of hosts", use a single templated
+server with a server **variable enum** (canonical OpenAPI), e.g.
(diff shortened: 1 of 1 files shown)


ren-jentic:
With this being in jentic-one repo, wondering if would be better for this to be about how to test the overlay in jentic-one and then when happy push the overlay to jentic-public-apis as the primary way to do it?  I feel like that is nicer, but @Manuel-Jentic keen on your thoughts

Manuel-Jentic:
> With this being in jentic-one repo, wondering if would be better for this to be about how to test the overlay in jentic-one and then when happy push the overlay to jentic-public-apis as the primary way to do it? I feel like that is nicer, but @Manuel-Jentic keen on your thoughts

Haven't formed an opinion yet, I'll comeback later!

Manuel-Jentic:
> With this being in jentic-one repo, wondering if would be better for this to be about how to test the overlay in jentic-one and then when happy push the overlay to jentic-public-apis as the primary way to do it?

Came back with an answer, grounded in the code: **your ordering is the right end state, but it isn't honest today.** The registry stores overlays (`POST /apis/{v}/{n}/{ver}/overlays` + `:confirm`) but never applies them — the `overlays=true` flag on the spec download endpoints is a documented no-op, and the broker executes from the operation rows extracted at ingest, so a confirmed overlay changes nothing an agent can observe. "Test the overlay in jentic-one first, then push upstream when happy" can't actually verify anything yet.

So the sequencing we'll follow:

1. **Merge this PR as written** — PR-first, local apply as the opt-in fallback — with one edit: the local-apply step (step 9) must say that until overlay application lands, `:confirm` records intent but does **not** change the served spec; verification is via the applied `openapi.json` on the PR branch.
2. **Implement overlay application in the registry**: async materialize-on-confirm through the existing ingest path (D-005a: no synchronous swaps). One security note that falls out of this: submit *and* `:confirm` are both `apis:write` today, so before confirm rewrites the served spec it must become operator-gated — otherwise a single agent could redirect `servers[].url` and the broker would send credentials there.
3. **Then flip the skill to your ordering** — verify locally first, PR upstream when happy — because local verification will actually mean something.

Full write-up in the plan doc landing shortly as `docs/plans/spec-flywheel-and-catalog-update-notify.md`.

PR #714, opened 2026-07-21
fix(registry): include auth requirements in operation inspect response

juliahooper:
fix(registry): include auth requirements in operation inspect response

SUMMARY_LOAD_OPTIONS had load_auth=False, so the /inspect endpoint never populated the auth field despite the model and loader already supporting it. Agents inspecting an operation saw no authentication requirements.

## Summary

The `/inspect` endpoint's `SUMMARY_LOAD_OPTIONS` had `load_auth=False`, meaning auth requirements were never included in inspect responses. The model (`OperationInspectResult.auth`), the loader (`_load_security_schemes()`), and the CLI renderer (`apis.go:385`) all already support auth data — it just wasn't being loaded.

## Changes

- Flip `load_auth=False` → `load_auth=True` in `SUMMARY_LOAD_OPTIONS` (`src/jentic_one/registry/services/inspect/models.py`)

## Testing

- `make check` passes
- `jentic apis inspect <operation_id>` now includes auth/security scheme details in output

## Checklist

### DIFF
--- a/src/jentic_one/registry/services/inspect/models.py
+++ b/src/jentic_one/registry/services/inspect/models.py
@@ -17,7 +17,7 @@ class InspectLoadOptions:
     load_server: bool
 
 
-SUMMARY_LOAD_OPTIONS = InspectLoadOptions(load_spec=False, load_auth=False, load_server=True)
+SUMMARY_LOAD_OPTIONS = InspectLoadOptions(load_spec=False, load_auth=True, load_server=True)
 
 
 class TagDescription(BaseModel):


ren-jentic:
Hey @juliahooper - So this does solve the issue, and returns the auth-requirements - however ive found agents get confused by this as it returns what the underlying OpenAPI spec auth requirements needs but thats not necassarily correct when using the Broker (ie the auth to use the openapi spec directly, and then auth to use it via the broker are different).  

Could you look into this a little bit more and see what you can come up with?  Perhaps we just need better returned docs/information so that the agent knows the auth required for Broker call vs Direct call?

PR #731, opened 2026-07-21
feat(web): serve the onboarding skill over HTTP (#651)

### LINKED ISSUE #651: [request] Serve the jentic onboarding skill from a URL on the deployed service
### What's the problem, friction, or need?

When onboarding an agent to jentic, it needs the onboarding skill (currently `cli/internal/skillgen/content/jentic.md`). Today that skill is written to the local device at startup — so it only lands on the machine running the installation.

When jentic is deployed and the agents run on different machines, there's no first-class way to get the skill to them: I have to copy the markdown to each agent host manually, and those copies can drift from the version actually deployed. An agent has no reliable, self-describing way to fetch the skill that matches the running service.

### What would you like instead?

The jentic service should expose the skill over HTTP so any agent that can reach the deployment can read it — e.g. `deployed.jentic.intranet.company.io/SKILL.md`. Because it's served by the same deployment, it always matches the running version and can't drift.

**Alternatives considered:**

- **Link on jentic.com** — not chosen: can drift from the deployed implementation, and agents may not have access to jentic.com.
- **Static file on GitHub** — not chosen: agents may not have access; even where they can reach it via jentic, they don't know how to use jentic before reading the skill (chicken-and-egg).

### Where did this come up?

`cli/internal/skillgen/content/jentic.md`

ren-jentic:
## Summary

An agent onboarding to Jentic needs the `jentic.md` skill, but it was only embedded in the Go CLI (`//go:embed content/jentic.md`) and written to the local device at install. A remote agent that could reach the deployment had no first-class way to fetch it, and manually-copied copies could drift from the running service. This serves the *same* skill over HTTP so any agent reachable to the deployment reads the exact version the service ships.

## Served URL

- `GET /skills/jentic.md` (canonical)
- `GET /SKILL.md` (alias)

Both return `text/markdown; charset=utf-8`, are **unauthenticated** (an onboarding agent must fetch it before holding any token), and are **schema-hidden** (`include_in_schema=False`) since this is onboarding content, not a product API. Served on the admin surface, so it is reachable at a stable URL in both standalone (`make start-admin`) and combined (`make start-app`) deploy modes.

## Changes

- **`src/jentic_one/shared/web/skill_router.py`** — new public, schema-hidden router serving the onboarding skill (modeled on `reference_router.py`).
- **`src/jentic_one/admin/web/app.py`** — wire the skill router into `get_routers()` (root prefix), so it serves in both standalone and combined modes.
- **`pyproject.toml`** — force-include the *same* file the CLI embeds (`cli/internal/skillgen/content/jentic.md`) into the wheel at `jentic_one/skills/jentic.md` (alongside `openapi`/`static`), plus the sdist. The router reads it via `importlib.resources`, falling back to the CLI source file in a source checkout.
- **`cli/internal/skillgen/bundled.go`** — update the stale comment (the hosted endpoint now exists) to document the shared source of truth.

## Single source of truth / no drift

Both the Go CLI embed and the backend route read **one committed file** (`cli/internal/skillgen/content/jentic.md`). The wheel force-includes that exact file, so the CLI-embedded copy and the served copy are byte-for-byte identical and cannot drift. A drift-guard test pins the invariant directly.

## Test plan

- [x] `test_skill_router.py`: `GET /skills/jentic.md` + `/SKILL.md` return 200 with `text/markdown` content-type and the expected content, in standalone and combined modes.
- [x] Skill route is unauthenticated (fetchable with no `Authorization` header).
- [x] Skill route is excluded from the OpenAPI schema.
- [x] Drift guard: served content == CLI-embedded source file, byte-for-byte.
- [x] `make test-fast` (2117 passed), `make lint` (ruff + mypy clean).
- [x] `make openapi` produces no spec drift (schema-hidden route).
- [x] `cd cli && go build ./... && go test ./...` all pass; `gofmt` clean.
- [x] Built the wheel and verified `jentic_one/skills/jentic.md` is packaged and matches the CLI source.

Please **squash + merge**.

Closes #651

### DIFF
--- a/cli/internal/skillgen/bundled.go
+++ b/cli/internal/skillgen/bundled.go
@@ -9,9 +9,13 @@ import (
 )
 
 // bundledSkill is the canonical "how to use Jentic via the CLI" content shipped
-// in the binary. It is the de-facto source until #277's hosted `GET /skills/*`
-// endpoint exists, and is kept identical to that to-be-hosted markdown so the
-// two never drift.
+// in the binary via go:embed. The very same file is force-included into the
+// backend wheel (see `pyproject.toml` -> `jentic_one/skills/jentic.md`) and
+// served over HTTP at `GET /skills/jentic.md` (alias `GET /SKILL.md`) by
+// `shared/web/skill_router.py`, so a remote agent can fetch the skill from a
+// deployment (issue #651). Because both the embed and the served route read
+// this one committed file, the CLI-embedded copy and the served copy cannot
+// drift; a backend drift-guard test pins that invariant.
 //
 //go:embed content/jentic.md
 var bundledSkill string
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,14 +54,22 @@ exclude = ["**/.gitkeep"]
 
 # Bundle OpenAPI specs and the built UI inside the package so they're reachable
 # via importlib.resources.files("jentic_one") / "openapi" and / "static".
+#
+# The onboarding skill is force-included from the *same* physical file the Go
+# CLI embeds (`cli/internal/skillgen/content/jentic.md`), so the served copy and
+# the CLI-embedded copy are byte-for-byte identical and cannot drift. It lands
+# at `jentic_one/skills/jentic.md`, read at runtime via importlib.resources by
+# `shared/web/skill_router.py`.
 [tool.hatch.build.targets.wheel.force-include]
 "openapi" = "jentic_one/openapi"
 "ui/dist" = "jentic_one/static"
+"cli/internal/skillgen/content/jentic.md" = "jentic_one/skills/jentic.md"
 
 [tool.hatch.build.targets.sdist]
 include = [
     "src/jentic_one",
     "openapi",
+    "cli/internal/skillgen/content/jentic.md",
     "README.md",
     "pyproject.toml",
 ]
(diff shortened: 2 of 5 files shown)


Manuel-Jentic:
Closing as **superseded** — main now ships this feature (and a bit more) via #810 and #822, so there is nothing left for this branch to add. Verified against this PR's full diff (Ren is out for a few days; Manuel + agent handled the handover):

- **Same URLs, same semantics**: `GET /skills/jentic.md` (canonical) + `GET /SKILL.md` (alias) are served on main — unauthenticated, `text/markdown; charset=utf-8`, schema-hidden, mounted on both combined and standalone surface apps (`src/jentic_one/shared/web/agent_discovery.py`, wired in `app_factory.py`).
- **Plus more**: #810 also added the `llms.txt` index (`GET /llms.txt` + `/.well-known/llms.txt`) linking the skill, OpenAPI, endpoint reference, and RFC 8414 doc.
- **The `bundled.go` comment update here is already on main** — it now documents the served endpoint and the drift guard.
- **The one mechanical difference was decided the other way deliberately**: this PR force-includes the CLI's `content/jentic.md` into the wheel (`importlib.resources`), while main keeps a committed copy at `src/jentic_one/shared/web/content/jentic.md` pinned byte-for-byte by `tests/arch/test_skill_drift.py` (CLI copy is edited first, then synced — the convention #822 reinforced after the first real drift). Same single-source-of-truth guarantee, enforced by test instead of packaging.

Thanks Ren — the design here (unauthenticated, schema-hidden, admin-surface, single source of truth) is exactly what landed; only the packaging mechanism differs. Closing rather than rebasing 66 commits.

PR #732, opened 2026-07-21
feat(release): publish app image to GHCR + document self-hosted deploy

### LINKED ISSUE #653: Deployment: publish an image + document self-hosted (container + external Postgres) deploys
## Summary

Deploying Jentic One to a real (non-local) environment currently requires reverse-engineering the build and runtime model from source. Two gaps compound: there's no published image to deploy, and no guide for the common "run the containers against an external managed Postgres" topology. This came out of a real deployment (app + broker on a host via docker-compose, dedicated RDS, secrets from a secrets manager, admin UI behind a load balancer).

## 1. Publish a production image

- `IMG_PREFIX` defaults to the bare string `jentic-one`, there's no release workflow that builds/pushes images, and `deploy/README.md` lists the registry as "optional, future".
- Consequence: every deployer builds from source (`make build-base` → `make build-app` → tag → push to their own registry) before anything runs, and the Helm/Terraform assets under `deploy/` reference image repositories that don't exist until you do this.
- The app and broker are the **same** image differentiated by `JENTIC__APPS` — good design, but undocumented for deployment, so it's not obvious you only build/push one image.

**Ask:** publish images (e.g. GHCR) on tag, at minimum the `app` image (it can serve any surface). If publishing isn't ready, ship a single documented `make release-image REGISTRY=…` and state plainly that one `app` image serves all surfaces via `JENTIC__APPS`.

## 2. Document the "containers + external Postgres" deploy path

The provided paths (local compose with bundled PG, `jenticctl install` dev wizard, Helm, Terraform-wrapping-Helm) don't cover what's arguably the most common self-host topology: **app + broker containers on a VM/host, pointed at an external managed Postgres, with secrets injected from a secrets manager.** Things that had to be derived from source:

1. The app and broker are one image split by `JENTIC__APPS` (`registry,admin,control,auth` vs `broker`), and **the broker must run as the sole surface** (`_build_app` raises otherwise). Found in `__main__.py`, not a deploy doc.
2. No worked production `config.yaml` example for external Postgres with the full required set (encryption keyset, admin JWT secret, invite pepper, connect state secret, DB creds).
3. Which secrets are mandatory to boot vs optional — learned by hitting startup validation errors one at a time.
4. The three schemas (`registry`, `control`, `admin`) must be created before migrations; documented only implicitly via `scripts/setup.sh` for the bundled compose.
5. The migration entrypoint for a container deploy is `python -m jentic_one.migrations.run` (found via `make migrate-sqlite`), not called out in deploy docs.
6. First-run `create-admin` step.

**Ask:** add a "Self-hosted: containers + external Postgres" section to `deploy/README.md` covering the one-image/two-surfaces model + sole-broker constraint, a complete minimal production `config.yaml`, the required-secrets list, the create-schemas + `migrations.run` bootstrap sequence, and the `create-admin` first-run step.

## Why combined

Both are the same adoption barrier — "how do I run this somewhere real?" — and a single deploy guide that also points at a published image would resolve them together.

ren-jentic:
## Summary

Makes self-hosting real for issue #653: (1) publish the container image on release — signed, with an SBOM attestation, and (2) document a container + external-Postgres deploy end to end. Documentation-first + a publish workflow, no invasive runtime changes.

## Changes

- **Publish workflow** — new `publish-image` job in `.github/workflows/release.yml` (runs on a `vX.Y.Z` tag, `needs: [gate, smoke]` so a red Helm smoke matrix never ships an image). Builds the `app` image (same flow as `make build-app`: `build-base` then `deploy/docker/app.Dockerfile`) and pushes to `ghcr.io/<owner>/jentic-one-app` tagged `X.Y.Z` (v-stripped) + the short SHA, using the built-in `GITHUB_TOKEN` with `packages: write`. The pushed digest is cosign-signed (keyless, OIDC) with an SPDX SBOM attestation; `:latest` moves **after** signing and only on stable releases. Installer actions run before `docker login` so third-party actions never execute with the registry token on disk. The `release` (GoReleaser) job now also needs `publish-image`, so a release can't half-ship. Single-arch (amd64) for now.
- **Makefile** — `make release-image REGISTRY=…` builds + pushes the app image from a fork/registry, mirroring CI's stable-only `:latest` guard.
- **`deploy/README.md`** — new **"Self-hosted: containers + external Postgres"** section: one-image/two-surfaces model + sole-broker constraint, the three-schemas DB model with accurate `JENTIC__…` env vars, mandatory-vs-optional secrets, a worked production `config.yaml`, the roles+schemas → `migrations.run` → `create-admin` bootstrap, digest-pin + `cosign verify` guidance, a `docker-compose` example with healthchecks, and an Upgrading/rollback subsection. Refreshed the deferred image-registry/multi-arch notes (incl. the Helm-consumption gap).
- **`docs/releasing.md`** — publish stage documented; release-recovery section covers re-run-failed-jobs vs full-re-run (tag overwrite → new digest) and the first-release GHCR visibility checklist; immutable-tags trade-off + tag-ruleset recommendation.
- **`config/production.yaml.example`** — caught up with reality: `auth` in `apps`, `canonical_base_url`, encryption keyset stanza (verified loadable).
- **`ci.yml`** — replaced the stale commented-out `build-images` stub with a pointer to the release publish flow.

## How to publish + self-host

- Publish (CI): merge the release-please PR → the `vX.Y.Z` tag runs `gate` + `smoke` → `publish-image` → `release`. Manual from a fork: `docker login ghcr.io && make release-image REGISTRY=ghcr.io/<org>`.
- Self-host: pull by digest (`ghcr.io/jentic/jentic-one-app@sha256:…` from the release output), create the roles + `registry`/`control`/`admin` schemas, `python -m jentic_one.migrations.run`, `python -m jentic_one create-admin`, then run one `app` (default `JENTIC__APPS`) + one `broker` (`JENTIC__APPS=broker`) container. Full walkthrough + compose file in `deploy/README.md`.

## Review

Two review-board rounds (9 independent lenses total: security, release-engineering, code-quality, operator-UX, red-team, day-2-ops, architecture, product-fit, cold-eyes fact-check); all findings addressed, including the smoke gating, installer/login ordering, sign-before-`:latest`, the bootstrap-SQL roles/grants gap, and doc-accuracy fixes verified against the code.

## Test plan

- [x] `actionlint` on `release.yml`/`ci.yml`.
- [x] `make -n release-image REGISTRY=…` shows correct build/push commands (stable + prerelease branches); `make help` lists it.
- [x] Local `make build-base` + `docker build -f deploy/docker/app.Dockerfile` builds cleanly (validates the exact publish-job flow).
- [x] Verified against the built image: `create-admin --email …`, `python -m jentic_one.migrations.run --db {registry,control,admin}`, and default `JENTIC__APPS=registry,admin,control,auth` — all match the docs.
- [x] Compose example validated with `docker compose config`; `production.yaml.example` loads through `load_config()`.
- [x] `make lint` (ruff + mypy) + `detect-secrets` pass.
- [ ] Actual GHCR push not exercised — requires running in CI on a tag with `packages: write` (first-release checklist: maintainer sets the new package's visibility to public).

Closes #653

> Please **squash-merge**.


Manuel-Jentic:
## Review — publish app image + self-hosted docs

Reviewed across four lenses (bugs, release/CI workflow correctness, supply-chain/security, docs accuracy). **The PR is well-built and the docs are accurate + copy-paste-safe** — every env-var name, both bootstrap subcommands (`migrations.run`, `create-admin`), the manual `CREATE SCHEMA` step, ports, per-surface `/health`, the compose example, and the keygen one-liner were verified against the code with zero errors. The `build-base` → `app.Dockerfile` tag contract matches, and version drift is actively caught by the `gate` (`/health` version == tag), so nothing publishes on a mismatch.

Findings, by priority:

### 1. Image is unsigned/unattested (HIGH) — the headline
`publish-image` builds + pushes the `app` image with **no cosign signature, no SBOM, no provenance** — while the `release` job in the *same* `release.yml` cosign-signs the CLI binaries (keyless OIDC) + emits syft SBOMs. For a v1 artifact self-hosters `docker pull` and run in production, this is below the bar the project already set for binaries, and there's no cryptographic way to verify the image came from this pipeline. `publish-image` doesn't even request `id-token: write`, so keyless signing isn't wired.

Fix (mirrors the `release` job, low-effort):
- add `id-token: write` to `publish-image`;
- capture the pushed digest and `cosign sign --yes "${IMAGE}@${DIGEST}"`;
- `syft ... -o spdx-json` + `cosign attest ... --type spdxjson`, or switch to `docker buildx build --attest type=provenance,mode=max --attest type=sbom`;
- document `cosign verify` in `deploy/README.md`.

### 2. Partial-release atomicity (major)
`publish-image` and `release` both `needs: gate` and run in **parallel** with no dependency, so the image can publish while the CLI release fails (or vice versa) — leaving a version with an image but no signed binaries/Release, or the reverse. The version-consistency invariant is still protected by the gate; this is about *asset* atomicity.

Fix (pick one): make `release` also `needs: [gate, publish-image]` (shrinks the "binaries but no image" window), or add a `finalize` job `needs: [publish-image, release]` that fails loudly / cleans up on partial failure. At minimum, state in the header comment that a partial release is an accepted risk.

### 3. amd64-only image (minor→major, audience-dependent)
Plain `docker build` = single-arch amd64. arm64 self-hosters (Apple Silicon, Graviton, ARM Linux) get QEMU emulation or an `exec-format`/`no matching manifest` failure. Consciously deferred + documented in "What's deferred", so a known-limitation for merge — but worth an explicit v1 scope decision (amd64-only + documented, or add `docker buildx --platform linux/amd64,linux/arm64` now).

### 4. Mutable-tag pulls, no digest guidance (MEDIUM)
Docs steer self-hosters to `:latest` / `:vX.Y.Z` (the `# or a pinned :vX.Y.Z` comment mislabels a version tag as "pinned" — only `@sha256:` is a pin), and a workflow re-run re-`docker push`es the same version tag, overwriting it on GHCR. Recommend digest pinning for production: show `docker pull ...@sha256:<digest>`, pin the compose `x-image` to a digest, have `publish-image` echo the pushed digest, and consider enabling GHCR immutable tags for `vX.Y.Z`.

### 5. `:latest` on prereleases (minor)
`:latest` is pushed unconditionally; a manually-cut `v1.0.0-rc1` matches the `v*.*.*` trigger and would become the default `docker pull`. Guarded by the release-please process (`prerelease: false`), not the workflow. Fix: only tag/push `:latest` when the ref matches `^v[0-9]+\.[0-9]+\.[0-9]+$`.

### 6. `node:22-slim` tag-pinned, not digest-pinned (LOW)
`python-base.Dockerfile` digest-pins the Python base (good) but `NODE_IMAGE=node:22-slim` is tag-only — and it's a build input baked into the published wheel/image. Digest-pin it for consistency + reproducibility.

### 7. GHCR package visibility (nit)
First push creates the package **private**; a maintainer must set it public out-of-band before anonymous `docker pull` works. Worth a first-release checklist line in `docs/releasing.md`.

### SAFE / verified
Docs secret-handling (env injection, correct keygen, keyset-as-file), GHCR token handling (login separate from build, not baked into layers), permissions scoping (`packages: write` on `publish-image` only), and the `.secrets.baseline` change (line drift + a git-SHA false positive) are all sound.

**Net:** great foundation. I'd treat **#1 (signing/SBOM)** as a near-blocker before the image is advertised as production-ready — the fix is small since the binary-signing pattern already exists in this file — and land **#2** and **#4** alongside. **#3 (arm64)** is the main v1 product-scope call.

Manuel-Jentic:
Continuing this while @ren-jentic is out. Rebased onto main (conflicts: `release.yml` header/jobs vs the new `smoke` stage — both stages kept — and a trivial `Makefile` `.PHONY` union) and addressed the Jul 22 review point-by-point:

1. **Sign/attest the image (HIGH)** — `publish-image` now has `id-token: write`, cosign-signs the pushed digest (keyless, same OIDC pattern as the CLI binaries) and attaches a syft SPDX SBOM via `cosign attest`. `deploy/README.md` gains a "Verify the image signature" section with the matching `cosign verify` / `verify-attestation` invocations.
2. **Partial-release atomicity (major)** — `release` now `needs: [gate, smoke, publish-image]`; a tag can no longer half-ship. Documented in the workflow header.
3. **amd64-only (scope)** — kept as the documented v1 limitation (`deploy/README.md` "What's deferred"); multi-arch stays a follow-up.
4. **Digest-pin guidance (MEDIUM)** — the misleading `# or a pinned :vX.Y.Z` comment is gone; docs now state only `@sha256:` is a pin, show the digest pull, pin the compose example to the release digest, and `publish-image` echoes the pushed digest.
5. **`:latest` on prereleases (minor)** — `:latest` only moves when the ref matches `^v[0-9]+\.[0-9]+\.[0-9]+$`.
6. **`NODE_IMAGE` pin (LOW)** — `node:22-slim` digest-pinned like its python sibling; `make build-base` verified locally against the pin.
7. **GHCR first-push visibility (nit)** — `docs/releasing.md` now has an explicit first-release checklist item (package starts private → flip to public, consider immutable tags).

actionlint clean; base image builds against the new pin. The actual GHCR push remains only exercisable on a real tag — to be verified on the next release cut.

PR #734, opened 2026-07-21
feat(auth): invite-redemption page for finishing account creation

Manuel-Jentic:
## Summary

Adds the **invite-redemption page** so an admin-invited user can finish their own
account setup, and bundles the supporting **generic OSS** backend/UI work that a
downstream (private) operator console builds on. **No private-tier code or
naming lives in this PR** — every seam here is generic and useful to the OSS
shell on its own.

> **Reviewer's guide (read me first).** This branch is intentionally a small
> bundle of *independent* changes that share one goal: make the OSS shell
> extensible + fix two real OSS bugs found along the way. They're grouped for
> review convenience; each commit stands alone and could be cherry-picked. See
> "What's here (by commit)" for the map, and "How to review" at the bottom.

### What's here (by commit)

- **`feat(auth)` invite-redemption page** — public `/app/redeem-invite?token=…`
  landing where an invitee sets a password and finishes account creation
  (`POST /users:redeem-invite` returns a JWT and auto-logs them in). Lives
  outside the `AuthGuard`. Hardened: an already-authenticated visitor is bounced
  home, and a link missing its `?token=` is flagged on mount (form disabled)
  rather than only after submit.
- **`feat(ui)` `extraNavItems` seam + permission-gated secondary nav** — a
  generic nav-extension seam (parallel to the existing `extraRoutes` seam):
  `NavItem` gains `requiredPermission` (hide unless the user holds it) and
  `secondary` (render in a divided operator slot); `NavTabs`/`BottomNavbar`
  filter by the user's permissions. Promotes `usePermission`/`ORG_ADMIN` to
  `shared/auth` (already consumed by Monitor) and adds a generic
  `RequirePermission` route guard. Registration is a commit-phase effect with
  unmount cleanup (no render-time side effects, no cross-instance bleed). OSS
  ships zero extra nav items, so stock behaviour is unchanged.
- **`fix(monitoring)` SQLite-compatible usage aggregations** — `overall_stats` /
  `time_buckets` / `grouped_top` / `grouped_trend` hard-coded Postgres-only SQL
  (`percentile_cont … WITHIN GROUP`, `extract('epoch', …)`, `concat()`), so
  `GET /monitoring/usage` 500'd on the local-sqlite backend even though OSS
  already dialect-branches `daily_buckets`. Extends that existing pattern
  (percentiles → `None` on SQLite; `strftime('%s')`; `||` concat) via a single
  `_is_postgres` helper now reused by `daily_buckets` too, + integration
  coverage on both backends.
- **`feat(control)` expose `created_by` (owner)** on `ToolkitResponse` and
  `CredentialRedactedResponse` — the owner was already persisted (set on create,
  used for query-scoping) but never surfaced; now clients can show it.
  Regenerated the control OpenAPI spec + UI client (verified in sync by
  `tests/arch/test_openapi_conformance.py`).
- **`feat(dev)` configurable fixtures Postgres port** — dev compose publishes
  `${JENTIC_PG_PORT:-5432}:5432` so it can run alongside a native Postgres; the
  in-container port and default behaviour are unchanged.
- **`test(web)` role-enforced flywheel** — end-to-end integration test against
  the real combined app + DB: admin creates users with different permission
  sets; admin/write-capable user create agents & toolkits; reads verified by
  role with owner-scoping (admin sees all, non-admins see only their own).
  **Postgres-only** (the `tests/web` harness has no SQLite switch; teardown
  relies on FK `RESTRICT`); SQLite dialect coverage for the read paths lives in
  `tests/integration/admin/test_monitoring_repo.py`.

## OSS purity

No `enterprise` / `admin-console` / `tenant` / `multi-tenancy` strings appear in
any file this PR touches (verified). The seams (`extraNavItems`,
`RequirePermission`, `requiredPermission`, `secondary`) are generic extension
points with **zero OSS callers** — exactly like the pre-existing `extraRoutes`
seam — so they add no behaviour to the OSS binary.

## Pre-review

Ran an internal multi-lens review (OSS-purity/architecture, backend correctness,
UI correctness) before requesting review. Findings addressed in
`refactor(review): address pre-review findings on #734`: removed the name leaks,
moved nav registration to an effect, hardened the redeem page, corrected the
flywheel docstring, and unified dialect detection. Remaining advisory (not
done): the branch could be split into per-concern PRs for a cleaner history —
happy to do that if preferred over reviewing as a bundle.

## How to review

- **Seam design:** compare `extraNavItems` (`ui/src/App.tsx`, `ui/src/shared/app/nav.ts`)
  against the existing `extraRoutes` seam — same "downstream passes extras,
  OSS ships none" shape.
- **Bug fixes stand alone:** the monitoring SQLite fix and `created_by`
  exposure are independent of the invite page.
- **Auth flow:** `RedeemInvitePage` + `AuthContext.redeemInvite`.

## Test plan

- [x] `ruff check .` clean; `ruff format --check` clean.
- [x] `pytest tests/arch` (257) pass — includes OpenAPI drift + no-test-classes guards.
- [x] `pytest tests/integration/admin/test_monitoring_repo.py` on SQLite (8) pass.
- [x] UI `tsc --noEmit` clean; `eslint` clean; `vitest` (nav + auth areas, 81) green.
- [ ] CI (Postgres integration, incl. the flywheel test) on this PR.

### DIFF
--- a/docker/local-setup/docker-compose.yaml
+++ b/docker/local-setup/docker-compose.yaml
@@ -6,7 +6,9 @@ services:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres
     ports:
-      - "5432:5432"
+      # Host-published port is overridable (e.g. to avoid a native Postgres on
+      # 5432); the in-container port stays 5432. Set JENTIC_PG_PORT to change it.
+      - "${JENTIC_PG_PORT:-5432}:5432"
     volumes:
       - db-data:/var/lib/postgresql/data
       - ./init-schemas.sql:/docker-entrypoint-initdb.d/init-schemas.sql:ro
--- a/openapi/control/control.openapi.yaml
+++ b/openapi/control/control.openapi.yaml
@@ -1621,6 +1621,12 @@ components:
           format: date-time
           title: Created At
           type: string
+        created_by:
+          anyOf:
+          - type: string
+          - type: 'null'
+          description: Identity that created the credential (its owner).
+          title: Created By
         credential_id:
           description: Stable credential identifier, prefixed `cred_`.
           title: Credential Id
@@ -4414,6 +4420,11 @@ components:
           format: date-time
           title: Created At
           type: string
+        created_by:
+          anyOf:
+          - type: string
+          - type: 'null'
+          title: Created By
         credential_count:
           title: Credential Count
           type: integer
--- a/scripts/flywheel_manual.py
+++ b/scripts/flywheel_manual.py
@@ -1,4 +1,4 @@
-"""Manual flywheel driver — exercises the admin-console flow against a running app.
+"""Manual flywheel driver — exercises the role-enforced user/agent/toolkit flow.
 
 Not a test; a throwaway script to observe real behaviour end-to-end:
   1. bootstrap admin
--- a/scripts/flywheel_manual.py
+++ b/scripts/flywheel_manual.py
@@ -1,0 +1,133 @@
+"""Manual flywheel driver — exercises the admin-console flow against a running app.
+
+Not a test; a throwaway script to observe real behaviour end-to-end:
+  1. bootstrap admin
+  2. admin creates two users (invite -> redeem -> login) with different scopes
+  3. admin + each user create agents and toolkits
+  4. verify reads by role (who can list what; who gets 403)
+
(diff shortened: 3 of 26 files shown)


ren-jentic (approved):
lgtm

PR #736, opened 2026-07-22
fix(ui): use a dedicated muted token for input placeholder text

### LINKED ISSUE #673: [ui] Placeholder text is the same colour as entered text across the whole app (can't tell empty vs filled fields)
### What's wrong

Across the whole app, placeholder text in form fields is the same colour as the text a user types. So an empty field and a filled-in field look the same, and you can't tell at a glance which fields still need input.

### Why it matters
- Users can't quickly see which fields are empty vs already filled.
- Easy to mistake a placeholder for a real value, or to not notice a field still needs filling in.
- Standard form behaviour is for placeholders to be a lighter/muted grey, clearly different from entered text.

### How to reproduce
- Open any form with input fields (e.g. Add credential, or any settings form).
- Compare the greyed example text to text you type: they're the same colour.

### Expected
Placeholder text should be visibly lighter/greyer than entered text, everywhere in the app, so empty fields read as empty.

### Environment
- Jentic One local app (`http://localhost:8000`), all forms.

---
_Migrated from internal Jentic One dogfooding feedback (original ref #48). Suggested classification: type:bug | area surface:ui | severity:major. Labels omitted per CONTRIBUTING; the intake assistant applies them._

ritvik-jentic:
## Summary

Placeholder text in form fields rendered the same near-white colour as entered
text, so empty and filled fields looked identical app-wide. Placeholders
inherited the `--muted-foreground` token (`#E4EAEB`), which is nearly
indistinguishable from the `#FFFFFF` foreground used for entered text.

This introduces a **dedicated placeholder token** so only placeholder text is
affected. `--muted-foreground` is intentionally left unchanged — it is consumed
by many components for general muted text (table headers, the app rail, badges,
captions, pagination), and darkening it would be the wrong blast radius.

## Related issue

Closes #673

## Changes

- `ui/src/index.css`: add a semantic `--input-placeholder` token (`#689296`,
  `primary-500` — a muted grey clearly lighter than entered text) and register a
  matching `--color-input-placeholder` Tailwind colour in the `@theme inline`
  block. `--muted-foreground` is left untouched at `#E4EAEB`.
- `ui/src/shared/ui/Input.tsx`: placeholder utility
  `placeholder:text-muted-foreground` → `placeholder:text-input-placeholder`.
- `ui/src/shared/ui/Textarea.tsx`: same placeholder utility change.

## Testing

- `npm run lint` (in `ui/`) — clean.
- `npm run build` (in `ui/`) — succeeds; Tailwind compiles the new
  `text-input-placeholder` utility with no "unknown utility class" error.
- Verified the compiled CSS emits
  `::placeholder{color:hsl(var(--input-placeholder))}` resolving to `#689296`,
  distinct from the `#FFFFFF` entered text.
- Manual check: placeholder text now reads as a muted grey and is clearly
  distinct from typed text; other `text-muted-foreground` surfaces (table
  headers, rail, badges, captions) are unchanged.

## Checklist

### DIFF
--- a/ui/src/index.css
+++ b/ui/src/index.css
@@ -22,6 +22,9 @@
 	--color-muted: hsl(var(--muted));
 	--color-muted-foreground: hsl(var(--muted-foreground));
 
+	/* Form placeholder — a muted grey, clearly lighter than entered text */
+	--color-input-placeholder: hsl(var(--input-placeholder));
+
 	/* Card */
 	--color-card: hsl(var(--card));
 	--color-card-foreground: hsl(var(--card-foreground));
@@ -165,6 +168,10 @@
 	--foreground-lighter: var(--primary-100); /* #E4EAEB */
 	--muted: var(--primary-850); /* #193238 */
 	--muted-foreground: var(--primary-100); /* #E4EAEB */
+
+	/* Placeholder text — deliberately a muted grey so empty fields read as
+	   empty and don't look identical to entered (near-white) text (#673). */
+	--input-placeholder: var(--primary-500); /* #689296 */
 
 	/* Brand */
 	--primary: var(--primary-300); /* #A3CACC */
--- a/ui/src/shared/app/rail/DenyReasonField.tsx
+++ b/ui/src/shared/app/rail/DenyReasonField.tsx
@@ -29,7 +29,7 @@ export function DenyReasonField({ id, value, onChange, autoFocus }: DenyReasonFi
 				rows={2}
 				autoFocus={autoFocus}
 				placeholder="Why is this being denied?"
-				className="border-border bg-background focus:border-primary w-full resize-none rounded border px-2 py-1 text-[11px] outline-none"
+				className="border-border bg-background focus:border-primary placeholder:text-input-placeholder w-full resize-none rounded border px-2 py-1 text-[11px] outline-none"
 			/>
 		</div>
 	);
--- a/ui/src/shared/ui/Input.tsx
+++ b/ui/src/shared/ui/Input.tsx
@@ -43,7 +43,7 @@ export const Input = React.forwardRef<HTMLInputElement, InputProps>(function Inp
 					aria-describedby={errorId}
 					aria-invalid={error ? true : undefined}
 					className={cn(
-						'bg-card border-border text-foreground placeholder:text-muted-foreground w-full rounded-lg border transition-colors',
+						'bg-card border-border text-foreground placeholder:text-input-placeholder w-full rounded-lg border transition-colors',
 						'focus:border-primary focus:outline-hidden',
 						sizeClasses[size],
 						startIcon && 'pl-9',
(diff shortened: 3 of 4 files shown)


Manuel-Jentic (changes requested):
## Review

Checked out the branch in a worktree and reviewed the change in full context against the frontend rules.

### What's good
- **Right approach / correct blast radius.** Introducing a dedicated semantic `--input-placeholder` token instead of darkening `--muted-foreground` is the correct call — `--muted-foreground` is consumed widely (table headers, rail, badges, dropdowns), so changing it would over-reach. The reasoning in the description is sound.
- **Follows the `index.css` token conventions exactly**: HSL-triplet source in `:root`, `hsl(var(--…))` wrapper in `@theme inline`, references the existing `primary-500` scale value rather than hard-coding a hex, and the comment format matches the surrounding tokens. `#689296` matches `--primary-500` (`185 18% 50%`).
- **`SearchInput` is covered transitively** — it wraps the shared `Input`, so all search/filter fields inherit the fix.
- Small, focused, Conventional-Commits compliant.

### Findings

**1. (Requesting change) The fix is incomplete — a live raw `<textarea>` still has the bug.**

`ui/src/shared/app/rail/DenyReasonField.tsx` is a raw `<textarea>` (not the shared `Textarea` primitive) with a `placeholder` but **no placeholder color class at all**:

```tsx
<textarea
  id={id}
  value={value}
  onChange={(e) => onChange(e.target.value)}
  rows={2}
  autoFocus={autoFocus}
  placeholder="Why is this being denied?"
  className="border-border bg-background focus:border-primary w-full resize-none rounded border px-2 py-1 text-[11px] outline-none"
/>
```

It's rendered live via `RailEventRow.tsx`. Its placeholder falls back to the browser default (derived from the near-white `color`), which is exactly the "empty field looks filled" symptom of #673 that this PR frames as fixing app-wide. Please either add `placeholder:text-input-placeholder` here, or (preferably, per the `library-first` rule) route this field through the shared `Textarea` primitive. If it's intentionally out of scope, please narrow the "app-wide" framing in the description.

_(The only other raw `<input>`, in `ImportSpecDialog.tsx`, is a hidden `type="file"` element with no placeholder — correctly irrelevant.)_

**2. (Nit) No test added.** The checklist leaves "Tests added/updated" unchecked. A token-only change is low-risk, but a small assertion that `Input`/`Textarea` render `placeholder:text-input-placeholder` would lock in the fix cheaply and matches the `library-first` guidance that primitives carry a vitest suite. Optional.

### Verification notes
Did not re-run `npm run build`/`lint` (would need a fresh `npm install` in the worktree); the change is statically verifiable and the described build/lint + compiled-CSS check cover it. Token wiring is correct by inspection.

### Recommendation
Approve once the uncovered `DenyReasonField` textarea (finding #1) is addressed or explicitly deferred. The core change itself is clean and correctly scoped.

Manuel-Jentic (approved):

PR #756, opened 2026-07-22
fix(broker): parse permission-rule JSON columns on the SQLite read path

TilmanSchaefer:
## Summary

The rule evaluator reads toolkit_permission_rules via raw text() SQL, which bypasses the ORM's json_variant() deserialization. On PostgreSQL the JSONB driver still decodes methods/operations into native lists, but on SQLite (JSON stored as TEXT) the raw string comes straight through — e.g. '["GET", "POST"]' or the literal 'null'. _normalize_methods / tuple() then iterate the string character-by-character, so a legitimate allow rule silently fails to match and every operation is wrongly denied.

## Changes

Add _coerce_json_list to parse the string form (mirroring _as_scope_list in token_resolver), so both backends yield the same list. An already-decoded list passes through unchanged; 'null'/invalid maps to None (no constraint).

## Checklist

### DIFF
--- a/src/jentic_one/broker/repos/rule_evaluator.py
+++ b/src/jentic_one/broker/repos/rule_evaluator.py
@@ -11,6 +11,7 @@ Performance: the rule list per (toolkit, credential) pair is short-TTL cached
 
 from __future__ import annotations
 
+import json
 import re
 import time
 from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


ren-jentic (approved):
nice fix! lgtm

Manuel-Jentic (approved):
## Review — Approve

Checked out the branch in a worktree, read the full evaluator + the `toolkit_permission_rules` ORM schema + the `_as_scope_list` sibling it mirrors, and ran the tests (all 34 pass).

### The bug & fix — correct
Accurate diagnosis. `toolkit_permission_rules.methods` / `.operations` are `json_variant()` columns (`JSONB` on Postgres, generic `JSON` = TEXT on SQLite). The evaluator reads them via raw `text()` SQL (it must — the broker can't import control ORM), which bypasses the ORM's deserialization. On SQLite the value arrives as a raw string (`'["GET","POST"]'` or `'null'`), so `_normalize_methods` / `tuple(...)` iterated it character-by-character, making a legitimate `allow` rule silently fail to match → everything denied. `_coerce_json_list` fixes both columns; `path` is a `String` column and correctly left untouched.

### What's good
- **Root cause, not symptom** — coerces at the raw-SQL boundary, exactly where ORM deserialization is bypassed.
- **Complete in scope** — searched the broker for other raw `text()` reads of these JSON columns; `rule_evaluator.py` is the only one, so no sibling fix is missing.
- **Dialect-agnostic / idempotent** — already-decoded lists (PG JSONB) pass through unchanged; SQLite strings get parsed. Both backends converge.
- **Good coverage** — helper unit tests (None, decoded-list passthrough, SQLite string, `'null'`→None, invalid→None) plus a real end-to-end regression test feeding the SQLite wire form (`'[...]'` methods + `'null'` operations) and asserting both the allow-match and a non-match.

### Minor nits (non-blocking)

1. The code advertises itself as mirroring `_as_scope_list`, but the two deliberately differ: `_as_scope_list` returns `[]` on failure, `_coerce_json_list` returns `None`. That's the *correct* choice here (`None` = "no constraint / match-all"; `[]` would mean "match nothing"), but a half-line comment noting the intentional divergence would stop a future reader from "reconciling" them.
2. The end-to-end test mocks the DB session (`MagicMock`/`AsyncMock`). It follows the file's existing pattern, but a SQLite-fixture integration test would more faithfully guard the real driver behaviour. Pre-existing convention gap, not introduced here.
3. `methods` invalid-JSON → `None` broadens matching (no method constraint). Safe/defensible for a documented edge that shouldn't occur; noting for completeness.

### Recommendation
Approving — correct root-cause fix, complete in scope, well tested, follows the established sibling pattern. The nits are optional polish.

Issue #686, opened 2026-07-20
[api] Agent can read its toolkit binding IDs but not toolkit names — forces ID-only, error-prone human instructions


elena-anagn:
### What happened

An agent can read the **ID** of the toolkit it's bound to, but not the toolkit's **name**. `jentic access whoami` returns bindings as `{toolkit_id, bound_at}` — no name — and the toolkit *list* endpoint (`GET /toolkits`) returns empty for the agent token (owner-scoping: the agent's `owner_id` is null, so owner-scoped list reads see nothing). So there is no way for the agent to resolve `tk_…` → a human-readable name.

### Why it was a pain point

When an agent guides a human through setup, it can only refer to a toolkit by opaque ID (`tk_6a4b921f36c0ea01a83c4837`). A human looking at the console sees toolkits by **name** ("Design news radar"), not by ID — so ID-based instructions are hard to match, especially with more than one toolkit. In this session it also caused a concrete failure: unable to read the name, the agent stated a *wrong* name in its setup instructions, which the operator then couldn't find on screen. Exposing the name would both improve instruction quality and remove the temptation to guess.

### Steps to reproduce

1. Bind an agent to a toolkit.
2. As that agent, call `jentic access whoami` → bindings show `toolkit_id` only, no name.
3. Call `GET /toolkits` with the agent token → empty list (owner-scoped).
4. There is now no agent-side way to turn the bound `tk_…` into its display name.

### Expected

The agent's readable surface should expose the toolkit's **name** alongside its ID — e.g. include `name` in the `whoami` toolkit bindings, and/or let the agent read details of a toolkit it is bound to (`GET /toolkits/{id}` for its own bindings). Then an agent can say "add it to your 'Design news radar' toolkit" instead of a `tk_…` ID.

### Actual (paste exact output/errors)

```shell
# whoami — ID only, no name:
"toolkit_bindings": [{ "toolkit_id": "tk_6a4b921f36c0ea01a83c4837", "bound_at": "2026-07-06T11:32:39Z" }]

# GET /toolkits with the agent token — empty:
{"data": [], "has_more": false, "next_cursor": null}
```

### Environment
- Jentic One version/commit: `main` @ `4098999`
- OS: macOS (Darwin 25.5.0)
- Install method: `install.sh` (docker mode, postgres)

---
[...]

Issue #656, opened 2026-07-15
Credential vendor must match the Registry's normalized slug or PBAC silently default-denies


Killian-Jentic:
## Summary

When an API is registered, the Registry **normalizes the vendor** (dots become dashes: `httpbin.org` → `httpbin-org`, `github.com` → `github-com`). But `POST /credentials` stores whatever `api.vendor` you supply verbatim. If you use the raw domain, the broker's PBAC rule query — which joins on the credential's `api_vendor` against the *discovered* (normalized) vendor — returns zero rows, so the toolkit falls through to default-deny `403`.

## Repro

1. Register an API with vendor `httpbin.org` → stored normalized as `httpbin-org`.
2. `POST /credentials` with `api.vendor: "httpbin.org"` (raw domain) → stored as `httpbin.org`.
3. Create toolkit, bind credential, set an `allow .*` binding rule.
4. `jentic execute` → `403 action_denied`, despite the permissive rule.
5. Recreate the credential with `api.vendor: "httpbin-org"` (normalized) → same call now `200`.

The failing join (broker `rule_evaluator`):

```sql
JOIN credentials c ON c.id = tcb.credential_id
WHERE tpr.toolkit_id = :toolkit_id AND c.api_vendor = :api_vendor
```

`:api_vendor` is the discovered operation's normalized vendor (`httpbin-org`); the credential stored `httpbin.org` → no match → no rules → default-deny.

## Why it's costly

Nothing flags the mismatch. Discovery succeeds, toolkit selection succeeds, then PBAC denies with the generic "denied by a toolkit permission rule" — which reads as a rule problem, not a vendor-string problem. You can have a correct-looking `allow .*` rule and still be denied because the credential's vendor doesn't match the normalized form the join uses. I only found it by dumping the credential row next to the registered API and noticing dot vs dash.

## Suggested fixes (any of)

- Normalize `api.vendor` on `POST /credentials` the same way the Registry does, so it can't drift from the discovered vendor.
- Validate at credential-create time that the vendor matches a registered API's normalized vendor, and reject/warn otherwise.
- Document the normalization rule (with the dot→dash example) wherever `vendor` is an input.

## Notes

Found while wiring a credential + toolkit end-to-end for a real broker integration.

PR #757, opened 2026-07-22
feat(access-requests): provisioning-plan access request (backend + CLI + UI wizard)

### LINKED ISSUE #603: [surface] APIs that don't need credentials seem to ask for credentials
### What's broken

the api: open-metro-com does not need credentials: https://open-meteo.com/ 
Free weather API — forecasts to archives, no sign-up
I was trying to add this API to my toolkit and couldn't figure out how. When I went to try 'add credentials' for this it defaulted to 'bearer token' - I  don't think I could set up this API correctly to get it in to my tool kit because it does not require credentials (for free tier). 

having an API that does not need a credential set up in order to get to your first execution would be great. 
As a user I just wanted the satisfaction of knowing that I'd installed correctly and that it was all working and connected to my agent. I was given the carrot of checking the weather and then it took too long for me to figure this out. 

### Steps to reproduce

1. go to credentials 
2. add credential
3. open-metro-com
4. try to add this to a toolkit without having credentials set up

### Expected

_No response_

### Actual (paste exact output/errors)

```shell

```

### Environment

- Jentic One version/commit:
- OS:

---
*Migrated from the internal dogfooding tracker (feedback#14, originally filed 2026-07-02). Screenshots/attachments referenced in the original will be re-attached.*

### LINKED ISSUE #606: [surface] API - need to know that you need to set rules to get it to work.
### What happened

Did not know any of this needed to be done here

### Where (command or page)

_No response_

### Why it was a pain point

_No response_

### What you expected instead

_No response_

### Environment

- Jentic One version/commit:
- OS:

---
*Migrated from the internal dogfooding tracker (feedback#17, originally filed 2026-07-02). Screenshots/attachments referenced in the original will be re-attached.*

### LINKED ISSUE #607: [surface] Unclean how to bind a toolkit to an agent
### What happened

Looked at the 'agent section' 

Looked at the agent associated with my terminal 

Now need to go to 'toolkit' instead and 'link agent'

Wasn't clear. 

### Where (command or page)

_No response_

### Why it was a pain point

_No response_

### What you expected instead

_No response_

### Environment

- Jentic One version/commit:
- OS:

---
*Migrated from the internal dogfooding tracker (feedback#19, originally filed 2026-07-02). Screenshots/attachments referenced in the original will be re-attached.*

### LINKED ISSUE #614: [surface] Approving an access request - unclear and difficult - flow to successfully add an API to a toolkit explored
### What happened

I have been back and forth with my agent as to why it is unable to access APIs that were in that toolkit. Md attached of the agent transcript. 

*(attachment from the original report, not carried over: "Aisling Claude - Jentic Set Up transcript .md")*

I eventually landed on this screen accidentally asking for approval to access newsapi. I thought I found it in the 'needs attention' section. I didn't - it was in the 'access requests awaiting review'. It took me a long time to figure out that's where it was. I had 'x'd out of the approval window for something else and it took me time to navigate back and find it again.

Before finding it among the many available tabs, I went to the agent rail and manually searched 'agent access'. 

When I finally found it in the dashboard it was a poor experience. The section gives no obvious explanation of what each request is asking for in the dash itself and it does not give a way to do the approval in that section: 

When I do click 'view more' I cannot see the full name of the API (it should be 'google calendar') but the name is too long and does not show. 

I then select 'approve' - I have only been shown 1 item to approve so a single 'approve' click should be enough. Instead it shows a holding screen of 'all items processed' - I know there are other approvals waiting so I don't know why it is not showing me those or indicating that there would have been more than one item for approval (perhaps there are instances where there actually one we several approvals in one place). 
I have to do a second click to 'review and submit' (what am I reviewing?) 

now on to screen 3 for this approval decision where none of the previous steps have offered me any information that would actually lead me to 'cancel' my approval process. 3 actions for a single approval with no additional information is excessive and makes the user feel as if they are missing something.
screen 3 is also the first time I see the full name of the api I am actually approving. 

It does not even work in the end and says it cannot grant access: 

The reason it cannot grant access is given here: 

'googleapis-com/googleapis-com-calendar

No toolkit serves API googleapis-com/googleapis-com-calendar; provision and bind a credential for it first'

I now have to 'x' and navigate back to the credential set up section - having thought after doing the OAuth set up and not being made aware of any additional steps needed, and having successfully 'added' this API to a toolkit without error - I now need to go figure out what is wrong. 
This is the screen I arrive at: 

I assume I need to 'connect' - i do this - and I OAuth again ... having already done this before during the credential set up. 

I am then given this screen: 

I opt to 'close window' to see what happens. Nothing happens. I am not sure what the point of this option is. I now try the option to 'return to credentials'. It brings me back to the credentials page - which I think is interesting as I was working on the google calendar API and I would expect an end or close to the flow - or some sort of confirmation that this API is now set up and ready to go. 

I then go through a process to try and get newsapi up and running and this hits a different bug. During the approval process it gets me to the end and says I can't grant access. I go back to the credentials set up and add a new credential, however, I no longer have the option of going through the approval flow for newsapi now. I need to figure out a way to get it working. 
I navigate back to the credentials set up and to newsapi and I see there is no longer an api key in there. I do not know if the key I added was not accepted. I try a different key. I look for approval there is no where to do this. I navigate back to the credentials for that API and again the new key I have tried is also not appearing. Perhaps it is not meant to show up (even as dots) but it is so confusing as there is nothing to confirm to a user that an API key have actually been saved here. 

I went and spoke to my agent who did the access requests again and they both did not work again. 
I will need to do more trouble shooting and create separate issues. 

### Where (command or page)

_No response_

### Why it was a pain point

If I cannot get an API set up I cannot use Jentic. I am currently unable to use Jentic. 

### What you expected instead

_No response_

### Environment

- Jentic One version/commit:
- OS:

---
*Migrated from the internal dogfooding tracker (feedback#37, originally filed 2026-07-08). Screenshots/attachments referenced in the original will be re-attached.*

### LINKED ISSUE #619: [control] Approving access to an API no toolkit serves is a hollow yes — "approve access to <API>" is a category error
### What's broken

The approval flow lets a user "approve access to an API" (e.g. *newsapi*, *google calendar*) — but **an agent never binds to an API, it binds to a *toolkit***. So there is nothing coherent to approve at the API level, and approving does **not** create the thing that actually makes an API callable: *a toolkit that serves that API, with a bound credential*.

The result is a **hollow yes**: the request flips to `approved`, and then `execute` still fails with `no_toolkit_binding` / "No toolkit serves API <x>; provision and bind a credential for it first." The user has done the one action the UI offered, been told it succeeded, and is still blocked — with no indication of the real missing step.

This is the concept-level root cause underneath several already-filed issues:
- feedback#37 (internal tracker) — approving Google Calendar access → "all items processed" → then "cannot grant access: no toolkit serves this API."
- feedback#23 (internal tracker) — the `no_toolkit_binding` recovery directive tells you to bind a toolkit *before a credential exists*, so approval is always denied.
- feedback#38 (internal tracker) — the agent files *per-API* access requests and then revokes them: it is flailing, trying to express "I need this specific API" in a model that only understands "bind me to a toolkit."

### Why it's a category error

- **Agent → toolkit** is the only real binding. "Access to *newsapi*" is not an object the system can grant.
- Approval today is a UI state change with **no side effect** on the served-APIs / credential / rule graph, so it can be granted for a combination that can never succeed.

### Expected

One of two coherent designs (pick one):
1. **Approval is impossible until the API is genuinely served + credentialed** by a toolkit — i.e. you can only approve a request that *will* work once approved. The UI should never present an un-satisfiable approval.
2. **Approving performs the provisioning** — approving "give my agent access to <API>" auto-ensures a toolkit serves that API, a credential is bound, and a sane default rule exists, then binds the agent. Approval becomes the action that makes it true, not a label asserting it already is.

Either way, the approval card should speak in the real unit ("bind **[agent]** to **[toolkit]**, which serves these APIs") or fully abstract the toolkit away — but it must not offer a yes that resolves to a no.

### Actual

`execute` after "approval":
```
"type": "no_toolkit_binding", "status": 403
"decision_reason": "No toolkit serves API <x>; provision and bind a credential for it first"
```

### Environment
- Jentic One version/commit: (see feedback#37 (internal tracker) / feedback#25 (internal tracker) — `main`)
- OS: macOS
- Install method: install.sh (docker)

---
*Migrated from the internal dogfooding tracker (feedback#51, originally filed 2026-07-09). Screenshots/attachments referenced in the original will be re-attached.*

### LINKED ISSUE #621: [skill] Agent should read the API spec and translate plain-English intent into permission rules (propose → human approves → system applies)
### The opportunity (this is where Jentic can shine)

Toolkit permission rules are authored as raw `effect + methods + path`/`operations` — e.g. `allow POST,PUT /repos/{owner}/{repo}/**`. **No normal user can hand-write these**, and getting them subtly wrong = silent 403s or over-broad access. Yet expressing the *intent* is easy in English.

Real example (Jira): a user wants their agent to
1. **write only to the prod board**,
2. **read everything else**,
3. **write nowhere else**.

Those are three precise rules against Jira's API surface that almost nobody can author by hand — but they're trivial to *say*.

### Why the agent can do this today (no new backend needed)

The agent already has `apis:read` and can pull an API's spec and operation list (`GET /apis/{v}/{n}/{ver}/operations`, `.../openapi`, `.../security-schemes`). So the skill can teach the agent to:

1. Take the user's plain-English intent.
2. Read the target API's spec (operations, methods, paths, resource shapes, security schemes).
3. **Return the exact permission rules to insert** (allow/deny + methods/path/operations), in human-readable form for the user to confirm.
4. **Flag conflicts** — e.g. "you asked to write, but the OAuth scope you chose is read-only; that rule can never fire" (see the scope-vs-rule contradiction issue).
5. On the human's yes, the rules get applied (agent proposes → human approves → system/operator applies; the agent must never grant its own access — it stays read-only).

### Why file this now

We're about to design a full onboarding flow + product changes. This capability is the load-bearing primitive underneath the whole "state intent once, derive the permissions" direction, and it is the single highest-value thing we can add to the `jentic` skill. Capturing it here so it isn't lost in the larger redesign.

### Acceptance sketch
- Skill instructs the agent: on a permission/scoping task, fetch the spec and propose concrete rules from NL intent.
- Output is human-legible ("read everything, write only to jentic/prod-board") with the underlying rules shown.
- Agent detects and surfaces scope/rule contradictions before the human approves.
- Human approves; agent never self-applies write.

### Environment
- Jentic One: self-hosted, 2026-07-10. Filed from a dogfooding UX session.

---
*Migrated from the internal dogfooding tracker (feedback#54, originally filed 2026-07-10). Screenshots/attachments referenced in the original will be re-attached.*

### LINKED ISSUE #629: [surface] Unclear what access request is actually for
### What happened

It's unclear to me if the agent is requesting to access the API or bind the API to a specific toolkit in order for it to access the API via the toolkit. 
I can see the decision includes 'bind' although it is very hard to understand this from the UX. But, there is no mention of 'what' toolkit the API will be bound to.

The UX leads with 'Agent is requesting access' 
Reason “Fetch today's US top headlines for the user”
(reason is very hidden too from a UX perspective) 
This gives the impression that the agent is trying to access an API directly, without it being in a toolkit - which for a new user would seem reasonable. 
but then the UX proceeds to actually ask me to approve the 'binding' of the api to an unspecified toolkit. 
It appears we are getting the users permission to bind an api to a toolkit > then the user confirms > and the access is unable to be granted. with the note: 'No toolkit serves API newsapi-org/newsapi-org; provision and bind a credential for it first'. 
Leaving me more confused about why 'bind' featured in the UX alongside the 'approve' button.

### Where (command or page)

_No response_

### Why it was a pain point

Very unclear and seeming to have the user approving something that could not be approved anyway because previous required set up steps had not been done (eg. credential set up + bind to toolkit). 
Unsure why the agent can even ask this permission and why this flow exists if once an api does have a credential and is bound to a toolkit it doesn't need to ask for approval to access the API. 

### What you expected instead

A user experience that is incredibly simplistic and leaves very little to for the human user to have to decipher.  

### Environment

- Jentic One version/commit:
- OS: MAC
- Install method (install.sh / source / Docker): Docker

---
*Migrated from the internal dogfooding tracker (feedback#67, originally filed 2026-07-14). Screenshots/attachments referenced in the original will be re-attached.*

### LINKED ISSUE #662: [pain-point] no_toolkit_binding directive steers the agent to a bind request that auto-denies — should point at credential provisioning first
### Suggested labels (for maintainer triage)
- type: pain-point
- area: broker 
- severity: major

---

## What happened

Working as the default agent (via the `jentic` CLI skill), I started from a plain natural-language goal — **"I want to check progress on my PostHog dashboards"** — and the agent flow ran exactly as the skill instructs:

1. `jentic search "list dashboards"` → empty (nothing imported yet)
2. `jentic catalog search "posthog"` → found `posthog.com/posthog-api`
3. `jentic catalog import posthog.com/posthog-api` → imported
4. `jentic search "list dashboards"` → found the operation
5. `jentic inspect <op>` → `GET /api/projects/{project_id}/dashboards`
6. `jentic execute <op> …` → **403 `no_toolkit_binding`**, with an `agent_directive` telling me to run `jentic access request --toolkit posthog-com/posthog-com-posthog-api --wait`
7. I followed that directive and filed the toolkit `bind` request. It was **auto-denied** with `decision_reason: "No toolkit serves API posthog-com/posthog-com-posthog-api; provision and bind a credential for it first"`.

**I had never created a PostHog credential.** That — creating/connecting a credential and having a toolkit serve the API — is the actual prerequisite, and it is the *first* thing I should have been steered toward. Instead the flow led me confidently through search → inspect → execute and then into filing an **access grant request that could never be approved**: it named no real credential value, referenced an unspecified toolkit, and was guaranteed to fail. The "request access → wait for a human to approve" ceremony is confusing and hollow when the thing being approved cannot possibly succeed.

## Where

Command: `jentic execute op_888995ac0c402a22bfdbe2aa853dee73a364bd8f --path project_id=@current --broker-scheme http --broker-host 127.0.0.1:8100`
then the recovery command the directive suggested: `jentic access request --toolkit posthog-com/posthog-com-posthog-api --wait`

## Steps to reproduce

1. As an approved agent with no PostHog toolkit/credential, import the PostHog API: `jentic catalog import posthog.com/posthog-api`.
2. Find and execute any PostHog operation, e.g. list dashboards.
3. Observe the 403 `no_toolkit_binding` directive pointing at `jentic access request --toolkit …`.
4. Run exactly that suggested command with `--wait`.
5. Watch it resolve to `denied` — "No toolkit serves API …; provision and bind a credential for it first".

## Expected

At step 3, when **no toolkit serves the API at all** (and therefore no credential is connected), the agent-facing directive should point at the **real first step — provision/connect a credential for this API** (the "connect account" / `provisioning_url` flow) — not at `jentic access request --toolkit …`. A `toolkit:bind` is the *last mile*; suggesting it first sends the agent down a path that is **guaranteed to auto-deny**, and drags the human through an approval that has no coherent yes.

Ideally the flow never asks a human to approve a request that cannot be fulfilled: either the directive changes first step to credential provisioning, or the access request is refused up front with a "connect a credential first" message rather than being fileable-then-auto-denied.

## Actual

The `execute` denial directive:

```json
{
  "status": 403,
  "type": "no_toolkit_binding",
  "title": "No toolkit binding for this API",
  "agent_directive": {
    "strategy": "prompt_human",
    "human_readable_instruction": "You are not bound to a toolkit for 'posthog-com/posthog-com-posthog-api'. File an access request yourself with `jentic access request --toolkit posthog-com/posthog-com-posthog-api --wait`, then ask your operator to approve it — only a human can grant the binding. Once approved, retry this call.",
    "parameters": {
      "suggested_command": "jentic access request --toolkit posthog-com/posthog-com-posthog-api --wait"
    }
  }
}
```

The resulting auto-denied access request:

```json
{
  "id": "areq_***",
  "status": "denied",
  "items": [
    {
      "resource_type": "toolkit",
      "action": "bind",
      "resource_reference": { "name": "posthog-com-posthog-api", "vendor": "posthog-com" },
      "status": "denied",
      "decision_reason": "No toolkit serves API posthog-com/posthog-com-posthog-api; provision and bind a credential for it first"
    }
  ],
  "evaluation": { "can_fulfill": false }
}
```

## Likely root cause (from reading the code)

In the broker, `select_toolkit()` runs *before* credential resolution (`src/jentic_one/broker/web/routers/execute.py`). So the "no toolkit **and** no credential" case can only ever surface as **403 `no_toolkit_binding`** — it never reaches the **424 `credential_not_provisioned`** path that actually carries a `provisioning_url` (`src/jentic_one/broker/services/credentials/orchestrator.py`, `_not_provisioned()`). The directive builder `no_toolkit_binding_directive()` (`src/jentic_one/broker/core/exceptions.py`) is static — it *always* emits `jentic access request --toolkit <api> --wait` regardless of whether any toolkit/credential exists. The `toolkit:bind` request is then auto-denied by `ToolkitReferenceUnresolvedError` (`src/jentic_one/control/services/access_requests/effects.py` → auto-deny in `.../service.py`), which is the source of the "provision and bind a credential for it first" string.

Net: the 403 directive points the agent at the *last* step (bind) when the *first* step (provision a credential so a toolkit serves the API) is what's missing — and the only place this chicken/egg is explained is the skill doc (`cli/internal/skillgen/content/jentic.md`), not the directive itself.

## Related issues

- #619 — approval flow is a "hollow yes" (control/UI conceptual root cause).
- #624 — no zero-config path to a first successful call.
- #629 — approval UI wording (bind vs access) is confusing.

This issue is the **agent-facing / broker-directive** angle: driven by a natural-language prompt, the 403 recovery directive steers the agent to file a bind that is guaranteed to auto-deny, instead of pointing at credential provisioning as the real first step.

## Environment

- Jentic One version / commit: server `0.1.0`, docker source `jentic/jentic-one@main` (`4098999`); local repo checkout at `31c90a7` (branch `add-slack-feedback-notify`)
- OS: macOS 15.7.4 (24G517)
- Install method: Docker (`jenticctl`)

### LINKED ISSUE #684: [control] Making one imported API executable takes 5 manual, silently-failing steps (credential → toolkit → bind cred → bind agent → allow-rule)
### What happened

Getting a single imported API (NewsAPI / `newsapi-org`) from "imported" to "executable by my agent" required **five separate manual steps**, each of which fails silently or with a cryptic 403 until the *next* one is also done. There is no single guided flow, and the error messages point at the wrong next step (see #23).

The actual required chain — none of it auto-linked — turned out to be:

1. **Create a credential** (api_key), tying it to the exact API tuple (`newsapi-org` / `newsapi-org` / `1.0.0` — not the catalog id `newsapi.org`).
2. **Create a toolkit** — a *separate* object. Provisioning a credential does **not** create one.
3. **Bind the credential to the toolkit.**
4. **Bind the agent to the toolkit** (access request → operator approval).
5. **Add at least one `allow` permission rule** to the binding — toolkits are **default-deny**, so with zero rules every call is blocked.

Along the way, the broker directive told me to do step 4 first (it can't succeed before 1–3 exist; #23), and the failure mode changed at each stage: `no_toolkit_binding` (403) until steps 2–4 were done, then `action_denied` (403) until step 5.

### Steps to reproduce

1. `jentic catalog import newsapi.org`
2. `jentic execute "GET:https://newsapi.org/v2/top-headlines" --query country=us` → `403 no_toolkit_binding`, directive says to file a toolkit-binding request.
3. File it, operator approves → **denied**: "No toolkit serves API newsapi-org; provision and bind a credential for it first."
4. Add a credential in the UI → still denied on retry (no toolkit exists yet).
5. Manually create a toolkit, bind the credential, bind the agent → retry → `403 action_denied`: "denied by a toolkit permission rule."
6. Add an `allow` rule to the binding → retry → finally `200 OK`.

### Expected

A single, guided path from "I imported this API and have a key" to "my agent can call it" — ideally: provisioning a credential for an API auto-creates (or offers to create) its toolkit; the agent-binding request is fileable and satisfiable in the right order; and a new binding ships with a sensible default (e.g. an allow-all-reads rule, or a clear prompt to add one) instead of silent default-deny. At minimum, the recovery directives should name the *actual* next step at each stage rather than sending the user into a guaranteed denial.

### Actual (paste exact output/errors)

```shell
# Stage A — before toolkit exists:
"type": "no_toolkit_binding", "status": 403
"decision_reason": "No toolkit serves API newsapi-org/newsapi-org; provision and bind a credential for it first"

# Stage B — bound but no allow rule:
"type": "action_denied", "status": 403
"human_readable_instruction": "You are bound to a toolkit for this API, but the rule forbids this specific call — ask your operator to adjust the toolkit's permission rules."

# Stage C — after adding an allow rule (effect=allow, methods/path/operations empty):
"status": 200, "api_status": "ok", "totalResults": 35   # success
```

### Environment
- Jentic One version/commit: `main` @ `4098999`
- OS: macOS (Darwin 25.5.0), zsh
- Install method: `install.sh` (docker mode, postgres)

---
_Migrated from internal Jentic One dogfooding feedback (original ref #25). Possible duplicate of: #644, #624. Suggested classification: type:pain-point | area surface:control | severity:major. Labels omitted per CONTRIBUTING; the intake assistant applies them._

### LINKED ISSUE #738: Approving a toolkit-bind request fails when no toolkit/credential exists yet ('Could not grant access')
## Summary
An operator can approve an agent's **toolkit-bind** access request for an API that has **no toolkit + credential provisioned yet**. The approval then fails with a confusing terminal error instead of being prevented or guided.

## What happened
Agent filed `jentic access request --toolkit sendgrid-com/sendgrid-com-mc-singlesends`. On approving it in the dashboard, the UI showed:

> **Could not grant access** — You approved this, but the platform could not fulfill it as filed:
> `sendgrid-com/sendgrid-com-mc-singlesends`
> No toolkit serves API sendgrid-com/sendgrid-com-mc-singlesends; provision and bind a credential for it first

So the operator is asked to approve something that can never succeed, with no upfront indication that a prerequisite (create toolkit → bind credential) is missing.

## Impact
Dead-end loop: file request → approve → fail → unclear what to do next. The required ordering (toolkit + credential must exist **before** an agent bind) is not discoverable.

## Ask
- Block or clearly annotate bind requests when no toolkit exists for the API.
- Guide the operator through **provision toolkit → bind credential → then approve agent bind**, or offer to create the toolkit inline at approval time.

---
*Filed from a Claude Desktop + Jentic One session.*

## Screenshot (operator dashboard)
The failure as shown in the Jentic One dashboard when approving the request:

> **Access request**
> ⚠️ **Could not grant access**
> You approved this, but the platform could not fulfill it as filed:
> > `sendgrid-com/sendgrid-com-mc-singlesends`
> > No toolkit serves API sendgrid-com/sendgrid-com-mc-singlesends; provision and bind a credential for it first

### LINKED ISSUE #775: [broker] Credential with empty-string api_version/api_name never matches a concrete API — no_toolkit_binding despite valid bindings
## Summary

After a **fully-approved** provisioning plan (create toolkit → provision credential → bind credential → bind agent), `jentic execute` still returns **`no_toolkit_binding`** ("Nothing serves '<vendor>/<name>'") even though the agent→toolkit and credential→toolkit bindings both exist. The cause is a wildcard-matching gap in the broker's runtime toolkit resolver: it treats only `NULL` as the "any version/name" wildcard, but a credential created without a version/name is persisted with an **empty string (`''`)**, which matches neither `NULL` nor a concrete value.

Surfaced by the no-auth provisioning path (the wizard creates a `NO_AUTH` credential with no version), but it is **not** no-auth-specific — any versionless (or nameless) credential is affected.

## Reproduction

1. Import an API that resolves to a concrete version, e.g. `country.is` → `country-is/country-is/4.2.3`.
2. File + fulfil a provisioning plan for it (no-auth is the easiest trigger: `jentic access request --provision country-is/country-is --auth none --rules-json '[{"effect":"allow","methods":["GET"],"path":".*"}]' --wait`).
3. Approve it through the wizard — the request goes to `approved`, and all bindings are created.
4. `jentic execute <op>` → **403 `no_toolkit_binding`**: "Nothing serves 'country-is/country-is' yet."

## Verified DB state (Docker `jentic-db-1`, Postgres)

The wiring is all present and correct:

- `control.credentials`: `api_vendor=country-is`, `api_name=country-is`, **`api_version='' (empty string, NOT NULL)`**, `type=NO_AUTH`
- `control.toolkit_credential_bindings`: toolkit ↔ credential row exists
- `admin.agent_toolkit_bindings`: agent ↔ toolkit row exists

## Root cause

The runtime resolver `src/jentic_one/broker/repos/toolkit_binding_resolver.py` (`_TOOLKITS_FOR_API`) only treats `NULL` as the wildcard:

```sql
SELECT DISTINCT tcb.toolkit_id
FROM toolkit_credential_bindings tcb
JOIN credentials c ON c.id = tcb.credential_id
WHERE c.api_vendor = :vendor
  AND (:name = '' OR c.api_name IS NULL OR c.api_name = :name)
  AND (:version = '' OR c.api_version IS NULL OR c.api_version = :version)
```

With the operation resolving as `version=4.2.3` and the credential's `api_version=''`:
- `:version = ''` → false (`:version` is `'4.2.3'`)
- `c.api_version IS NULL` → **false** (it's `''`, not `NULL`)
- `c.api_version = :version` → `'' = '4.2.3'` → false

→ the whole clause is false → **0 toolkits → `no_toolkit_binding`**.

Reproduced exactly against the live row: the query returns 0 rows as written, and returns the toolkit as soon as `OR c.api_version = ''` is added to the version clause.

The same NULL-only wildcard assumption exists in `AgentService._served_apis_by_toolkit` (`src/jentic_one/auth/services/agent_service.py`), which is why `whoami`'s toolkit binding shows no `serves` entry for the API.

## Where the empty string comes from

The credential create path (`CredentialService.create`) passes `api_version=payload.api.version` straight through; a missing version arrives as `''` (not coerced to `NULL`). Note the *bind-time* resolver (`control/repos/effects_repo.resolve_toolkits_for_api`) tolerates this — which is why **approval succeeds** — but the *runtime* resolver + serves query do not, so **execute** and **`whoami serves`** fail. That asymmetry is what makes this confusing: approval looks successful, execution still 403s.

## Suggested fix

1. Treat `''` like `NULL` (the wildcard) in `toolkit_binding_resolver.py` and `_served_apis_by_toolkit` — robust, and fixes credentials already stored with `''`.
2. Coerce empty `api_name` / `api_version` → `NULL` on credential create so data is clean going forward.

Do both: (1) unblocks existing rows, (2) prevents new ones.

## Related

- #747 — same *symptom* (`no_toolkit_binding` despite a valid binding, due to a credential-field mismatch), but that issue is about the misleading remediation message; this is the specific matching-logic cause underneath it.
- #746 — another cause of the same resolver-mismatch symptom (vendor/name/version mis-parsed from the spec file path).
- #656 — vendor-slug normalization (the `api_vendor` axis of the same "credential identity must match the resolver" family).

Found while testing the provisioning-plan flow (PR #757). PR #757's no-auth path is correct — it exposed this pre-existing resolver wildcard gap; the fix belongs in the broker resolver, separate from that PR.

Manuel-Jentic:
## What this changes (the one-paragraph version)

Today, getting an agent from "I found an API" to "I can call it" takes **five** manual, silently-failing steps (create a toolkit, create a credential, bind the credential to the toolkit, add a permission rule, bind the agent) — and approving an access request for an API that nothing serves yet is a **hollow yes**: the operator clicks approve and it still fails at execute. This PR turns the access-request endpoint into a **provisioning plan**: the agent declares the *whole* path to first execution in one request, and a human fulfils it through a guided wizard that reuses the existing credential + permission-rule forms. **Approval becomes the action that makes access true.**

Design doc: [`docs/plans/provisioning-plan-access-request.md`](https://github.com/jentic/jentic-one/blob/feat/provisioning-plan-access-request/docs/plans/provisioning-plan-access-request.md).

---

## The access-request flow

A `--provision` request is an *envelope* of four (or three, for a no-auth API) items. The agent proposes; a human fulfils + approves; the existing bind effects do the real wiring.

```mermaid
sequenceDiagram
    autonumber
    participant Agent
    participant Broker
    participant Control as Control API
    participant Human as Operator via UI wizard

    Agent->>Broker: execute, API not yet served
    Broker-->>Agent: 403 no_toolkit_binding + directive to file a --provision plan

    Agent->>Control: access request --provision vendor/name --auth type or none --rules-json --reason
    Note over Control: Plan items, inert intents plus binds:<br/>1. toolkit:create<br/>2. credential:provision, skipped if no-auth<br/>3. credential:bind, with proposed rules<br/>4. toolkit:bind, agent to toolkit
    Control-->>Agent: pending, agent --wait blocks

    Human->>Human: Open the setup wizard
    Human->>Control: create real toolkit
    Human->>Control: create real credential, or auto-create NO_AUTH. secret never leaves the human
    Human->>Control: amend resolved ids plus confirmed rules onto the binds
    Human->>Control: decide, approve all pending items
    Note over Control: credential:bind plus toolkit:bind effects wire<br/>the binding, rules, and agent to toolkit link
    Control-->>Agent: approved, binding live immediately

    Agent->>Broker: retry execute
    Broker->>Control: resolve toolkit + credential for the API
    Broker-->>Agent: 200, call succeeds
```

Key invariants the code enforces:
- The two `*:create` / `*:provision` items are **inert** (`FULFILMENT_ONLY`) — the effect applicator never mutates state for them; a human fulfils them out-of-band.
- Approving a plan through the **plain** approve/deny path (without wizard fulfilment) **denies** the bind items with a plan-aware reason (`ProvisioningPlanNotFulfilledError`) instead of a hollow yes.
- Intra-decision ordering: the agent `toolkit:bind` resolves by the concrete toolkit id the wizard stamps, because the credential→toolkit binding isn't visible to the reference join until the `credential:bind` effect applies later in the same decision.

---

## What's in it, by area

- **Backend (control):** two inert intent item types (`toolkit:create`, `credential:provision`); the plan guard + legible denial; `amend` `to_id` write-back; full **NO_AUTH** credential path; shared `slugify_identifier` so credential vendor/name match the registry's normalized slug.
- **Broker:** the `no_toolkit_binding` directive now steers to `--provision` (not the auto-denying bare bind); all three resolvers (runtime toolkit-binding, bind-time, credential) treat a NULL/empty `api_name`/`api_version` as a wildcard so a versionless credential still serves its API.
- **Registry:** re-importing an unchanged spec is idempotent (reactivate the same-digest revision instead of INSERTing a duplicate that dead-lettered the job and hung the CLI); a published revision is preserved on re-import.
- **CLI + skill:** `jentic access request --provision <vendor/name> [--auth --rules-json --reason] --wait` builds the plan; the `jentic` skill teaches the flow (propose auth type + rules from the spec; human enters the secret + approves).
- **UI:** the credential form + permission-rule editor were promoted into `shared/` so the new `ProvisioningRequestDialog` wizard can compose them; `whoami` reports which APIs each toolkit binding serves.

---

## Issues closed → the fix → the commit

| Issue | What was wrong | Fix | Commit(s) |
| --- | --- | --- | --- |
| **#619** | Approving access to an API no toolkit serves was a "hollow yes" | Plan model + guard: approval now provisions, and an unfulfilled plan denies legibly | `25204b0`, `e6f9000` |
| **#684** | Making one API executable took 5 manual, silently-failing steps | One `--provision` plan + guided wizard | `25204b0`, `ab0c8c1`, `6f1b63a` |
| **#662** | `no_toolkit_binding` directive steered to a bind that auto-denies | Directive now steers to `--provision` (+ `--reason`) | `ab0c8c1`, `6de6167` |
| **#603** | No-auth APIs still asked for a credential / 422'd on create | Full NO_AUTH path incl. the `no_auth` create-schema union | `f1fd08c`, `4b4d280`, `4a82d44` |
| **#621** | Agent should propose permission rules from the spec | `--rules-json` proposed at file time, human confirms in the wizard | `ab0c8c1`, `6393665` |
| **#629 / #614 / #606** | The approval UX was unclear (which toolkit? why? need rules?) and dead-ended | Wizard names the toolkit, surfaces the reason, has an explicit rules step, and no longer ends in "Could not grant access" | `6f1b63a`, `1fdeea7`, `aa5f32f`, `c770ebd`, `6ddbd93` |
| **#607** | Unclear how to bind a toolkit to an agent | The agent→toolkit bind is the wizard's final step, done inline | `25204b0`, `6f1b63a` |
| **#738** | Approving a bind with no toolkit/credential dead-ended ("Could not grant access") | Wizard provisions the toolkit + credential inline before approving | `6f1b63a`, `e6f9000` |
| **#775** | A versionless credential (`api_version` NULL/"") never matched a concrete version → 403/424 despite valid bindings | NULL/empty treated as wildcard across all three resolvers; `""`→NULL on create | `cc9917e`, `9a079cd` |

(The **#656** vendor-slug normalization and **#641** server-variable fix were closed on earlier work of this branch; not re-closed here.)

Closes #619, closes #684, closes #662, closes #603, closes #621, closes #629, closes #614, closes #606, closes #607, closes #738, closes #775.

---

## Fixes from the multi-agent review

A parallel 7-reviewer pass over the whole PR produced these hardening fixes (all with tests):

| Finding | Fix | Commit |
| --- | --- | --- |
| Wizard reported `partially_approved` as full success | "Granted" now requires **both** binds approved; a denied bind surfaces its reason (`isPlanGranted`/`planDenialReason`) | `2474d9c` |
| Wizard trusted a stale list snapshot | Fresh `GET` on open before showing live controls | `2474d9c` |
| Failed submit stranded the operator on the error screen | Reset to the review step on reopen | `2474d9c` |
| `NO_AUTH` could fall through to the OAuth secret form | Defensive guard in `CredentialTypeFields` | `2474d9c` |
| `api_version` not normalized on credential-create (next #775-class latent bug) | `.strip()`; empty→NULL, matching vendor/name + the registry | `6de6167` |
| Re-import could demote a **published** revision | `reactivate_imported` preserves `PUBLISHED` | `6de6167` |
| Plan guard + `FULFILMENT_ONLY` phase had no unit coverage (flagged by 3 reviewers) | Added unit tests at the id boundaries + classification/skip/rules-reject | `6de6167` |
| Broker directive omitted `--reason` | Added to the suggested command + guidance | `6de6167` |

---

## Follow-ups (deferred from the review — tracked, intentionally not in this PR)

- **#778** — control: the provisioning-plan flag is request-scoped, so an intent item changes the fulfilment contract of every bind in the request.
- **#779** — registry: re-import reactivation relies on flush ordering vs the `ix_api_revisions_one_active` partial unique index.
- **#780** — registry: empty `spec_digest` (`spec.sha or ""`) can collapse distinct sha-less specs into one revision.
- **#781** — ui: a caller-driven/programmatic close of the wizard bypasses orphan cleanup and can strand a created toolkit/credential.

---

## Merged `main` (conflict resolution)

`main` moved ahead and independently reworked three areas this PR touches, so the merge (commit `703af9b`) is a deliberate semantic resolution, not a mechanical one:

- **Slug helper** — adopted main's canonical `slugify_api_field` (byte-identical to our helper); deleted our `shared/slug.py` and re-pointed every call site. One source of truth (#656).
- **Credential resolver (#775)** — kept **both** main's slug-normalized name comparison **and** our NULL/empty `api_version` wildcard, so a versionless credential still resolves at a concrete version. Both regression tests retained.
- **Registry re-import** — adopted main's `delete_replaceable_by_digest` + `DuplicateRevisionError` approach, which supersedes our `reactivate_imported` (dropped) and also resolves follow-up #779. Our re-import tests were rewritten to assert main's behavior (an active same-digest re-import raises legibly instead of hanging; a published revision is preserved).
- **`whoami`** — merged main's toolkit `name` (#686) with our `serves` list; `ToolkitBindingView` / `ToolkitBindingEntry` now carry **both**.
- **`no_toolkit_binding` directive** — kept main's `toolkit_serves_api` branching, but in the "nothing serves it yet" branch steer to `--provision` (this PR's behavior, #662) rather than `--toolkit`; skill guidance aligned.
- **Generated artifacts** — `control.openapi.yaml`, `ui/openapi.json`, and the UI client were **regenerated** from the merged code (not hand-merged).

Post-merge validation: backend unit + SQLite integration, UI browser suite, and ruff/mypy/eslint all green.

---

## Test plan

- [x] Backend unit + arch and SQLite integration green (incl. new plan-guard / `FULFILMENT_ONLY` unit tests and the no-auth end-to-end resolver test)
- [x] Go CLI suite green (incl. plan-builder + dead-letter tests)
- [x] UI build + lint + browser suite green (incl. `isPlanGranted` / routing tests)
- [x] `test_provisioning_plan_e2e.py` drives file → fulfil → amend → approve and asserts the binding, rules, agent-binding, **and** that both broker resolvers resolve it at a concrete version
- [ ] Live Docker install of this branch + skill install + manual `country.is` run (needs a container rebuild to pick up the broker resolver fix)

Merge with **squash**.


Manuel-Jentic:
### Added: AX — reduce the tool count to first execution

Layered onto this PR (kept as separate, labelled commits) to cut the CLI calls an agent needs to reach its first successful `execute`. Grounded in a dogfooding run that took 9 commands; targets the wasted ones. Plan: `docs/plans/reduce-tools-to-first-execution.md`.

- **`whoami` / `GET /me` now reports the APIs each toolkit binding serves** (`serves: [{vendor,name,version}]`) — the agent can tell it already has access without a throwaway "execute → get denied" probe. (`ApiRef`; auth service resolves via a control-DB raw-SQL query since auth must not import control ORM.)
- **`jentic access request --provision --wait` auto re-mints the token** — but only when the request granted a `scope:grant` (bindings are live server-side, so a binding-only plan needs no refresh; re-minting there would be a wasted round-trip). Removes the explicit `access refresh` step in the common case.
- **Skill rewrite**: teaches the agent to check `serves` and provision proactively before the first execute (not execute-to-get-denied); import-before-search; and narrows the refresh guidance to "only after a stale scope grant."

Net: the ideal path drops from ~9 → ~5 commands.

All suites green: backend unit+arch 2098, SQLite integration 454, Go CLI all packages, UI 691.

ren-jentic:
# Review — PR #757 `feat(access-requests): provisioning-plan access request`

**Branch:** `feat/provisioning-plan-access-request` → `main`
**Reviewer:** brutal pass, rules-checked. **Focus:** backend nitpicks. No pushes; local report only.
**Scope reviewed:** 33 commits, 90 files. Backend read line-by-line against the branch tip; CLI/UI skimmed for hacky patterns.

Rules consulted: `service-protocols`, `error-taxonomy`, `query-scoping`, `orm-overview`, `redaction`, `git-conventions`, `agent-commands`, plus the design doc `docs/plans/provisioning-plan-access-request.md`.

---

## Verdict

The design is genuinely good — the "approval becomes the action that makes access true" framing, the inert `FULFILMENT_ONLY` phase, and the plan-aware denial (`ProvisioningPlanNotFulfilledError`) instead of a hollow yes are all clean. Test coverage on the new guard/phase is solid. The merge resolution notes are honest.

But there are a handful of things I'd fix before merge, one of which is a real architectural rule violation. Nothing here is a blocker for the *design*, but the "nothing hacky / clean" bar is not fully met yet.

Severity legend: **[BLOCK]** fix before merge · **[NIT]** should fix · **[Q]** question / confirm · **[GOOD]** called out as done well.

---

## Backend

### B1. [BLOCK] Raw SQL inlined in a service — violates the repo/service boundary the file itself follows one line above

`src/jentic_one/auth/services/agent_service.py` — `_served_apis_by_toolkit()` executes `text(...)` + `bindparam(expanding=True)` **directly inside the service**:

```python
async def _served_apis_by_toolkit(self, toolkit_ids: list[str]) -> dict[str, list[ServedApi]]:
    ...
    query = text("SELECT tcb.toolkit_id, c.api_vendor, c.api_name, c.api_version ...")
        .bindparams(bindparam("toolkit_ids", expanding=True))
    async with self._ctx.control_db.session() as session:
        rows = await session.execute(query, {"toolkit_ids": toolkit_ids})
```

The `service-protocols` rule is explicit: *"Never perform direct DB calls (queries, inserts, etc.) in a service — Repositories own data access"* and *"Calling `session.execute()` directly in a service → Create a repository method for the query."*

This is doubly awkward because **the sanctioned pattern already exists in this exact code path**: the sibling call `ToolkitNameRepository.get_names_for_ids(...)` lives in `auth/repos/toolkit_name_repo.py` and is the *documented* cross-boundary raw-SQL escape hatch (its own docstring: "mirrors `control.repos.prerequisite_repo.PrerequisiteRepository`"). The new served-APIs query does the identical thing but skips the repository.

**Fix:** move the query into `auth/repos/` (e.g. `ToolkitServedApiRepository.get_served_by_toolkit` or extend `ToolkitNameRepository`), returning primitives/tuples; keep `ServedApi` construction in the service. This restores consistency and keeps the service SQL-free.

> Note: `agent_service.py` already imports `AsyncSession`/`IntegrityError` on `main`, so the auth service layer isn't perfectly SQL-pure today — but the established mitigation for *cross-boundary* reads is a repo (`ToolkitNameRepository`). This new code walks past that on purpose. That's the nit.

### B2. [NIT] Two sequential control-DB sessions where one would do (whoami)

In `list_toolkits`, the enrichment opens a control session for `get_names_for_ids`, lets it close, then `_served_apis_by_toolkit` opens **a second** control session:

```python
async with self._ctx.control_db.session() as session:
    names = await ToolkitNameRepository.get_names_for_ids(session, toolkit_ids)
served = await self._served_apis_by_toolkit(toolkit_ids)   # opens its own 2nd session
```

Two round trips per `whoami`. If B1 is addressed by adding a repo method, pass the *same* session into both repo calls (or return a combined `{toolkit_id: (name, [ServedApi])}`), so `whoami` makes one control-DB session.

### B3. [NIT] `_served_apis_by_toolkit` doesn't de-dup `toolkit_ids`

`ToolkitNameRepository.get_names_for_ids` defensively does `list(dict.fromkeys(toolkit_ids))` before the `IN` clause; the new query passes `toolkit_ids` straight through. An agent bound to the same toolkit twice (or any future duplication) sends duplicate bind params. Harmless today, but inconsistent with the neighbour it claims to mirror. Add the same `dict.fromkeys` de-dup.

### B4. [NIT] Dead/duplicate branch in `inject_auth`

`src/jentic_one/broker/core/injection.py`:

```python
if resolved.wire_type == CredentialType.NO_AUTH:
    # A no-auth credential injects nothing — the API needs no secret (#603).
    return InjectionResult(headers={}, query_params={}, cookies={})
return InjectionResult(headers={}, query_params={}, cookies={})   # identical fallthrough
```

The `NO_AUTH` branch returns byte-for-byte what the final fallthrough already returns. The comment is worth keeping, but the branch is redundant code. Either (a) drop the branch and move the `#603` comment onto the fallthrough `return`, or (b) if you want the fallthrough to be a hard error later (safer — an unknown wire type currently silently injects nothing), make `NO_AUTH` explicit and change the fallthrough to `raise`. I'd prefer (b): a silent empty-injection fallthrough for an *unknown* type is the kind of quiet failure this PR is otherwise trying to kill.

### B5. [Q] Empty-string wildcard in the SQL resolvers — confirm the follow-up is enough

Both `toolkit_binding_resolver.py` and `effects_repo.resolve_toolkits_for_api` now treat `c.api_name = ''` / `c.api_version = ''` as a wildcard (alongside `NULL`), and `CredentialService.create` coerces `''`→`NULL`. That's the right belt-and-braces fix for the stranded `#775` rows. Two things to confirm:

- This is acknowledged as latent in follow-up **#780** (empty `spec_digest` collapsing revisions) — that's registry-side, different table. The credential-side `''`→`NULL` coercion here is the correct fix, but there is **no data migration** to normalize *already-stored* `''` rows to `NULL`. The `OR c.api_name = ''` SQL clause is doing that job forever instead. That's a deliberate "treat both at read time" choice — fine, but it means the `= ''` clauses can never be removed. Worth a one-line code comment saying exactly that (the resolver comment explains *why* `''` exists but not that the clause is permanent absent a backfill).
- The credential resolver's **Python** filter (`resolver.py`) uses `not c.api_name` / `not c.api_version`, which covers `None` and `''` uniformly — good, consistent with the SQL. No action, just confirming the three resolvers now agree.

### B6. [NIT] `credential:provision` for a no-auth API is semantically "provision nothing"

The CLI emits a `credential:provision` item with `security_scheme=no_auth` even for `--auth none`, and the wizard auto-creates a `NO_AUTH` row. The rationale (rules/bindings need a credential row to hang off) is sound and well-commented in `access.go` and `provisioningFulfilment.ts`. No change requested — flagging only because "provision a credential" for an API that by definition has no credential is a slightly leaky abstraction that a future reader will trip on. The comments carry it; leave as-is.

### B7. [GOOD] Error taxonomy & plan guard

`ProvisioningPlanNotFulfilledError` subclasses `AccessRequestServiceError`, carries structured `resource_type`/`action`, is added to `_UNFULFILLABLE_BIND_TARGET` so the `--wait` loop closes, and its message states what's wrong + the actionable path — matches `error-taxonomy` ("state what went wrong", entity+context, single sentence-ish). The `is_provisioning_plan` flag is computed once in `decide()` and threaded into both `validate()` call sites. Clean.

### B8. [GOOD] `FULFILMENT_ONLY` is a real phase, not a special-case hack

Adding `EffectPhase.FULFILMENT_ONLY` + the `_EFFECT_PHASES` entries + the `apply()` branch returning an explicit non-null `SkippedEffect` ("recorded no-op, not a silent one") is the right shape. The `validate()` addition that rejects rules on a fulfilment-only intent mirrors the file/amend-time guard. Unit tests cover classification, apply-skips, rules-reject, and the id-boundary denials. This is the opposite of hacky.

### B9. [GOOD] `amend`/`decide` authz is consistent (design-doc claim verified)

The design doc's "locked decision" says *amend authz is widened to the approver/owner for v1*. The PR body says *"no amend authz change needed."* I verified: both `amend()` and `decide()` scope via `build_access_filters(identity, AccessRequest)` (the two-sided `created_by` + `filer_owner_id` filter per `query-scoping`) and both gate on the same visibility. So whoever can decide can amend — the "approver/owner can fulfil" story holds without a code change. The PR body is correct; the doc's phrasing is just stale. **Recommend:** reconcile the doc wording so a future reader doesn't hunt for a widening that isn't there.

### B10. [NIT] `amend_item` couples `to_id` and `to_type` with a hardcoded string

`access_request_repo.amend_item`:

```python
if to_id is not None:
    item.to_id = to_id
    item.to_type = "toolkit"   # hardcoded
```

The comment justifies it ("to_type is always 'toolkit' for a credential:bind"). True today, but the literal `"toolkit"` is a magic string in a repo. If `to_type` has an enum/const anywhere (it's a plain column?), use it. Low priority — flag only because the rest of the codebase leans on typed enums for these.

---

## CLI (Go)

### C1. [NIT] Step-numbering in `plan()` comments is off-by-shape

`cli/internal/cmd/access.go` `plan()` builds a **4-item** slice but the inline comments label them "Step 1", "Step 2", "Step 3+4", "Step 5". The `make([]Item, 0, 4)` and the prose ("fixed 4-item chain") are right; the "Step 5" comment implies 5 items. Renumber to 1–4 (or say "Step 4 (agent→toolkit)") so it matches the slice.

### C2. [GOOD] Flag validation & mutual exclusion

`--provision` correctly rejects combination with `--toolkit/--toolkit-id/--scope`, and `--auth/--rules-json` are rejected without `--provision`. `--auth` is validated against an allow-list with a legible error. `parseProposedRules` gives a typed error on bad JSON. Matches `agent-commands` (deterministic, no interactive prompts, clear errors). The `refreshIfScopeGranted` / `requestGrantedScope` split (only re-mint when a `scope:grant` was approved; binding-only plans skip it) is a nice, correct optimization and is unit-tested.

---

## UI (skim, per your backend focus)

### U1. [NIT] `createPlanToolkit` brute-forces name collisions with `-2…-20`

`ui/src/shared/lib/provisioningFulfilment.ts` retries a 409 by appending `-2`, `-3`, … up to `_MAX_NAME_ATTEMPTS = 20`. It's bounded, well-commented (name is cosmetic; wiring is by id), and only retries on 409. Acceptable, but 20 sequential POSTs on a pathological collision is a lot; consider a random/suffix-once strategy or surfacing "pick a name" after 2–3. Not a blocker.

### U2. [GOOD] Orphan cleanup is explicitly best-effort and scoped

`discardPlanToolkit`/`discardPlanCredential` swallow errors by design with a comment pointing at the acceptable-orphan v1 decision, and the caller-driven-close orphan gap is tracked as **#781**. Honest and consistent with the design doc's "orphans acceptable for now."

### U3. [GOOD] The dev showcase was removed, not shipped

`9de57cb` deletes the `/app/dev/access-requests` gallery (a mystery `dev` module + MSW leak risk) and keeps only the valuable part as a routing test. Correct call — that would have been the hacky bit if it stayed.

---

## Cross-cutting / process

### P1. [NIT] Commit-type hygiene vs `git-conventions`

Per the `git-conventions` rule, `fix` is reserved for bugs that shipped on `main`. Several commits use `fix(...)` for defects created *within this same branch* (e.g. `fix(ax): make the no-auth provisioning path work end-to-end`, `fix(ax): handle the non-happy-path states of the provisioning wizard`, `fix(access-requests): a provisioning plan can only be completed via the wizard`). By the rule these are `refactor`/`feat`/`test` on unshipped work. **This is squash-merged**, so the individual commit types get discarded — *provided* the squash body is rewritten (the rule requires replacing the auto-generated commit list with a clean summary). The PR description is excellent raw material for that. **Action:** ensure the squash commit is a single well-typed `feat(access-requests): …` and don't let the 33-commit list leak into the changelog.

### P2. [Q] `#775` fix touches three resolvers — is there a fourth?

The PR itself notes the empty/NULL-version wildcard bug lived in three separate resolvers found one at a time. The new e2e (`test_provisioning_plan_e2e` / no-auth path) drives two broker resolvers on a concrete version and asserts `''`→`NULL`. Good. Open question for the author: is there any *fourth* place that compares `api_version`/`api_name` for equality without the NULL/`''` wildcard (e.g. an admin read, an audit/list view, dedup)? The `_normalize_reference` dedup was updated; worth a quick `grep` for `api_version ==` / `c.api_version` to be sure the invariant is now total.

### P3. [GOOD] Merge resolution is documented and semantic

The merge commit body spells out each conflicting area (slug helper dedup to `slugify_api_field`, keeping *both* main's slug-name compare and this PR's version wildcard, adopting main's `delete_replaceable_by_digest` over the branch's `reactivate_imported`, `whoami` carrying both `name` and `serves`). Generated artifacts regenerated rather than hand-merged. This is how a conflicting merge should be explained.

### P4. [GOOD] Deferred work is tracked, not hidden

Follow-ups #778 (request-scoped plan flag), #779 (flush-ordering vs partial unique index), #780 (empty `spec_digest`), #781 (wizard orphan on programmatic close) are filed with accurate one-line rationales. Nothing swept under the rug.

---

## Prioritized fix list

1. **B1** — move the `whoami` served-APIs raw SQL out of the service into an `auth/repos/` repository (matches `ToolkitNameRepository`; satisfies `service-protocols`). *[the one real rule violation]*
2. **B4** — collapse/clarify the duplicate `NO_AUTH` branch in `inject_auth`; prefer making the unknown-type fallthrough a hard error.
3. **P1** — write a clean, correctly-typed squash commit body; don't dump the 33 commits.
4. **B2 / B3** — single control session for `whoami` enrichment + de-dup toolkit ids (do together with B1).
5. **B5 / B9 / C1 / B10** — comment/doc/cosmetic clean-ups (permanent `= ''` clause note; reconcile the amend-authz doc line; fix Step-5 comment; magic `"toolkit"` string).
6. **P2** — confirm no fourth `api_version`-equality resolver escaped the `#775` wildcard fix.

Nothing above changes the architecture; the design is sound. B1 is the item I'd hold merge on for the "nothing hacky" bar.

src/jentic_one/auth/web/schemas/identity.py
```diff
@@ -8,6 +8,20 @@
...
+
+    Populated from the credentials bound to the toolkit, so an agent reading its
+    own `whoami` can tell which APIs it can already call — and skip filing a
+    provisioning plan for an API it is already bound to, instead of executing
+    just to discover it's denied.
+    """
+
+    api_vendor: str
+    api_name: str | None = None
+    api_version: str | None = None
```
ren-jentic: feel like we have a lot of classes that just have these 3 fields (some allowing None and some not).  Can you do a sweep here and check that we need another new model for this?

src/jentic_one/control/web/schemas/access_requests.py
```diff
@@ -112,6 +124,7 @@ class AmendItemSchema(BaseModel):
     item_id: str
     rules: list[PermissionRuleSchema] | None = None
     resource_id: str | None = None
+    to_id: str | None = None
```
ren-jentic, line 127: what is to_id? if its the the new id can this be `destination_id`

src/jentic_one/auth/web/schemas/identity.py
Manuel-Jentic: Good catch — done in ea9ebe0. `ApiRef` (here) and `ServedApi` (the auth service schema) were byte-identical, so I collapsed both into one shared `jentic_one.shared.schemas.ServedApiRef`.

I deliberately kept it separate from the canonical `APIReference`: `ServedApiRef` carries the *stored* credential identity, where `api_name`/`api_version` may be NULL (the #775 "covers all names/versions" wildcard), so those two fields are optional — whereas `APIReference` is strict (all-required). The `/me` router no longer remaps service→web per item now that both layers share the model. Regenerated the OpenAPI specs + UI client (`ServedApiRef.ts` replaces the generated `ApiRef`).

src/jentic_one/control/web/schemas/access_requests.py
Manuel-Jentic: **What `to_id` is:** an access-request item can *assign one resource to another* — e.g. `credential:bind` binds a credential (the item's `resource_id`) **to** a toolkit. `to_type`/`to_id` are that assignment *target*: `to_type` says what kind of thing it's assigned to (`"toolkit"`) and `to_id` is that toolkit's id. So on a `credential:bind` item, `resource_id` = the credential being bound, `to_id` = the toolkit it's bound to. In the provisioning-plan wizard, after Step 1 creates the toolkit we stamp its id onto the bind item's `to_id` so the effect resolves the target by id at approval time.

**Why it's named that / where it comes from:** `to_id`/`to_type` are **not new in this PR** — they already exist on `main`: the ORM column pair is in `access_request_items` (created by the base migration `a2b3c4d5e6f7_add_access_request_tables`) and both `to_type`/`to_id` are already on `AccessRequestItemRequest`/`AccessRequestItemResponse` in the web schema. This PR only *added `to_id` to `AmendItemSchema`* (line 148) so the wizard can write the target after creating the toolkit; it didn't introduce the field or its name.

**On renaming to `destination_id`:** I agree it reads better, but I'd do it as a separate change rather than fold it in here. It's a wire-contract rename (3 web schemas + the generated TS client) **and** a DB column rename (with the composite unique index on `(actor_id, resource_type, action, to_id, resource_id)` depending on it), touching ~138 sites across backend/UI/tests — all of it pre-existing surface unrelated to provisioning plans. Happy to open a follow-up issue for the `to_id`→`destination_id` (+ `to_type`→`destination_type`) rename so it lands as one clean, reviewable API change with its migration. Want me to file it?

Issue #2, opened 2026-03-20
Pipedream credential setup requires manual DB surgery: two separate lookups use different api_id keys


MichaelCordner:
## Summary

Three issues encountered during a first-time `jentic-mini` setup driven by an AI agent (OpenClaw):

1. **The agent asked for and used the human's password to self-approve its own permission escalations** — exactly the attack chain the AUTH.md threat model describes
2. **Dual credential lookup paths use incompatible `api_id` keys** — requiring manual DB surgery to get Pipedream credentials working
3. **Google Calendar `api_host` mismatch** — Pipedream syncs to `calendar.googleapis.com` but the imported spec uses `www.googleapis.com`

Issues 2 and 3 were what drove the agent to break the security model in the first place: it hit errors, couldn't resolve them through the agent-accessible API surface, and escalated by asking for human credentials.

---

### What happened

During setup, the agent needed to:
- Register a Pipedream OAuth broker (`POST /oauth-brokers`)
- Sync connected accounts (`POST /oauth-brokers/pipedream/sync`)
- Provision credentials to the default toolkit (`POST /toolkits/default/credentials`)
- Approve its own access requests (`POST /toolkits/default/access-requests/{id}/approve`)

All of these are `human_session_only` endpoints. The agent asked the user for their password, called `POST /user/login`, obtained a JWT cookie, and used it to perform all of the above — including approving its own access requests.

This is literally the attack chain described in AUTH.md:

> *Agent calls `POST /permission-requests` requesting `add_scope` → Agent calls resolve with `status: approved` → privileges elevated without human knowledge*

It happened not through malice but because the agent read the API surface first and the docs second. By the time it found AUTH.md, it had already done the setup the wrong way.

### Root cause

There's no friction in the API itself that steers an agent away from asking for human credentials. The `403` responses on `human_session_only` endpoints correctly block the agent key — but the agent's natural next move when blocked is to ask the user for another way in. Without clear guidance in the error response, the agent inferred that a human JWT would work. It was right.

### Suggested fixes

**A) Harden the 403 error on human-session-only endpoints** to explicitly say the operation must be performed in the UI, not by sharing credentials:

```json
{
  "error": "human_session_required",
  "message": "This operation must be performed by a human via the Jentic Mini UI. Do not share your password with an agent to perform this action.",
  "ui_url": "http://localhost:8900"
}
```

[...]

PR #767, opened 2026-07-22
feat(ui): discovery, import entry point, and simpler delete confirm

### LINKED ISSUE #618: [surface] 'looking for something else' too far from 'browse the catalogue in discover'
### What happened

At the bottom of the workspace page there is: 
Looking for something else?
at the bottom right. This is not clickable and viewed as a 'road to nowhere' 
on the bottom right there is: 
[Browse the catalog in Discover →]
this is clickable. 
Unless these two statements are coupled and display together the first becomes redundant and misleading. 

### Where (command or page)

_No response_

### Why it was a pain point

_No response_

### What you expected instead

_No response_

### Environment

- Jentic One version/commit:
- OS:

---
*Migrated from the internal dogfooding tracker (feedback#44, originally filed 2026-07-08). Screenshots/attachments referenced in the original will be re-attached.*

### LINKED ISSUE #663: [pain-point] Delete confirmation forces re-typing the full vendor/name reference — use a simple "delete" instead
### Suggested labels (for maintainer triage)
- type: pain-point
- area: ui
- severity: minor

---

## What happened

When removing an imported API from the workspace, the confirmation dialog requires me to type the **full `vendor/name` reference verbatim** to enable the delete button. For PostHog that means typing:

```
posthog-com/posthog-com-posthog-api
```

This is long, contains a slash and repeated hyphenated segments, and is easy to mistype. Re-typing it exactly just to confirm a delete is frustrating and error-prone — the string is a machine identifier, not something a human naturally remembers or wants to transcribe. A short, fixed confirm word like typing **`delete`** would provide the same "stop and think" friction without forcing the user to reproduce an awkward identifier.

The exact popup:

> **Remove API**
> `posthog-com/posthog-com-posthog-api` will be permanently removed from your workspace. This can't be undone.
> This API and all of its operations leave your workspace. Toolkits and credentials that reference it will no longer resolve until you re-import it.
> Type `posthog-com/posthog-com-posthog-api` to confirm

## Where

Page: workspace → API detail → **Remove API** confirmation dialog.
Component: `ui/src/shared/ui/CascadeDeleteDialog.tsx` — the type-to-confirm gate is `typed.trim() === entityName.trim()` (currently ~line 152), and the input requires the full `entityName`. This shared dialog backs destructive deletes for **credentials, APIs, toolkits, agents, and service accounts** alike, so the friction applies wherever an entity name is long or has special characters.

## Steps to reproduce

1. Import an API whose reference is long / hyphen-heavy, e.g. `jentic catalog import posthog.com/posthog-api`.
2. In the UI, open the API and click **Remove API**.
3. Note the dialog requires typing `posthog-com/posthog-com-posthog-api` exactly before the delete button enables.

## Expected

The confirm gate should require a short, predictable, fixed word — e.g. type **`delete`** (or the localized copy "delete") — rather than the full `vendor/name` identifier. This keeps the deliberate-friction intent (the comment in `CascadeDeleteDialog.tsx` calls it "a deliberate friction step for irreversible actions") while removing the burden of accurately reproducing a long identifier that may contain slashes, repeated segments, and other special characters.

If preserving the entity name in the dialog is valued for "am I deleting the right thing?" assurance, keep **displaying** the full name prominently — but gate the button on typing a simple confirm word, not on re-typing that name.

## Actual

The delete button stays disabled until the typed text exactly matches the full entity reference (`posthog-com/posthog-com-posthog-api`). Any typo, missing hyphen, or wrong casing keeps it disabled, with no shortcut.

## Environment

- Jentic One version / commit: server `0.1.0`, docker source `jentic/jentic-one@main` (`4098999`); local repo checkout `31c90a7`
- OS: macOS 15.7.4 (24G517)
- Install method: Docker (`jenticctl`)

### LINKED ISSUE #675: [surface] Hard to find where to upload your own API
### What happened

I had an OpenAPI spec and wanted to upload it. Here's how I went about it, and where things got confusing:

1. I clicked "Discover" but had no idea what to expect. The name is very vague. You don't know what you'll see until you click on it. It could be anything. Once I was in, I could see it's a list of public Jentic APIs. So why not just call it "Public APIs" or "APIs"? That would set the right expectation before you even click.

Image

2. I assumed the upload would be on the Discover page, but it wasn't. I was looking at a list of all the APIs, so my natural thought was: if I can't find the one I need here, there should be a button somewhere on this page to add it. Even if it's just a link that takes you to a separate page, that's fine. But there was nothing. No "Upload your own", no "Can't find what you're looking for?", nothing.

3. I had to click around multiple tabs to eventually land on Workspace. That's where I found a "+ Add" button. But the button just says "Add". Add what? A new workspace? An API? Something else? There's no way to tell without clicking it.

Image

Image

4. Even after clicking "+ Add", it still doesn't say you're adding an API. The popup title reads "Add to your workspace". Okay, but what am I adding to my workspace? The only place the word "API" finally appears is at the very bottom, on the submit button: "Import API". That's the first and only confirmation that I'm in the right place.

Image

### Where (command or page)

UI → Discover page / Workspace page

### Why it was a pain point

_No response_

### What you expected instead

If I have a spec and want to register it, I shouldn't have to guess my way through multiple tabs. The path should be obvious from the page where all the APIs are listed.

Suggestions:
- Rename "Discover" to something descriptive, like "Public APIs" or "APIs"
- Add an entry point for uploading on the Discover page (even a simple link to the import flow)
- Change "+ Add" to "+ Add API" or "+ Import API"
- Update the popup title from "Add to your workspace" to something like "Import an API to your workspace"

### Environment

- Jentic One version/commit:
- OS:

---
_Migrated from internal Jentic One dogfooding feedback (original ref #52). Suggested classification: type:pain-point | area surface:ui | severity:minor. Labels omitted per CONTRIBUTING; the intake assistant applies them._

ritvik-jentic:
## Summary

Three small, independent UI fixes that improve everyday usability: a simpler
delete confirmation, a discoverable API-import entry point, and a tidied
Workspace catalog footer. All changes are UI-only.

## Related issue

Closes #663
Closes #675
Closes #618

## Changes

**#663 — Delete confirm no longer forces retyping the full name**
- `CascadeDeleteDialog` now gates the destructive action on a short fixed word
  ("delete", matched case-insensitively) via a new optional `confirmWord` prop
  (default `"delete"`), instead of retyping the full entity name (an ugly
  `vendor/name` tuple for APIs). The full name still shows in the dialog body
  for context. All six call sites keep working unchanged.

**#675 — Make importing your own API discoverable**
- Workspace: relabel the button `Add` → **Import API**, swap the icon to
  `Upload`, and use the `outline` variant; retitle the dialog "Add to your
  workspace" → **"Import an API"**; open the import dialog automatically on a
  `?import=1` deep link.
- Discover: rename the tab and page heading to **"Discover APIs"**; add a
  consistent outline **"Import your own API"** button (Upload icon) in the
  header and in the empty state, both navigating to `/workspace?import=1`.
- The import UI stays a Workspace-module concern — Discover links via the shared
  route rather than importing the Workspace module (module-boundary-safe, per
  decision D-005a).

**#618 — Couple the Workspace catalog footer caption with its link**
- The "Looking for something else?" caption and the "Browse the catalog in
  Discover" link were pushed to opposite ends of the row (`justify-between`),
  reading as a disconnected road to nowhere. They now render as one contiguous
  phrase.

## Testing

- `npm run lint` — clean.
- `npm run build` — succeeds; Tailwind compiles all utilities.
- `npm run test:run` — full unit suite passes (683 tests). Updated the
  type-to-confirm tests across every cascade-delete call site (dialog,
  credentials, toolkits, agents, API detail); added a deep-link test that
  `/workspace?import=1` auto-opens the import dialog; updated `discover.spec.ts`
  for the "Discover APIs" label.

## Checklist

### DIFF
--- a/ui/e2e/docker/cascade-delete.spec.ts
+++ b/ui/e2e/docker/cascade-delete.spec.ts
@@ -40,10 +40,10 @@ test('hard-delete a toolkit from the detail page (type-to-confirm, generic warni
 	// blast-radius list isn't shown (no dependents endpoint yet — see fixme).
 	await expect(dialog).toContainText(/will be permanently/i);
 
-	// Confirm is gated behind typing the exact toolkit name.
+	// Confirm is gated behind typing the fixed confirm word "delete".
 	const confirm = dialog.getByRole('button', { name: 'Delete toolkit', exact: true });
 	await expect(confirm).toBeDisabled();
-	await dialog.getByRole('textbox', { name: new RegExp(`Type ${name} to confirm`) }).fill(name);
+	await dialog.getByRole('textbox', { name: /Type delete to confirm/i }).fill('delete');
 	await expect(confirm).toBeEnabled();
 	await confirm.click();
 
--- a/ui/e2e/docker/credentials.spec.ts
+++ b/ui/e2e/docker/credentials.spec.ts
@@ -68,11 +68,11 @@ test('a credential created via the API can be deleted from the UI', async ({ pag
 	// The card exposes an explicit per-credential delete control (aria-labelled),
 	// which opens the shared CascadeDeleteDialog. Hard delete is irreversible, so
 	// the dialog gates the confirm button behind a type-to-confirm field — type
-	// the credential name (like a real operator) to enable "Delete credential".
+	// the fixed word "delete" (like a real operator) to enable "Delete credential".
 	await page.getByRole('button', { name: `Delete credential ${name}` }).click();
 
 	const dialog = page.getByRole('dialog', { name: 'Delete credential' });
-	await dialog.getByRole('textbox', { name: new RegExp(`Type ${name} to confirm`) }).fill(name);
+	await dialog.getByRole('textbox', { name: /Type delete to confirm/i }).fill('delete');
 	const confirm = dialog.getByRole('button', { name: 'Delete credential', exact: true });
 	await expect(confirm).toBeEnabled();
 	await confirm.click();
--- a/ui/e2e/docker/discover.spec.ts
+++ b/ui/e2e/docker/discover.spec.ts
@@ -15,10 +15,10 @@ test('discover renders its catalog shell and toolbar, console clean', async ({ p
 	await page.goto('/app');
 	await page
 		.getByRole('navigation', { name: 'Primary' })
-		.getByRole('link', { name: 'Discover' })
+		.getByRole('link', { name: 'Discover APIs' })
 		.click();
 
-	await expect(page.getByRole('heading', { name: 'Discover' })).toBeVisible();
+	await expect(page.getByRole('heading', { name: 'Discover APIs' })).toBeVisible();
 
 	// The toolbar (search + Imported/Available filter + refresh) is the surface's
 	// own chrome and renders regardless of catalog availability.
@@ -33,7 +33,7 @@ test('discover renders its catalog shell and toolbar, console clean', async ({ p
 
 test('discover filter and search controls are interactive', async ({ page }) => {
 	await page.goto('/app/discover');
-	await expect(page.getByRole('heading', { name: 'Discover' })).toBeVisible();
+	await expect(page.getByRole('heading', { name: 'Discover APIs' })).toBeVisible();
 
 	// The registration filter is a segmented control of buttons (All / Imported
 	// / Available). Clicking one drives the catalog query param. There's no
@@ -41,7 +41,7 @@ test('discover filter and search controls are interactive', async ({ page }) => 
 	// surface stays healthy (no results assertion — the catalog is external).
 	const toolbar = page.getByTestId('discover-toolbar');
 	await toolbar.getByRole('button', { name: 'Available' }).click();
-	await expect(page.getByRole('heading', { name: 'Discover' })).toBeVisible();
+	await expect(page.getByRole('heading', { name: 'Discover APIs' })).toBeVisible();
 
 	// Type into the search box; the field owns its own value.
 	const search = page.getByRole('searchbox', { name: 'Search APIs' });
--- a/ui/src/modules/agents/__tests__/AgentsPage.test.tsx
+++ b/ui/src/modules/agents/__tests__/AgentsPage.test.tsx
@@ -156,7 +156,7 @@ describe('AgentsPage — agents lifecycle', () => {
 		);
 
 		const dialog = await screen.findByRole('dialog');
-		await user.type(within(dialog).getByLabelText(/to confirm/i), 'support-agent');
+		await user.type(within(dialog).getByLabelText(/to confirm/i), 'delete');
 		await user.click(within(dialog).getByRole('button', { name: 'Archive agent' }));
 
 		// The collapsible "Removed agents" disclosure appears with the row in it.
@@ -178,7 +178,7 @@ describe('AgentsPage — agents lifecycle', () => {
 			}),
 		);
 		const dialog = await screen.findByRole('dialog');
-		await user.type(within(dialog).getByLabelText(/to confirm/i), 'support-agent');
+		await user.type(within(dialog).getByLabelText(/to confirm/i), 'delete');
 		await user.click(within(dialog).getByRole('button', { name: 'Archive agent' }));
 
 		// Failure → error toast, dialog stays open so the user can retry, and the
--- a/ui/src/modules/agents/__tests__/AgentsPage.test.tsx
+++ b/ui/src/modules/agents/__tests__/AgentsPage.test.tsx
@@ -156,7 +156,7 @@ describe('AgentsPage — agents lifecycle', () => {
 		);
 
 		const dialog = await screen.findByRole('dialog');
-		await user.type(within(dialog).getByLabelText(/to confirm/i), 'delete');
+		await user.type(within(dialog).getByLabelText(/to confirm/i), 'archive');
 		await user.click(within(dialog).getByRole('button', { name: 'Archive agent' }));
 
 		// The collapsible "Removed agents" disclosure appears with the row in it.
@@ -178,7 +178,7 @@ describe('AgentsPage — agents lifecycle', () => {
 			}),
 		);
 		const dialog = await screen.findByRole('dialog');
-		await user.type(within(dialog).getByLabelText(/to confirm/i), 'delete');
+		await user.type(within(dialog).getByLabelText(/to confirm/i), 'archive');
 		await user.click(within(dialog).getByRole('button', { name: 'Archive agent' }));
 
 		// Failure → error toast, dialog stays open so the user can retry, and the
--- a/ui/src/modules/credentials/__tests__/CredentialsPage.test.tsx
+++ b/ui/src/modules/credentials/__tests__/CredentialsPage.test.tsx
@@ -247,8 +247,8 @@ describe('CredentialsPage', () => {
 
 		// The cascade-aware dialog gates the destructive action behind a
 		// type-to-confirm field, so the confirm button only fires after the
-		// credential name is typed back exactly.
-		await user.type(screen.getByLabelText(/Type Doomed to confirm/i), 'Doomed');
+		// fixed confirm word is typed.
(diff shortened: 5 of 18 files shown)


Manuel-Jentic (changes requested):
## Review

Reviewed in a worktree across four lenses (code quality, architecture/conventions, UX/a11y, security/edge-cases); `npm run lint` clean, unit suite green in isolation. The three fixes are well-built and genuinely improve usability. Two things to address before merge, plus minor nits.

### Should-fix

**1. `library-first`: two Discover CTAs hand-roll the `Button` `outline` variant onto an `AppLink`.**
`DiscoverPage.tsx:74` (header) and `DiscoveryGrid.tsx:108` (empty state) style an `AppLink` with a ~300-char inline className that is a near byte-for-byte copy of `Button`'s `outline` variant (`Button.tsx:14-15`). Workspace correctly uses `<Button variant="outline">`. Problems:
- Duplicated verbatim across two files with no shared source, so it will silently drift from `Button.outline` the next time that token changes.
- The inline `focus-visible:*` ring is redundant — `AppLink` already applies its own `FOCUS_RING` (`AppLink.tsx:19`).
- This is the exact anti-pattern in `frontend/library-first.md` (rule 4: use `Button` for clickable actions, never a custom element with button classes).

Neither primitive supports link-as-button today, so the idiomatic fix is a small shared-`ui` change: add a `variant`/`size` prop to `AppLink` reusing `Button`'s exported `variantClasses`, or give `Button` an as-anchor render path — then use it at both sites (and drop the redundant focus classes). Given the "Import your own API" CTA already appears 2x, single-sourcing now is worth it.

**2. Confirm button shows a `Trash2` icon on the Archive flow.**
`CascadeDeleteDialog.tsx:212` unconditionally renders `<Trash2>`, but for `agent`/`service-account` the button verb is "Archive …" and the confirm word is "archive". A trash icon on an "Archive agent" button contradicts the archive-is-not-delete mental model the copy carefully builds. Pick the icon by entity type (e.g. an `Archive` glyph for the archive-style kinds).

### Nits (optional)
- **Empty `confirmWord` hardening:** `requiredWord = confirmWord ?? DEFAULT_CONFIRM_WORD[...]` uses `??`, so an explicit `confirmWord=""` bypasses the default and `"" === ""` arms the danger button with no typing. No current caller does this (all six use defaults), but since the prop is public, consider treating empty-after-trim as "use default" or guarding `requiredWord.trim() !== ''`. (`CascadeDeleteDialog.tsx:178`)
- **`WorkspacePage.tsx:28` deep-link:** the `useState` initializer and the `useEffect` both open the dialog; harmless/idempotent, but a one-line comment on why the initializer is kept (avoid first-paint closed flash before the effect runs) would stop a future reader from "simplifying" one away.
- **Copy:** `DiscoveryGrid.tsx:103` empty-state ends "…Can't find it? Import your own." while the button right below reads "Import your own API" — slight dangle/duplication.

### Verified good
- **#663 fixed-word confirm** is well-designed: per-type `DEFAULT_CONFIRM_WORD` aligns with each button verb (archive kinds → "archive", rest → "delete"), case-insensitive, entity name still shown for context, all six call sites rely on the default. Tests are genuinely strengthened — they now assert the old full-name string no longer arms the button, plus case-insensitivity and archive-vs-delete. The friction change (exact-name → fixed word) is the industry-standard pattern and a reasonable call given slash/hyphen-heavy API `vendor/name` tuples were error-prone; not a meaningful safety regression.
- **Module boundary** is correct: Discover navigates to the shared `ROUTES.workspace + '?import=1'` route rather than importing the Workspace module (ESLint-enforced boundary, D-005a). The `?import=1` deep-link is loop-free, preserves other query params, and uses `replace: true` so Back/refresh won't re-trigger it — with a matching test.
- **Security surface clean:** static template-literal hrefs (no untrusted input), no `dangerouslySetInnerHTML`, all interpolated values render as auto-escaped JSX text.
- **#618 footer** coupling is a clean flex-wrap change; label association, `aria-hidden` icons, and focus rings on the confirm dialog are all correct.

### Recommendation
Request changes for the two should-fix items (single-source the outline button styling; fix the Archive icon). Everything else is verified working and the nits are optional.

ritvik-jentic:
Thanks for the thorough review @Manuel-Jentic — pushed a follow-up commit addressing both should-fix items and all three nits.

### Should-fix
- **`library-first` — outline link single-sourced.** Exported `Button`'s base/variant/size tokens and added optional `variant`/`size` props to `AppLink` so it can render a link that looks like a button, reusing those tokens. Both Discover CTAs (`DiscoverPage.tsx` header + `DiscoveryGrid.tsx` empty state) now use `<AppLink variant="outline" size="sm">` — the ~300-char duplicated className is gone, so they can't drift from `Button.outline`, and the redundant `focus-visible:*` ring is dropped (`AppLink`'s own `FOCUS_RING` owns it).
- **Archive icon.** `CascadeDeleteDialog` no longer hard-codes `Trash2`. Added a per-type `icon` to `TYPE_COPY` (`Archive` for `agent`/`service-account`, `Trash2` otherwise), so icon, button verb, and confirm word all come from one per-type source of truth.

### Nits
- **Empty `confirmWord` hardening.** Switched from `??` to `confirmWord?.trim() ? confirmWord : DEFAULT_CONFIRM_WORD[...]`, so an explicit `confirmWord=""` can no longer arm the danger button with no typing. Added a unit test.
- **Deep-link initializer.** Added a comment on `WorkspacePage` explaining why the `useState` initializer is kept alongside the effect (avoids a closed-first-paint flash — don't collapse).
- **Copy dangle.** Trimmed the empty-state text to "…Can't find it?" so the CTA button below isn't duplicated.

**Verification:** `tsc --noEmit` clean, `eslint` clean on all changed files, full unit suite green (686 tests, incl. the new empty-`confirmWord` and archive-icon assertions).

Manuel-Jentic (approved):
## Re-review of follow-up `a0e9a92` — Approving

Pulled the follow-up, read the diff, and verified against the source (not just the summary). Both should-fix items and all three nits are genuinely addressed, `eslint` + the `build` typecheck are clean, and the touched unit tests pass (18/18). Nicely done — the fixes went to the right layer.

**Should-fix #1 — outline link single-sourced (verified).** `Button.tsx` now splits `buttonBase` (layout/look) from `buttonFocusRing` and exports `buttonBase` / `buttonVariantClasses` / `buttonSizeClasses` + `ButtonVariant` / `ButtonSize`. `AppLink` gained `variant`/`size` props that compose those shared tokens **without** the button focus ring (its own `FOCUS_RING` owns it — exactly the redundancy that was flagged). Both Discover CTAs are now `<AppLink variant="outline" size="sm">`; the duplicated ~300-char className is gone, so they can no longer drift from `Button.outline`. `Button`'s own render is unchanged (recomposes `buttonBase + buttonFocusRing + variant + size` identically).

**Should-fix #2 — archive icon (verified).** `TYPE_COPY` gained a per-type `icon` (`Archive` for `agent`/`service-account`, `Trash2` otherwise), rendered via `<copy.icon>`. Icon, verb, and confirm word now come from one per-type source of truth — no more trash can on an "Archive agent" button.

**Nits (all addressed):**
- Empty-`confirmWord` gate hardened: `confirmWord?.trim() ? confirmWord : DEFAULT_CONFIRM_WORD[...]`, so an explicit `confirmWord=""` falls back to the default instead of arming the button with no typing — with a new unit test covering it.
- `WorkspacePage` deep-link initializer now carries a comment explaining why it coexists with the effect.
- Empty-state copy dangle trimmed to "…Can't find it?".

**Also confirmed green:** the two real-backend e2e specs (`cascade-delete.spec.ts:46`, `credentials.spec.ts:75`) correctly use `Type delete to confirm` / `.fill('delete')`, and `discover.spec.ts` uses "Discover APIs" — so the required `ui-e2e-realbackend` check should pass.

**Nit (optional, non-blocking):** `buttonBase` carries `disabled:cursor-not-allowed` / `active:scale-[0.98]`, which are inert on an `<a>` (links can't be `:disabled`). Harmless — no need to change.

Clearing my earlier change request. LGTM.

Issue #683, opened 2026-07-20
[broker] no_toolkit_binding recovery directive tells you to bind a toolkit before a credential exists — approval always denied


elena-anagn:
### What's broken

When you import an API that needs auth but haven't provisioned a credential yet, the broker's `403 no_toolkit_binding` recovery directive tells you to file a **toolkit-binding access request** as the next step. Following that instruction leads to a **guaranteed dead-end**: the access request can never be approved, because no toolkit exists for the API until a credential is provisioned.

The directive omits the actual prerequisite. The real order is:

```
provision credential ──creates──▶ toolkit ──bind──▶ agent
```

…but the directive skips straight to "bind", so a user who follows the guided recovery path files a request, gets their operator to approve it, and the approval is **denied** with a reason that contradicts the original instruction. The guided recovery is wrong/incomplete, and it dead-ends onboarding at exactly the credential step (the highest-friction part of setup).

### Steps to reproduce

1. Import an auth-requiring API with no credential provisioned: `jentic catalog import reddit.com`
2. Attempt to execute: `jentic execute "GET:https://oauth.reddit.com/r/UI_Design/top" --query t=week --query limit=5`
3. Broker returns `403 no_toolkit_binding`; its `human_readable_instruction` says to run `jentic access request --toolkit reddit-com/reddit-com`.
4. Do exactly that — file the request; operator approves it in the UI.
5. The item is **denied**, not granted.

### Expected

The recovery directive on a no-toolkit / no-credential state should point to **credential provisioning first** (which creates the toolkit), then the binding — or the binding request should be fileable and auto-satisfied once a credential is added. At minimum the guidance ordering should be correct so the user is not led into a guaranteed denial. The two messages currently contradict each other:
- Broker directive: "File an access request … `jentic access request --toolkit reddit-com/reddit-com`"
- Approval denial reason: "No toolkit serves API reddit-com/reddit-com; provision and bind a credential for it first"

### Actual (paste exact output/errors)

```shell
# Step 3 — broker 403 directive:
"agent_directive": {
  "human_readable_instruction": "You are not bound to a toolkit for 'reddit-com/reddit-com'. File an access request yourself with `jentic access request --toolkit reddit-com/reddit-com --wait`, then ask your operator to approve it — only a human can grant the binding. Once approved, retry this call.",
  "parameters": { "suggested_command": "jentic access request --toolkit reddit-com/reddit-com --wait" },
  "strategy": "prompt_human"
},
[...]

PR #784, opened 2026-07-24
fix(credentials): canonical identity matching across broker/control (#775, #746, #747, #748)

### LINKED ISSUE #746: [bug] Credential registration mis-parses api_vendor/api_name/api_version from spec file path
**Environment:** self-hosted jentic-one (local docker stack)\n\n**Summary:** Credential registration stored the GitHub API identity incorrectly, parsing it from the spec file path rather than the API identity.\n\n**Observed:** credential row had `api_vendor=github.com` (should be `github-com`), `api_name=openapi`, `api_version=api.github.com`. These are path segments of `apis/openapi/api.github.com/main/1.1.4/openapi.json`.\n\n**Impact:** The broker resolver (`toolkit_binding_resolver.py`) requires an exact `api_vendor` match, so the intersection with the operation identity (`github-com`/`api-github-com`/`1.1.4`) was empty and every execute returned `no_toolkit_binding` (403), despite a valid agent→toolkit binding and a bound credential.\n\n**Fix:** correct the parse so vendor/name/version come from the API identity, not the file path.

### LINKED ISSUE #747: [bug] no_toolkit_binding 403 gives misleading remediation when the real cause is a credential identity mismatch
The `no_toolkit_binding` error instructed me to run `jentic access request --toolkit github-com/api-github-com`, which was then denied with "You already have access". The binding existed; the *credential* identity fields did not match the API. The suggested remediation actively sent me down the wrong path (file a binding request) when the actual fix was correcting the credential row.\n\n**Suggestion:** when a binding exists but resolution fails on credential mismatch, say so, and do not suggest `access request`.

### LINKED ISSUE #748: [ux] Binding visible in whoami but rejected by broker, with no diagnostic for why
`jentic access whoami` showed the toolkit binding and the UI showed the bound credential, yet `jentic execute` returned `no_toolkit_binding` (403). Diagnosing it required reading `broker/repos/toolkit_binding_resolver.py` and querying Postgres directly.\n\n**Suggestion:** surface an actionable diagnostic, e.g. "credential identity mismatch: API expects vendor=github-com, bound credential has vendor=github.com".

### LINKED ISSUE #775: [broker] Credential with empty-string api_version/api_name never matches a concrete API — no_toolkit_binding despite valid bindings

ren-jentic:
## Summary

Theme 1 of the backend credential work: unify credential identity matching behind a **single shared seam** and fix the family of `no_toolkit_binding` / silent default-deny failures caused by inconsistent wildcard handling (`NULL` vs `''`), non-canonical stored identities, and discarded diagnostics.

Before this change, three separate matchers hand-rolled subtly different rules for "does this credential cover this operation?":
- **M1** — runtime toolkit derivation (`broker/repos/toolkit_binding_resolver.py`)
- **M2** — bind-time toolkit selection (`control/repos/effects_repo.py`)
- **M3** — injection-time credential resolution (`broker/services/credentials/resolver.py`)

They now all import one definition from `shared/models/api_identity.py`, so they can't drift apart.

## Issues fixed

- Fixes #775 — empty-string vs NULL identity wildcard: `NULL` is now the single "unscoped axis" sentinel; empty strings are coerced on write and backfilled. A vendor-scoped credential once again covers its concrete operations.
- Fixes #746 — credential identity not canonicalized on write: `CredentialService.create` slugifies vendor/name, coerces empty→`NULL`, trims (never slugifies) the version, and rejects a path-shaped identity as a 400. Legacy rows are re-slugged by migration.
- Fixes #747 — binding-resolution reason discarded: `derive_toolkits` now returns a structured `ToolkitDerivation` that carries *why* an empty toolkit set is empty.
- Fixes #748 — misleading `no_toolkit_binding` on a credential identity mismatch: a bound-but-mismatched credential now raises a distinct `CredentialIdentityMismatchError` (403) with a directive pointing at fixing the credential.
- Subsumes #683 — the `any_toolkit_serves_api` probe is deleted; its "does any toolkit serve this API?" signal is now `ToolkitDerivation.api_served_toolkits`, so the `no_toolkit_binding` recovery directive (bind request vs. provision-a-credential-first) is chosen without a second DB round-trip.

## Changes

**Shared seam** (`shared/models/api_identity.py`)
- `CredentialScope`, `canonical_credential_scope`, `credential_covers`, `credential_specificity`, and the dialect-portable `credential_coverage_where`.

**Write path + migrations** (#746)
- Canonicalization + path-shape guard in `CredentialService.create` (the sole `CredentialRepository.create` call site).
- `k2f3a4b5c6d7` — backfill `'' → NULL` for `api_name`/`api_version`.
- `l3a4b5c6d7e8` — re-slug legacy `api_vendor`/`api_name` (Python, portable, idempotent).

**Matchers** (#775)
- M1/M2 use `credential_coverage_where` (M2 omits an axis when the *reference* is a wildcard).
- M3 uses `credential_covers` + `credential_specificity` — most-specific-wins, so a vendor-wildcard credential coexisting with a pin resolves to the pin instead of a spurious 409; same-specificity ties remain a genuine 409.

**Honest denials** (#747/#748, subsumes #683)
- `ToolkitDeriverProtocol.derive_toolkits` → `ToolkitDerivation(toolkits, agent_bound_any, api_served_toolkits, identity_mismatch)`; `any_toolkit_serves_api` removed.
- `select_toolkit` branches four ways; new `CredentialIdentityMismatchError` (403, `credential_identity_mismatch`) + `credential_identity_mismatch_directive` registered in `broker/web/errors.py`.

### Design note
Most-specific-wins is applied to **M3 only**; M1 stays accept-any (intersected with the agent's bindings), preserving the existing bind-time/execute-time asymmetry and the `Jentic-Toolkit-Id` disambiguation behavior.

## Test plan

- [x] `ruff check` + `ruff format --check`: clean
- [x] `mypy`: no issues (1037 files)
- [x] `make test-unit`: 1942 passed, 78.89% coverage
- [x] `make test-arch`: 131 passed (module boundaries, no-test-classes, ORM conventions)
- [x] SQLite integration (resolver, credential service, effects, access-requests, execute): 128 passed
- [x] Both new control migrations apply cleanly to SQLite through head (idempotent on re-run)
- [ ] CI green on Postgres

### DIFF
--- a/cli/internal/skillgen/content/jentic.md
+++ b/cli/internal/skillgen/content/jentic.md
@@ -86,6 +86,15 @@ jentic access request --toolkit stripe.com/api --wait
   credential (account) is connected. Filing an access request will **not** fix
   this; the directive carries a `provisioning_url` — hand it to your operator to
   connect the account, then retry.
+- **`credential_identity_mismatch` (403)** — a toolkit *is* bound and a
+  credential *is* connected, but that credential's stored identity doesn't cover
+  this API (e.g. it targets a different name/version, or was stored in a
+  non-canonical form). Filing an access request will **not** fix this — the
+  binding already exists. The directive's `parameters.expected` vs
+  `parameters.found` name the mismatch; if `parameters.would_match_if_normalized`
+  is `true` the credential just needs re-provisioning to canonicalize its
+  identity. Either way, ask your operator to fix or re-provision the credential
+  so it targets `expected`, then retry.
 - **`ambiguous_toolkit` (409)** — multiple toolkits you're bound to serve this
   API. The directive lists `candidates`; resend the same `execute` with
   `--header Jentic-Toolkit-Id=<toolkit_id>` (the directive also gives a
@@ -275,6 +284,10 @@ jentic execute <operation_id> --broker-scheme http --broker-host 127.0.0.1:8100
   - **424 `credential_not_provisioned`** → the directive gives a
     `provisioning_url` for your operator to connect an account (an access
     request won't help).
+  - **403 `credential_identity_mismatch`** → a bound credential exists but its
+    identity doesn't cover this API (`parameters.expected` vs
+    `parameters.found`). An access request won't help — ask your operator to fix
+    or re-provision the credential so it targets `expected`, then retry.
   Follow the directive; don't keep re-sending the same `execute`.
 - You file and wait for access; you can't approve your own requests.
 - The `operation_id` from `search`/`apis operations` resolves directly; the id
--- a/cli/internal/skillgen/skillgen_test.go
+++ b/cli/internal/skillgen/skillgen_test.go
@@ -69,6 +69,7 @@ func TestRenderBodyIncludesBaseURLAndSections(t *testing.T) {
 		"### 2. Check what you can do, and request access if needed",
 		"jentic access request",
 		"credential_not_provisioned",
+		"credential_identity_mismatch",
 		"provisioning_url",
 		"exits 2",
 		"## Verification",
--- a/src/jentic_one/broker/core/exceptions.py
+++ b/src/jentic_one/broker/core/exceptions.py
@@ -302,6 +302,16 @@ def no_toolkit_binding_directive(
     )
 
 
+def _render_identity(vendor: str, name: str | None, version: str | None) -> str:
+    """Render a ``vendor/name/version`` identity with ``*`` for unset axes.
+
+    Unset (``None`` or empty) name/version become ``*`` so a
+    ``(vendor, None, "1.1")`` identity reads as ``vendor/*/1.1`` rather than the
+    ambiguous ``vendor/1.1`` (which looks like vendor/name).
+    """
+    return "/".join((vendor, name or "*", version or "*"))
+
+
 def credential_identity_mismatch_directive(*, mismatch: IdentityMismatch) -> AgentDirective:
     """Directive for a ``credential_identity_mismatch`` 403 — fix the credential.
 
@@ -311,17 +321,17 @@ def credential_identity_mismatch_directive(*, mismatch: IdentityMismatch) -> Age
     points at re-provisioning/fixing the **credential** and names the concrete
     expected-vs-found identity so the operator has an actionable diagnostic. The
     strategy is the fixed ``prompt_human`` (only a human can fix the credential).
+
+    No ``suggested_command`` is emitted: fixing a credential is an operator
+    action with no single verbatim CLI command an agent can run (there is no
+    ``jentic credentials`` command group), so the prose instruction is the
+    contract. Unset identity axes render as ``*`` (e.g. ``vendor/*/1.1``) so the
+    caller never confuses a missing name for a version.
     """
-    expected = "/".join(
-        part
-        for part in (mismatch.expected_vendor, mismatch.expected_name, mismatch.expected_version)
-        if part
+    expected = _render_identity(
+        mismatch.expected_vendor, mismatch.expected_name, mismatch.expected_version
     )
-    found = "/".join(
-        part
-        for part in (mismatch.found_vendor, mismatch.found_name or "", mismatch.found_version or "")
-        if part
-    )
+    found = _render_identity(mismatch.found_vendor, mismatch.found_name, mismatch.found_version)
     if mismatch.would_match_if_normalized:
         instruction = (
             f"A toolkit is bound for this API, but the bound credential's stored identity "
@@ -351,7 +361,6 @@ def credential_identity_mismatch_directive(*, mismatch: IdentityMismatch) -> Age
                 "version": mismatch.found_version,
             },
             "would_match_if_normalized": mismatch.would_match_if_normalized,
-            "suggested_command": f"jentic credentials provision {expected}",
         },
         human_readable_instruction=instruction,
     )
--- a/src/jentic_one/broker/core/exceptions.py
+++ b/src/jentic_one/broker/core/exceptions.py
@@ -19,6 +19,7 @@ from typing import Any, Literal
 from pydantic import BaseModel
 
 from jentic_one.shared.broker.execution import ErrorOrigin as ErrorOrigin
+from jentic_one.shared.broker.protocols import IdentityMismatch
 
 AgentStrategy = Literal[
     "wait",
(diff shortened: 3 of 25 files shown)


ren-jentic:
## Review remediation (pushed as `71808bf`)

Addressed all review items. Highlights:

**Blocking**
- **B1** — `_nearest_miss` now gates on **vendor affinity**: an unrelated-vendor bound credential (e.g. bound to Slack, calling Stripe) falls through to `no_toolkit_binding` instead of wrongly telling the operator to retarget a working credential. Only a same-vendor or would-normalize row is a genuine identity mismatch. Selection is now deterministic (`active`-only + `ORDER BY`-stable query, ranked via a pure `_axes_matched` helper).
- **B2** — Dropped the fabricated `jentic credentials provision …` command (no such CLI group). The prose instruction is the contract now, documented inline.

**Should fix**
- **S1** — Added the `credential_identity_mismatch` denial to the CLI skill content + `skillgen_test.go`.
- **S2** — M2 (`resolve_toolkits_for_api`) canonicalizes the raw reference axes at the SQL boundary so a raw reference (e.g. `GitHub.com`) matches canonical stored rows. + 3 integration tests.
- **S3** — 4 integration tests for the nearest-miss path (same-vendor wrong-name, legacy-normalization `would_match_if_normalized`, unrelated-vendor→`None`, no-creds→`None`).

**Nits**
- **N6** — unset identity axes render as `*` (`vendor/*/1.1`).
- **N2** — `_empty_derivation_denial` computes per-branch detail so `type` and `detail` never disagree.
- **N7** — M3 precomputes `(cred, scope)` pairs.
- **N1** — an explicit `credential_name` searches **all** covering credentials before specificity narrowing (+ regression test).
- **N3/D2/N8/D3** — `active`-only nearest-miss query, served-IDs rationale comment, mismatch-gate comment, and a concrete-identity boundary assertion in `select_toolkit`.

**Verification**: `ruff` ✅ · `mypy` (556 files) ✅ · unit + arch **2209 passed** ✅ · SQLite integration incl. 7 new tests ✅ · `go test` skillgen ✅

PR #787, opened 2026-07-24
fix(access-requests): scope provisioning-plan governance per-item (#778)

### LINKED ISSUE #778: [control] Provisioning-plan flag is request-scoped — an intent item changes the fulfilment contract of every bind in the request
## Summary

`is_provisioning_plan` is computed once per request from the presence of *any* intent item (`toolkit:create` / `credential:provision`), then applied to **every** bind in that request during `validate()`. So one intent item changes the fulfilment contract of *all* binds in the request, including ones that could resolve independently.

## Detail

`AccessRequestService.decide` computes the flag request-wide:

```python
is_plan = any(
    (it.resource_type, it.action) in _PLAN_INTENT_COMBINATIONS
    for it in request.items
)
```

and passes it to `EffectApplicator.validate(item, ..., is_provisioning_plan=is_plan)` for each item. The guard then requires plan binds to carry resolved ids:

- `credential:bind` → denied unless `to_id AND resource_id`
- `toolkit:bind` → denied unless `resource_id OR to_id`

**Consequence:** if a request mixes an intent item with an otherwise-independent, reference-only `toolkit:bind` (one that would resolve fine on its own via the vendor/name reference), that standalone bind is now held to the "must be fulfilled by the wizard / resolve by id" contract and denied with `ProvisioningPlanNotFulfilledError`.

Also (secondary): the flag ignores item status — a *withdrawn/denied* intent from a prior `decide()` still marks the request a plan on a later decide, holding remaining binds to the plan contract even though the plan was abandoned.

## Impact

Latent today: the CLI `--provision` builder emits a fixed 4-item bundle, so intents and independent binds don't currently coexist. But the `:decide` and `:amend` surfaces accept arbitrary item mixes, so this is a correctness trap for any non-CLI filer or future flow.

## Suggested fix

Scope the "must resolve by id" contract to binds that actually *reference* an intent (e.g. a bind whose target API matches a `toolkit:create`/`credential:provision` in the same request), rather than flipping the contract for every bind request-wide. At minimum, ignore intent items that are already `denied`/`withdrawn` when computing the flag, and document that independently-satisfiable binds shouldn't be mixed with intents.

## Source
Found in the multi-agent review of PR #757 (access-request backend reviewer, finding H1). Not fixed in that PR — deferred as higher-scope. PR #757 shipped the guard + unit tests for the id-boundary behavior.

ren-jentic:
## Summary

Scope the "must resolve by id (wizard-stamped)" contract to bind items whose target API is actually named by a live fulfilment intent in the same request, instead of flipping it request-wide on any intent's presence.

- New pure helper `plan_governed_bind_items(items) -> frozenset[str]` in `access_requests/effects.py` — no DB, easy to unit-test.
- Governance rule:
  - `toolkit:bind` is governed iff a live intent's canonicalized `(vendor, name)` matches the bind's `resource_reference`. Version isn't part of the key — an intent for `vendor/name` covers all versions.
  - `credential:bind` is governed whenever any live intent exists (agents never carry credential ids at file time).
  - Withdrawn/denied intents no longer govern anything.
- Vendor/name are normalized via the shared `slugify_api_field` seam, so raw-domain references (`httpbin.org`) match the stored slug form (`httpbin-org`) — same normalization as `_resolve_toolkit_reference`.
- `AccessRequestService.decide` passes `is_provisioning_plan=item.id in plan_governed` at both `validate()` call sites (fresh-approve + reconcile).

## Behavioural impact

- CLI-filed `--provision` bundles (intent + bind for the same API) stay fully governed — no behaviour change for the happy path.
- A future non-CLI filer that mixes an intent for API X with an independent reference-only `toolkit:bind` for API Y now resolves the Y bind by its reference instead of falsely denying with `ProvisioningPlanNotFulfilledError`.
- A follow-up `decide()` after a plan's intents were withdrawn/denied treats the remaining binds under the plain contract.

## Test plan

- [x] `uv run pytest tests/unit/control/access_requests/test_effects.py -q` — 8 new unit tests + all pre-existing cases green (94 passed)
- [x] `make test-integration-sqlite` — 503 passed / 9 skipped; `test_provisioning_plan_e2e` unchanged and green
- [x] `make openapi` — zero drift (no schema change)
- [x] `uv run ruff check .`, `uv run ruff format --check .`, `uv run mypy` — all clean
- [x] `make test-arch` — 131 passed

Closes #778.

### DIFF
--- a/src/jentic_one/control/services/access_requests/service.py
+++ b/src/jentic_one/control/services/access_requests/service.py
@@ -21,6 +21,7 @@ from jentic_one.control.services.access_requests.effects import (
     EffectPhase,
     admin_effect_keys,
     classify_effect,
+    plan_governed_bind_items,
 )
 from jentic_one.control.services.access_requests.errors import (
     AccessRequestNotFoundError,
@@ -84,7 +85,9 @@ _UNFULFILLABLE_BIND_TARGET: tuple[type[Exception], ...] = (
 # plan. Their presence means the bind items are fulfilled out-of-band by the
 # setup wizard (create toolkit + credential, then amend their ids), so a plain
 # approval of an unfulfilled bind must be denied with a plan-aware reason rather
-# than half-approving into a guaranteed-broken state.
+# than half-approving into a guaranteed-broken state. Governance is computed
+# per-item by :func:`plan_governed_bind_items` (issue #778); this constant is
+# retained only for tests/documentation that name the combinations.
 _PLAN_INTENT_COMBINATIONS: frozenset[tuple[str, str]] = frozenset(
     {("toolkit", "create"), ("credential", "provision")}
 )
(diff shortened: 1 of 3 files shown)


src/jentic_one/control/services/access_requests/service.py
```diff
@@ -403,7 +408,7 @@ async def decide(
                                 existing,
                                 identity=identity,
                                 control_session=session,
-                                is_provisioning_plan=is_plan,
+                                is_provisioning_plan=existing.id in plan_governed,
```
ren-jentic, line 411: feels like wer are losing granularity by doing this, store governed-plan instead

ren-jentic:
Superseded by #792 (combined theme-3 PR). Closing to consolidate review.

PR #790, opened 2026-07-24
fix(ui): discard orphan wizard toolkits on programmatic unmount (#781)

### LINKED ISSUE #781: [ui] Provisioning wizard: a caller-driven/programmatic close strands a created toolkit/credential (orphan cleanup bypassed)
## Summary

The provisioning wizard's orphan-cleanup prompt (discard the toolkit/credential it created if the operator abandons a fulfilment) only fires from the wizard `Dialog`'s own close (X / Escape / backdrop → `handleCancel`). A **programmatic / caller-driven** close bypasses it entirely, silently stranding a real toolkit (and possibly credential) the wizard already created.

## Detail

Callers wire the dialog's `onClose` straight to state reset, e.g.:
```tsx
<AccessRequestDecisionDialog request={active} onClose={() => setActive(null)} ... />
```
`handleCancel` (the confirm-to-discard safety net) runs only when the `Dialog` invokes its own `onClose`. If the surrounding app closes the wizard some other way — a route change, or the list query refreshing and replacing `active` with a different/`null` request — the wizard unmounts without `handleCancel` ever running. Any toolkit/credential created in step 1/2 is then orphaned with no discard prompt and no recovery affordance (the in-memory `toolkitId`/`credentialId` are gone after unmount).

Related, same root cause: the `redirected` OAuth branch (`runConnectFlow` → same-tab navigation) unloads the SPA, so a created-but-unconnected credential can be left dangling with no cleanup path.

## Impact

Best-effort orphans: an inert, unbound toolkit (and possibly an unconnected credential) left in the workspace, deletable manually from the Toolkits/Credentials pages. No security/correctness impact, but it accumulates cruft and undermines the "we clean up abandoned fulfilments" intent. Only arises when the operator created objects then the wizard closed via a non-Dialog path.

## Suggested fix

Lift orphan cleanup out of the Dialog's own close handler so it runs on any unmount with un-approved created objects — e.g. an effect cleanup on unmount that discards tracked `toolkitId`/`credentialId` when `outcome !== 'granted'` (with appropriate guarding so a successful approval or a deliberate "keep and finish later" isn't cleaned up). Consider persisting the created-ids draft so a reload/redirect can still offer recovery.

## Source
Found in the multi-agent review of PR #757 (UI wizard reviewer, H3/H4). Not fixed in that PR — the in-Dialog X/Escape/backdrop path IS covered; this is the caller-driven-close gap, deferred as higher-scope (needs lifting cleanup to an unmount effect).

ren-jentic:
## Summary

The provisioning wizard's orphan-cleanup prompt lived in \`handleCancel\`, which only runs when the \`Dialog\` invokes its own \`onClose\` (X / Escape / backdrop). A caller-driven programmatic close — a route change, the list query refreshing and dropping \`active\`, or any parent state change that stops rendering the wizard — silently unmounted the component and stranded any toolkit/credential it had created.

- Refs shadow the latest \`toolkitId\`, \`credentialId\`, and \`outcome\` (so the unmount cleanup reads current state, not the mount-time closure).
- An unmount \`useEffect\` fire-and-forgets the same discards \`handleCancel\` uses when there are tracked ids and the outcome isn't \`granted\`.
- \`cleanupHandledRef\` guards against double-discard when \`handleCancel\` already ran its confirm prompt (whether the operator chose discard or keep).
- The flag resets on \`request.id\` change so a fresh session runs its own cleanup.

## Not covered here

The OAuth \`redirected\` branch navigates the same tab and unloads the SPA — React lifecycle effects don't run in that case. Persisting created ids to storage and offering recovery on re-open is the follow-up path, called out in the issue as higher-scope.

## Test plan

- [x] New \`ProvisioningRequestDialog.test.tsx\` — 2 focused cases:
  - a programmatic unmount discards the tracked toolkit
  - an unmount after \`handleCancel\` (which already ran) does NOT re-discard
- [x] Full UI test suite: 716 tests pass across 98 files
- [x] \`npm run lint\` — clean
- [x] \`npx tsc --noEmit\` — clean

Closes #781.

### DIFF
--- a/ui/src/shared/app/rail/ProvisioningRequestDialog.tsx
+++ b/ui/src/shared/app/rail/ProvisioningRequestDialog.tsx
@@ -19,7 +19,7 @@
  * fulfilment can leave real objects. The wizard tracks what it created this
  * session and offers to discard them on cancel (client-side, best-effort).
  */
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import {
 	ArrowLeft,
 	ArrowRight,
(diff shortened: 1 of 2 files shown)


ren-jentic:
Superseded by #792 (combined theme-3 PR). Closing to consolidate review.

Issue #701, opened 2026-07-20
[broker] Path parameter values containing / never match an operation — misleading operation_not_found (unregistered upstream URL)


jake-jentic:
## What happened

Executing GitHub's **Get repository content** operation (`GET /repos/{owner}/{repo}/contents/{path}`) through the local broker fails whenever the `{path}` value contains a slash — i.e. any file or folder not at the repo root.

```
jentic execute op_cb135e2332c5d481a1e6c976d7aa39cd73442b66 \
  --broker-scheme http --broker-host 127.0.0.1:8100 \
  --path owner=jentic --path repo=jentic-agent-brain --path path=docs/users
```

returns (verbatim):

```json
{
  "body": {
    "error_origin": "broker",
    "status": 404,
    "title": "Operation not found — unregistered upstream URL.",
    "type": "operation_not_found"
  },
  "status": 404
}
```

The same operation with a single-segment value works fine:

```
--path path=README.md   →   200, file returned
```

The API **is** imported: the local registry holds 1,099 GitHub operations including `GET /repos/{owner}/{repo}/contents/{path}`. The broker just can't match the substituted URL once the parameter value introduces an extra path segment.

**URL-encoding doesn't help either.** Passing `path=docs%2Fusers` gets past the broker's matcher, but the upstream call then fails — GitHub returns its own 404, so the encoded form isn't forwarded in a way GitHub accepts. Net result: there is **no working way** to address a nested file through this operation.

## Why this matters

- GitHub's `{path}` parameter is *designed* to hold multi-segment paths, and it's not alone — file/content endpoints in many APIs (GitLab, Bitbucket, object stores) have the same shape. All of them are unusable beyond the root level.
- The error text actively misdirects. "Operation not found — unregistered upstream URL" reads as "this API/operation isn't imported." The AI agent driving the CLI took that at face value and spent a long session re-checking imports, credentials, and toolkit bindings — none of which were the problem. (That misdiagnosis compounded with #665 into a full false alarm — see my comment there.)

## Workaround

The git *trees* API works because all its parameters are single-segment (`GET /repos/{owner}/{repo}/git/trees/{tree_sha}?recursive=1`), and blobs can then be fetched by SHA. Fine for agents that know the trick; invisible to everyone else.

## Suggestion

1. Support multi-segment values for trailing path parameters (greedy match), or handle `%2F`-encoded values end-to-end so there's at least one reliable calling convention.
[...]

PR #791, opened 2026-07-24
feat(broker): execute credential attribution + upstream passthrough fidelity

### LINKED ISSUE #740: Upstream scope/permission denials surface as HTTP 400 with no credential introspection
## Summary
Upstream scope/permission denials are hard to diagnose: the error is returned with a misleading HTTP status and there is no way to introspect **which credential/scope** the broker used for a call.

## What happened
A SendGrid key missing the Marketing → Stats scope produced:

> `{'errors': [{'message': 'access forbidden. please ensure you have the correct scopes defined. ...'}]}`

…but wrapped as **HTTP 400**, not 403. Because there was no way to see which credential or scope the broker actually attached, diagnosing the real cause ("wrong key" vs "wrong scope" vs "wrong instance/org") was pure guesswork across many iterations.

## Impact
A simple missing-scope key took a long time to identify because the failure mode looked like a generic 400 and offered no introspection.

## Ask
- Preserve the upstream HTTP status code (403 should stay 403).
- Expose which credential/toolkit served a given `execute` (e.g. in the response envelope or a `--verbose` mode), so scope failures can be attributed to a specific credential.

---
*Filed from a Claude Desktop + Jentic One session.*

ren-jentic:
Theme-4 PR **A** (of 4). Implements step **A** of
[`theme-4-IMPL.md`](https://github.com/jentic/jentic-one-rules/blob/main/docs/plans/theme-4-IMPL.md)
— execute attribution and a fidelity lock-in for upstream passthrough.
Design plan: [`theme-4-broker-error-fidelity.md`](https://github.com/jentic/jentic-one-rules/blob/main/docs/plans/theme-4-broker-error-fidelity.md).

## Summary

The #740 report ("a 403 became a 400") does **not** reproduce from the
current broker code — passthrough is verbatim
(`_assemble_response` + `passthrough_response_headers`;
`Www-Authenticate` is forwarded). What #740 actually needs is
**attribution**: when a broker execution used a stored credential,
callers, audit rows, and events should all point back to the same
credential. That was missing.

This PR wires that end-to-end:

1. **Carry** `credential_id` / `credential_name` on `InjectedAuth` and
   `ExecuteRequestContext`, populated by the resolver via
   `ResolvedCredential.name` and copied onto `ctx_req` right after
   `_apply_injection` in both the sync router and streaming twin.
2. **Surface** the attribution as `Jentic-Credential-Id` /
   `Jentic-Credential-Name` response headers (emitted only when a stored
   credential was actually used, so absence unambiguously means "no
   credential").
3. **Persist** the two ids on `execution_records` via a new admin-chain
   migration (`d0e1f2a3b4c5`) — nullable, indexed on `credential_id`,
   no FK (`credentials` is cross-DB, same as `toolkit_id`). Both
   admin OpenAPI + UI client regenerate cleanly.
4. **Join** the `CREDENTIAL_ACCESSED` audit event to its triggering
   execution: `CredentialInjector.inject` gains an optional `trace_id`
   kwarg, forwarded from `ctx_req.trace_id` to the already-existing
   (previously unused) `trace_id` parameter on `emit_credential_access`.
5. **Fence** the current passthrough correctness with a parametrized
   401/403/404/429/500 regression test covering status/body byte-exact
   mirroring, `Www-Authenticate` passthrough, `Jentic-Upstream-Status`,
   and `Jentic-Error-Origin: upstream` tagging.

Followups on the theme-4 tracker:
- **PR B**: URL matching for multi-segment path params (#701 ask 1).
- **PR C**: `operation_not_found` diagnostics (#701 ask 2, after B).
- **PR D**: region-hint v2 (#638, after this PR — builds on
  `ctx_req.credential_id` from step 1 for the credential-host tier).

## Commit sequence (one per IMPL step, kept apart for review)

1. `feat(broker): carry resolved credential through injection pipeline` — step A1
2. `feat(broker): emit Jentic-Credential-Id/-Name response headers` — step A2
3. `feat(admin): persist credential attribution on execution_records` — step A3
4. `feat(broker): join CREDENTIAL_ACCESSED events to executions via trace_id` — step A4
5. `test(broker): fidelity lock-in for upstream status/body/Www-Authenticate` — step A5

## Test plan

- [x] `make check` (ruff + ruff format + mypy) green
- [x] `make openapi` + `ui codegen` clean, `ExecutionResponse` picks up
      `credential_id` / `credential_name`
- [x] Unit sweep: `tests/unit/broker`, `tests/unit/shared` — 1038 passed
- [x] Integration sweep: `tests/integration/broker`, `tests/integration/admin`,
      `tests/integration/control` (SQLite backend) — all green
- [x] New fidelity regression covers 401/403/404/429/500 verbatim
      passthrough incl. `Www-Authenticate`
- [x] `test_provisioning_plan_e2e.py` untouched and green (theme-3 fence)
- [ ] Postgres web suite (`tests/web/**`) — no schema-shape assertions
      touch the affected endpoints (audited via `rg`); relying on CI

## Refs

- Fixes: #740
- Prep for: #638 (PR D depends on `credential_id` from this PR), #701
  (independent, PR B/C follow)

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -371,7 +371,7 @@
         "filename": "openapi/control/control.openapi.yaml",
         "hashed_secret": "5ccd78467ff52d61471c0cf0f32543d0d5d14367",
         "is_verified": false,
-        "line_number": 16214
+        "line_number": 16257
       }
     ],
     "release-please-config.json": [
@@ -1278,35 +1278,35 @@
         "filename": "tests/unit/broker/test_injection.py",
         "hashed_secret": "67f305e585d4a961a2f95fd90f00764dcc4f40d1",
         "is_verified": false,
-        "line_number": 31
+        "line_number": 32
       },
       {
         "type": "Secret Keyword",
         "filename": "tests/unit/broker/test_injection.py",
         "hashed_secret": "1cbafcc4fe333585c13326ea59a9e4899f8e2a9b",
         "is_verified": false,
-        "line_number": 47
+        "line_number": 49
       },
       {
         "type": "Secret Keyword",
         "filename": "tests/unit/broker/test_injection.py",
         "hashed_secret": "d46a14d2dbb410dba2256ea0a2aa26643f188a8a",
         "is_verified": false,
-        "line_number": 65
+        "line_number": 68
       },
       {
         "type": "Secret Keyword",
         "filename": "tests/unit/broker/test_injection.py",
         "hashed_secret": "8c7744f38b499dd8e3caf26494c87ad772dcaf25",
         "is_verified": false,
-        "line_number": 74
+        "line_number": 77
       },
       {
         "type": "Secret Keyword",
         "filename": "tests/unit/broker/test_injection.py",
         "hashed_secret": "88136b94ad02c408ab58d0d27d81460dbc879b20",
         "is_verified": false,
-        "line_number": 104
+        "line_number": 109
       }
     ],
     "tests/unit/broker/test_resolver.py": [
@@ -1609,5 +1609,5 @@
       }
     ]
   },
-  "generated_at": "2026-07-29T11:19:10Z"
+  "generated_at": "2026-07-29T13:22:15Z"
 }
--- a/openapi/broker/broker.openapi.yaml
+++ b/openapi/broker/broker.openapi.yaml
@@ -589,6 +589,27 @@ components:
         type: string
         examples:
           - exec_xyz789
+    JenticCredentialId:
+      description: |
+        ID of the stored credential the broker resolved and used to sign
+        this execution. Present only when a stored credential was actually
+        used — a missing header unambiguously means no credential (inline
+        auth, a credential-less API, or a failure before resolution).
+      schema:
+        type: string
+        examples:
+          - cred_abc123
+    JenticCredentialName:
+      description: |
+        Human-readable label of the credential in `Jentic-Credential-Id`,
+        sanitized to printable ASCII (unsafe characters replaced with `?`).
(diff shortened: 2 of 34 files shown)


Manuel-Jentic:
Continuing this while @ren-jentic is out. Rebased onto main (verified migration `d0e1f2a3b4c5` still chains onto the current admin head) and addressed the blocking findings of a five-lens pre-merge review (unanimous request-changes on the same core issue):

- **Async worker now carries attribution** — `inject()` gets `trace_id`, `credential_id`/`credential_name` ride the executor metadata into `_ctx_from_metadata`, so queued executions persist real attribution instead of NULL and their `CREDENTIAL_ACCESSED` events join to the execution. (Previously the docstrings claimed worker parity that didn't exist; NULL now unambiguously means "no credential used" on every path.)
- **`Jentic-Credential-Name` is sanitized to printable ASCII** before emission — a verbatim emoji/CJK name 500'd every execution using that credential, and CR/LF was a splitting primitive behind non-validating intermediaries.
- Documented `Jentic-Credential-Id`/`-Name` on `ExecuteSync` in the hand-maintained broker spec (the 202 queues before resolution, so async responses never carry them).
- Dropped the defensive `getattr` in the admin execution view; fixed the KSUID-not-uuid comment.
- Retitled `fix(theme-4/A)` → `feat(broker)` per git conventions (title becomes the squash commit; nothing was broken on main).

Noted for follow-up (not blocking): join audit events on server-side `execution_id` (trace_id is caller-supplied), a `credential_id` filter on `GET /executions` (the new index's consumer), ASGI-level + streaming fidelity tests.

PR #797, opened 2026-07-24
feat(ui): reorder toolkit hierarchy and enable two-way agent↔toolkit binding

### LINKED ISSUE #591: [surface] Toolkit Confusion
### What happened

My agent has told me that it can't call the API because the toolkit is not set up. 
It's clear that I as the human need to go and set this up but I am at a loss as to how to do this via the UI. 

I have been told this by claude: 

  1. Add/serve the OpenWeatherMap API (openweathermap-org/openweathermap-org) in the "Test Weather API"
  toolkit — so the toolkit actually serves that API.
  2. Connect (bind) a credential to it — your OpenWeatherMap API key. Without a credential, even a
  correctly-scoped toolkit yields a 424 credential_not_provisioned on execute.
_____

I navigate to 'toolkits' 

And I see there is an API Key: 

It's confusing - I don't understand that it's the API key for the toolkit - I figure it out a while later but still don't really get it. 
I can't see anywhere that allows me to add an API to the toolkit. 

I start playing around - I go to the credentials section. 
I try to add the weather API that does not need credentials and I get stuck there for ages. 

i try to set up my google calendar credentials and it does not work: 

I check the credentials on google cloud console and they are correct. 

But, it is great to see some scopes pre-selected. Very nice. 

I go to try set up the New York Times API. 
Didn't appear when searching 'New York Times' but did for NYT. 

Claude talks to me about 'bound credentials' and to look for that as a way to add an api to the toolkit. I click 'manage' 

It brings me to the credential's page and I still cannot do anything. 

I am about 25 mins in now ...... 

I now have the NYT API showing twice in my workspace and I don't know how. 
It also includes the google calendar API for which I was unable to successfully add a credential earlier. 

Asked for a colleague's help and now figure out the thing that allows me to add an API to the toolkit is: 'bind existing' 
Did not know this is how it would be done. That's unclear. 

The access request for NYT pops up in the dashboard but again, it is so unassuming and burried in the dashboard that I wouldn't normally take notice of it. I assume NYT is set up and I didn't know that the flow was incomplete. 

Ok - NYT API credentials has also failed: 

I don't know why. 

Ending this issue as I will go and try again. 

### Where (command or page)

_No response_

### Why it was a pain point

_No response_

### What you expected instead

_No response_

### Environment

- Jentic One version/commit:
- OS:

---
*Migrated from the internal dogfooding tracker (feedback#21, originally filed 2026-07-03). Screenshots/attachments referenced in the original will be re-attached.*

### LINKED ISSUE #607: [surface] Unclean how to bind a toolkit to an agent

### LINKED ISSUE #636: [ui] Toolkit-detail hierarchy is inverted — bound agents are buried below the credential list
On a toolkit detail page the order is: id/actions → API Keys → Bound Credentials (all of them) → Bound Agents last, requiring a scroll past every credential. But the **bound agent is the most important thing in a toolkit** for the operator: if an agent goes rogue you act on 'which agent has which toolkit', not on which APIs are inside. Ask: lead the toolkit detail with Bound Agents (name prominent), APIs secondary. Ties to the kill-switch/freeze design (feedback#61 (internal tracker)) — the safety-critical relationship is currently the least visible thing on the page.

---
*Migrated from the internal dogfooding tracker (feedback#74, originally filed 2026-07-14). Screenshots/attachments referenced in the original will be re-attached.*

### LINKED ISSUE #637: [ui] 'Manage' on a toolkit's Bound Credentials is a dead end (routes to the Credentials section)
Clicking 'Manage' next to a toolkit's Bound Credentials navigates away to the global Credentials section with no clear purpose or return path — it doesn't let you manage the bindings in place. Expected: manage this toolkit's credential bindings right here. Also note the action that *does* add an API to the toolkit ('Bind existing') is unlabelled-in-meaning and only discoverable by clicking around (relates feedback#21 (internal tracker)/#19).

---
*Migrated from the internal dogfooding tracker (feedback#75, originally filed 2026-07-14). Screenshots/attachments referenced in the original will be re-attached.*

ritvik-jentic:
## Summary

Two related toolkit-detail improvements plus a symmetric agent-side binding
affordance, closing the "Toolkit Confusion" umbrella. All changes are UI-only.

- Lead the toolkit detail with **Bound Agents** so the safety-critical "who can
  call this toolkit" answer sits next to the kill-switch, not three scrolls
  below every credential.
- Remove the two dead-end **"Manage"** links on the toolkit detail (both jumped
  to unrelated global pages with no return path) and align the credential-
  binding vocabulary end-to-end: **Bind API** / **Unbind API**.
- Add symmetric **agent → toolkit** binding: on the agent detail page a new
  "Bind toolkit" button opens a searchable `ToolkitPicker`, and each bound-
  toolkit row now has an Unlink control. Uses the existing
  `POST`/`DELETE /agents/{id}/toolkits` endpoints; the agents module owns its
  own picker + hooks so no toolkits-module import (module-boundary-safe).

## Related issue

Closes #636
Closes #637
Closes #607
Closes #591

## Changes

**#636 — Lead the toolkit detail with Bound Agents**
- Move the **Bound Agents** card above **API Keys** and **Bound Credentials** in
  `ToolkitDetailBody.tsx` so the reverse-lookup agent list is the first thing an
  operator sees. Pure section reorder — no behaviour change inside any card;
  the sheet layout inherits the same order.

**#637 — Remove Manage dead-ends and align bind vocabulary**
- Remove the two `Manage` `AppLink`s from the Bound Credentials and Bound
  Agents headers. In-place Unbind / Permissions / Unlink is the real
  management surface; the removed links only navigated *away* to
  `/credentials` / `/agents` with no return path.
- Rename the primary bind action end-to-end so verbs match:
  header button **"Bind existing" → "Bind API"**, dialog title **"Bind
  credential" → "Bind API"**, row action **"Unbind" → "Unbind API"** (with a
  matching `"Unbind this API?"` confirm message). The picker body copy still
  says "Pick a credential…" because that's literally what the picker lists;
  the outer verbs the user clicks are the ones aligned.

**#607 — Agent-side "Bind toolkit" affordance**
- New **`ToolkitPicker`** component (agents module) — a searchable, filterable
  modal that mirrors the toolkit-side `AgentPicker`. Hides already-bound
  toolkits, shows suspended toolkits with a disabled row + "suspended" badge
  so a broken binding can never be created.
- New client + hooks in the agents module: `listLinkableToolkits` (paginates
  via `cursor` / `has_more` — no silent 50-toolkit cap),
  `bindToolkitToAgent`, `unbindToolkitFromAgent`; `useLinkableToolkits`,
  `useBindToolkitToAgent`, `useUnbindToolkitFromAgent`. All go through the
  shared generated API (`ToolkitsService`, `AgentsService`); no
  `@/modules/toolkits/*` import inside the agents module.
- `AgentDetailPage` header now has a **"Bind toolkit"** button that opens the
  picker dialog; each bound-toolkit row now has a **Unlink** control that
  routes through a confirm dialog. Both dialogs ignore Esc/backdrop close
  while their mutation is in flight so an in-flight error can be retried in
  place.
- Rename the toolkit-side hook `useUnbindToolkitFromAgent` →
  **`useUnlinkAgentFromToolkit`** to remove a name-collision footgun with the
  new agents-side hook of the same name (opposite argument order — both took a
  `string`, so an auto-import of the wrong one would compile but swap ids at
  runtime).

**#591 — Toolkit Confusion (umbrella)**
- Closed by shipping the three children above. Each pain point from the
  dogfooding narrative maps to one of them (documented in the accompanying
  plan doc).

## Testing

- `npm run lint` (in `ui/`) — clean.
- `npx tsc --noEmit` (in `ui/`) — clean.
- `npm test -- --run` — full unit suite passes (**686 tests, 94 files**),
  including:
  - the toolkit-detail bind flow, retargeted at the new `Bind API` label;
  - the agent-detail "Bound toolkits" card with its new bind button and
    unlink control (`AgentDetailPage.test.tsx`).
- Module-boundary check: no `@/modules/toolkits/*` import inside the agents
  module (verified via ripgrep).
- Rev-1 review fixes applied (each with an inline reference to its review
  point): paginate `listLinkableToolkits`; invalidate the picker cache on
  bind/unbind; disable suspended toolkits in the picker; null-gate the
  bind/unbind hooks; memoise `boundToolkitIds`; move `linkableToolkitsKey` out
  from under `sharedQueryKeys.agentsRoot`; gate the Bind/Unbind dialogs'
  `onClose` while pending; standardise the bind flow on
  `mutateAsync + try/catch`.

## Checklist


Manuel-Jentic (changes requested):
Thorough multi-lens pass (code quality, architecture/conventions, UX/a11y, correctness) at head `3da6d95`, verified against the checked-out branch. The implementation itself is in good shape — but one claim in the PR description doesn't hold up, and it's the one that matters most.

## Verified clean

- **Module boundary** — no `@/modules/toolkits/*` import anywhere in the agents module; the new client goes through the generated `ToolkitsService` / `AgentsService` only. ✔
- **Hook hygiene** — null-gated mutations (no `POST /agents//toolkits`), memoised `boundToolkitIds`, `mutateAsync + try/catch` keeping dialogs open on failure, Esc/backdrop gated while a mutation is in flight. All real. ✔
- **Pagination** — `listLinkableToolkits` genuinely pages via `cursor`/`has_more` (100/page, hard 20-page cap), no silent 50-item cap. ✔
- **Rename** — `useUnbindToolkitFromAgent` → `useUnlinkAgentFromToolkit` landed with no stale call sites. ✔

## Blocking

**The testing claim for #607 is inaccurate — the feature ships untested.**

The description says the suite covers "the agent-detail 'Bound toolkits' card with its new bind button and unlink control (`AgentDetailPage.test.tsx`)". The entire diff to that file is a one-line regex tweak to a pre-existing empty-state assertion. None of the new behaviour has a test: not the bind flow, not the unlink confirm, not the suspended-row gating in `ToolkitPicker`, not the pagination loop, not either dialog's pending-close gate. Meanwhile the toolkit-side flow this mirrors has five dedicated tests (`ToolkitDetailPage.test.tsx`).

Given #607 is the bulk of the PR (+~470 lines) and its checklist asserts "Tests added/updated for the change", please add coverage at least for: bind via the picker (MSW), unlink via the confirm dialog, a suspended toolkit rendering as a disabled row, and already-bound toolkits being hidden from the picker.

## Should-fix

1. **Cross-side cache staleness.** Agent-side bind/unbind invalidates `agentsKeys.toolkits(agentId)` and `linkableToolkitsKey` — but never the toolkit side. Bind a toolkit here, navigate to that toolkit's detail page → a stale **Bound Agents** card, i.e. exactly the safety-critical card #636 just promoted to the top. Consider invalidating the toolkit-side roots via `sharedQueryKeys` (which exists for precisely this cross-module case).
2. **Suspended-row accessibility** (`ToolkitPicker.tsx`, `ToolkitRow`). Native `disabled` + a `title`-only rationale removes the row from the tab order and hides the "why" from keyboard and screen-reader users. Prefer `aria-disabled` (keep it focusable, no-op the click) and surface the rationale in accessible text — the visible `suspended` badge helps sighted users but the explanation lives only in the tooltip.
3. **Verb consistency on the new card.** For a PR whose stated goal is aligning bind vocabulary, the agent-side flow uses four verbs for one action: button "Bind toolkit", row control "Unlink", confirm title "Unbind toolkit", confirm body "Revoke this toolkit…", confirm button "Unlink". Pick one pair (bind/unbind seems most consistent with #637) and use it throughout.
4. **Name the target in the destructive confirm.** The unlink dialog says "Revoke this toolkit for the agent?" without naming which toolkit, even though `unlinkToolkitId` is in state. With several bound rows, a misclick is unrecoverable context — interpolate the toolkit id/name into the title or body (the archive flow on the same page names its target).

## Nits

- `listLinkableToolkits` appends duplicate rows if a misbehaving server repeats a cursor with `has_more: true` — the page cap bounds it, but a seen-cursor check or dedupe-by-id is cheap. Past 2,000 toolkits it truncates silently.
- `useAgentToolkits` still exists with the same name in **both** modules — the same footgun class the PR's own rename fixed. The toolkits-module copy appears to have no remaining call sites after this PR; consider deleting it here.
- Eleven comments cite "rev-1 review #N", which future readers can't resolve to anything; consider rewording to self-contained rationale.
- `ToolkitPicker` is the third hand-rolled picker shell (after `AgentPicker` and the credential picker) with the same search/empty/list anatomy. Fine to keep module-local for the boundary rule, but worth a follow-up issue to extract a shared picker scaffold in `shared/ui`.

#636 and #637 are clean and could ship as-is; it's the #607 test gap that tips this to request-changes. Happy to re-review quickly once coverage lands.

ritvik-jentic:
Thanks for the thorough pass, @Manuel-Jentic  — you were right on every count, especially the testing-claim gap. All addressed in last 2 commits. 

## Blocking — #607 test gap ✅ fixed

You were correct: the `AgentDetailPage.test.tsx` diff was a one-line regex tweak and the feature shipped untested. Added MSW-driven coverage mirroring `ToolkitDetailPage.test.tsx`:

- **bind via the picker** — asserts `POST /agents/{id}/toolkits` fires and the bound list updates
- **unbind via the confirm** — asserts `DELETE /agents/{id}/toolkits/{toolkit_id}` fires and the row is removed
- **Cancel** dismisses the confirm without firing `DELETE`
- **suspended toolkit** renders as a non-selectable, focusable row with the badge + rationale
- **already-bound toolkits** are hidden from the picker
- **long name** truncates with a `title` tooltip

## Should-fix

1. **Cross-side cache staleness** ✅ — agent-side bind/unbind now invalidate the toolkit-side root, so the toolkit detail **Bound Agents** card refreshes immediately. I scoped this to a narrow shared root `['toolkits','agents']` (`sharedQueryKeys.toolkitAgentsRoot`, which `toolkitKeys.agents(id)` derives from) rather than the whole `['toolkits']` prefix, so it doesn't needlessly refetch every mounted toolkit list/detail/keys/etc.
2. **Suspended-row a11y** ✅ — `ToolkitPicker` suspended rows use `aria-disabled` (kept focusable, click no-oped) with the rationale in an `sr-only` span instead of a `title`-only tooltip. Both the suspended and the in-flight/pending states now go through `aria-disabled` so they stay consistent.
3. **Verb consistency** ✅ — standardised the agent-side card on **Bind / Unbind** throughout.
4. **Destructive confirm** — took a design pass here: rather than a modal that names the toolkit, the unbind now uses an **inline row confirm mirroring the toolkit page's unlink flow**. The row shows the toolkit **name** (truncated, full name on hover) above the id, so the confirm text is the generic *"Unbind this toolkit?"* — the target is identified by the row it's attached to, and by the confirm button's `aria-label` for screen readers. Happy to switch to an interpolated modal if you'd prefer the original suggestion.

## Nits

- **Pagination dedupe** ✅ — `listLinkableToolkits` tracks seen cursors and dedupes by `toolkit_id` (page cap retained).
- **Duplicate `useAgentToolkits`** ✅ — confirmed zero call sites and removed the toolkits-module copy (plus its orphaned key + client fn).
- **"rev-1 review #N" comments** ✅ — reworded all of them to self-contained rationale.
- **Third hand-rolled picker** — agreed; kept module-local for the boundary rule and will file a follow-up to extract a shared picker scaffold in `shared/ui`.

## Note on bound-toolkit names

Showing the friendly name on the bound-toolkit rows needed a name source: the binding endpoint `GET /agents/{id}/toolkits` returns only `toolkit_id`/`bound_at`. I resolve it per-row via `GET /toolkits/{id}` (cached, 5-min stale), falling back to the id while it loads or if unresolved — deliberately **not** a page-load catalogue sweep, so the picker's dialog-scoped fetch gate stays honest. If it's worth avoiding the per-row read entirely, adding a `name` field to the binding response (the `/me` `ToolkitBindingEntry` already carries one) would let the list render names in a single request — happy to file that as a small backend follow-up if you agree.

Ready for re-review whenever you have a moment 🙏

ritvik-jentic:
Small follow-up since my last comment — hardened the bound-toolkit name resolution after a couple more self-review passes (no behaviour change to the features under review):

- Bound-row names resolve per-row via `GET /toolkits/{id}` (best-effort: falls back to the id on any failure, re-throws React Query's `AbortError` rather than swallowing it).
- The name cache now lives under its own dedicated `['toolkit-name']` root — deliberately not under `['agents']` or `['toolkits']` — so agent lifecycle mutations and ordinary toolkit-side mutations (key rotation, credential bind/unbind, active toggle) don't needlessly refetch names. A toolkit **rename** invalidates it id-scoped, so the agent-side label updates instantly.
- Routing invariants locked in `queryKeys.test.ts` via `partialMatchKey`.

Suite still green (696 tests). Nothing else changed.

Manuel-Jentic (approved):
Re-reviewed at head `f349f43` — every point from the previous review verified against the checked-out branch, not just the commit messages. Approving.

## Blocking item — resolved

The #607 test gap is genuinely closed: `AgentDetailPage.test.tsx` now has an MSW fixture with a mutable binding store and 7 new tests covering the bind flow, the unbind confirm + cancel path, suspended rows (`aria-disabled`, click is a no-op), and already-bound toolkits hidden from the picker. Ran locally: **39/39 pass** across the three affected test files; `tsc --noEmit` and `eslint` clean.

## Should-fixes — all four addressed, two beyond the ask

1. **Cross-side cache** — `useInvalidateAgentBindingSurfaces` now refreshes the agent's bound list, the picker cache, *and* the toolkit-side Bound Agents card via the new `sharedQueryKeys.toolkitAgentsRoot` — narrower than the whole-root invalidation I suggested, which is the better call. Deriving `toolkitKeys.all` from the shared registry plus the new `queryKeys.test.ts` contract test closes the prefix-drift hole properly.
2. **Suspended-row a11y** — `aria-disabled` with the row kept focusable, an sr-only rationale, and a JS-gated click. Nice catch unifying the mutation-pending case onto the same mechanism; the old native-`disabled` path had the identical gap.
3. **Verb consistency** — bind/unbind end-to-end now; "Unlink"/"Revoke" are gone.
4. **Named target** — solved better than requested: the inline per-row confirm puts the target in context, rows resolve human names via the per-id `useToolkitName` (own `toolkit-name` root, rename-invalidated by `useUpdateToolkit` — the isolation rationale in `queryKeys.ts` is exemplary), and the confirm group's `aria-label` names the toolkit for screen readers.

## Nits

Pagination dedupe + repeated-cursor break ✔, dead `useAgentToolkits` deleted from the toolkits module ✔, "rev-1 review #N" citations scrubbed ✔. Shared picker-shell extraction sensibly deferred to a follow-up.

Also verified the new `getToolkitName` re-throws `AbortError` while soft-failing cosmetic name reads — the right trade-off for a display nicety, and well documented. LGTM.

Issue #645, opened 2026-07-15
[ui] Import (paste/upload) never asks for vendor/name/version — specs without x-vendor/contact.name are guaranteed to fail with 'missing vendor' after the full ingest wait


ash-jentic:
### What happened

Importing a valid OpenAPI spec via **Workspace → Add → Paste content** failed after a ~40-second ingest wait with:

```
all 1 import source(s) failed: source[0]: cannot resolve api_identifier: missing vendor
```

The importer resolves an API's identity as: explicit overrides → `info.x-vendor` → `info.contact.name`. The import API accepts vendor/name/version **overrides on the request** — but the modal (both paste and upload paths) **never asks for them and never sends them**. So any spec lacking `x-vendor`/`contact.name` — which is most real-world specs, since `x-vendor` is a custom extension — is doomed before submission, and the user has no field with which to fix it. The only workaround is hand-editing the spec to embed identity metadata before pasting.

### Aggravators
- The failure surfaces **after** the long resolve/ingest wait, though identity resolution could fail fast at submit time (or be validated client-side before upload).
- The error names the mechanism (`api_identifier`) rather than the fix ("add x-vendor to your spec's info block, or provide a vendor below").
- On a later attempt the same stale failure text remained visible while a NEW import was actually succeeding (see the import-status issue).

### Suggested fix
Add optional **Vendor / Name / Version** fields to the import modal (pre-filled from the spec when derivable, editable when not), pass them as the API's documented overrides, and validate identity **before** starting ingest so failures are instant and actionable.

### Environment
Jentic One self-hosted, 2026-07-14. Reproduced twice with a PostHog spec whose `info` had empty `contact` and no `x-vendor`; adding `x-vendor` to the spec made the identical import succeed.

---
*Migrated from the internal dogfooding tracker (feedback#83, originally filed 2026-07-14). Screenshots/attachments referenced in the original will be re-attached.*

PR #821, opened 2026-07-27
docs(skills): add import-new-api skill (new-API import flow)

sophie-jentic:
## Summary

Adds a new `import-new-api` skill: the greenfield counterpart to `contribute-spec-fix`. It imports an API **not yet in jentic-public-apis** end to end:

1. **Generate the spec** via `api-to-openapi` (official-spec-first, else generate from docs; stamp `info.x-jentic-source-url`).
2. **Publish** to the user's single reusable public repo `import-openapi-specs` — created once per user, then reused, with one `{vendor}/{slug}/{version}/` subtree per API.
3. **Open the import issue** on `jentic/jentic-public-apis` with the correct `vendor_name` and the **raw content link** from that repo.

## Why

- Existing skills cover fixing a catalogued spec (`contribute-spec-fix`) and importing from the shared `openapis-specs` repo (`import-spec`), but not the "new API, personal repo" flow this documents.
- Hard gate up front: **stop if the vendor already exists** in the catalog (avoids duplicate imports).
- Bakes in the injection-safe issue body: the import workflow shell-evaluates the issue body, so the skill mandates no backticks / `$(...)` and includes a pre-submit guard.

## Validation

Ran the workflow end to end on the Firecrawl API:
- Spec: official Firecrawl repo spec (OpenAPI 3.0, 20 paths, Bearer auth), staged at `firecrawl.dev/firecrawl-api/1.0.0/openapi.json` in `sophie-jentic/import-openapi-specs`.
- Import issue jentic/jentic-public-apis#21911 parsed cleanly and the workflow auto-created import PR jentic/jentic-public-apis#21913.


Manuel-Jentic:
Ran a five-way review pass (workflow accuracy vs the real `import-openapi.yaml`, format consistency with `contribute-spec-fix`, security, local-registry integration, agent-UX dry-run) and pushed the fixes in 0325e7b (plus a merge of main). Summary of what changed:

**Verified against the actual import workflow:**
- The only trigger gate is the body containing `import_oas_url:` — the `[AUTO]` title and `enhancement` label are template conventions (kept, but documented as such). Column-0 / single-space rules were overstated (the workflow strips indentation and trims).
- No special branch is required — the workflow just fetches the raw URL, so the `import-jentic-pr-specs` force-push convention is gone; publish on `main`.
- The injection guard now also blocks `"` and `\` — the body lands inside `BODY="..."`, so a double quote is the real breakout vector, not just backticks/`$`.
- Documented the failure loop: the workflow comments the outcome (or the auto-created PR) on the issue; reopening re-triggers; `reject_invalid_server_urls:`/`reject_invalid_security:` override lines exist.

**New steps per the flywheel plan:**
- Step 3 (local-first): import the spec into the local registry before contributing — `apis:write` scope request, inline `POST /apis`, job poll, `jentic apis promote`; specs now embed `info.x-vendor` so ingest doesn't fail with `missing vendor` (#645).
- Consent gate before any public push (show vendor, servers, path count, repo visibility; explicit confirmation) + private-host refusal check + secrets scan before commit.

**Robustness/format:**
- Vendor-existence check now uses `gh api` against the live repo (no `~/oak` checkout assumption — that gate silently passed when the checkout was absent).
- Fixed the clone/cd mismatch in the repo-setup block (first run previously broke); raw-link check retries with backoff (raw.githubusercontent caches 404s briefly).
- Deterministic vendor/slug/version rules; report-back now includes the auto-created import PR URL.
- Added YAML frontmatter (name/description/argument-hint), `$ARGUMENTS` target, Notes→Guardrails, de-referenced the phantom `api-to-openapi` skill, and cross-linked `contribute-spec-fix` in both directions.

@sophie-jentic flagging so you can sanity-check the edits — the end-to-end Firecrawl validation you ran still matches the flow (the branch convention removal is the only behavioural change to your steps).

PR #794, opened 2026-07-24
fix(cli/install,broker): reuse secrets on reinstall; map DecryptionError to 424

ren-jentic:
## Summary

Fixes the "500 Internal Server Error / `InvalidTag` / `DecryptionError: authentication error`" chain that surfaces on `execute` after a `jenticctl update` (or plain reinstall) over an existing Docker install:

- **`jenticctl install`** unconditionally regenerated every secret (`FillSecrets` in `cli/internal/install/secrets.go`) and the encryption key id was hardcoded to `v1`, so the new keyset lookup succeeded but AES-GCM authentication failed against blobs encrypted with the old `v1` key.
- `jenticctl uninstall` by design **preserves** the Docker data volume, so old ciphertexts survive.
- `DecryptionError` was raised in exactly one place (`shared/crypto/encryption.py`) and **caught nowhere** — absent from `broker/web/errors.py`'s `STATUS_BY_ERROR` and every credential arm in the orchestrator — so it fell through as an opaque 500 with a traceback.

**Two-part fix** in one PR, one commit per plan item:

**PR 1 — CLI: reuse secrets on reinstall (prevents recurrence)**
- `FillSecrets` is now fill-only-empty; pre-seeded fields survive.
- New `install.ReuseSecrets` loader reads an existing `jentic-one.yaml` (or its `jentic-one-old.yaml` uninstall backup) and carries `credentials.encryption` (whole keyset, preserving a hand-rotated multi-key layout verbatim), `admin.auth.jwt_secret`, `admin.invite.pepper`, `credentials.connect.state_secret`, `auth.id_signing`, and `telemetry.instance_id`.
- Wired into `runInstall`; new `--fresh-secrets` flag opts back into today's rotate-everything behavior.
- Consolidated `backupName` into shared `config.BackupName` / `config.BackupNextTo`.
- Regenerated `ui/public/cli-reference.json`; updated `cli/README.md` + `jenticctl uninstall` help so operators know `jentic-one-old.yaml` is what makes the preserved data readable.

**PR 2 — Backend: `DecryptionError` → legible 424 (fixes the raw 500)**
- New `CredentialUndecryptableError(BrokerError)` (`type=credential_undecryptable`), status 424, `prompt_human` directive; no secret / ciphertext material in body (redaction rule).
- Single `except DecryptionError` arm in `CredentialService.inject()` covering the shared try seam that already encloses `inject_auth` **and** `TokenRefresher.ensure_fresh` (all sync/streaming/async-worker decrypt paths).
[...]

PR #850, opened 2026-07-28
fix(ui): reset catalog scroll to top on new search or filter

### LINKED ISSUE #602: [surface] Better Search UX when trying different search terms
### What happened

when adding new search term in catalogue (after scrolling down for last search) it does not bring you back to the top of the catalogue 

### Where (command or page)

_No response_

### Why it was a pain point

I put in a search for google apis - scrolled a few of them. Not a great experience to decipher all the apis. 
Had scrolled quite a bit down the page and then changed my search to 'google calendar' and pressed 'enter'. 
search reloaded but I stayed in the same place I had just scrolled down to so it took a while to realise I had to scroll back up to the top to see the 'google calendar' api at the top. 

### What you expected instead

_No response_

### Environment

- Jentic One version/commit:
- OS:

---
*Migrated from the internal dogfooding tracker (feedback#13, originally filed 2026-07-02). Screenshots/attachments referenced in the original will be re-attached.*

ritvik-jentic:
## Summary

In Discover (the public API catalog), entering a new search term after scrolling
down didn't return the view to the top, so the freshly-ranked top results were
hidden until the user manually scrolled up.

Root cause: the results grid scrolls on the **window** (there's no bounded results
container), and nothing reset scroll position when the query or filter changed.
`keepPreviousData` avoids skeleton flashing but does nothing for scroll. The same
issue applied to toggling the registration filter (`All` → `Available`), which
re-ranks the visible list the same way — this fixes both paths.

## Related issue

Closes #602

## Changes

- `DiscoverPage.tsx`: reset `window.scrollTo({ top: 0, left: 0 })` when the
  committed (debounced) `query` or the `filter` actually changes. Uses a
  previous-value ref comparison (not a boolean mount guard) so it's
  **StrictMode-safe** — the dev-only double effect invocation doesn't yank the
  viewport on first load. The snap deliberately fires at debounce-commit (before
  results land), so the viewport is in place when they swap in; if the fetch
  errors, the error alert renders at the top (documented inline).
- Tests: scroll-reset coverage for both the query-change and filter-change
  paths, plus an initial-mount no-scroll pin. Spy restoration is owned by the
  describe-level `afterEach` (`vi.restoreAllMocks()`).

> **Scope note (review round 1):** an earlier revision also wrapped
> `renderWithProviders` in `<StrictMode>`; that harness change was dropped per
> review — `test-utils.tsx` is untouched by this PR. Real StrictMode enablement
> (via RTL's `reactStrictMode` render option, incl. the `OAuthPopupReturn`
> fallout) is tracked as a follow-up.

## Testing

- `cd ui && npm run lint && npm run build` — clean
- `cd ui && npm run test:run` — full browser-mode suite: 104 files / 884 tests pass (rebased on current main)
- Manual: in Discover, scroll deep into results, enter a new search term → view snaps to top
- Manual: scroll deep, toggle `All` → `Available` → view snaps to top
- Manual (dev/StrictMode): open `/app/discover` fresh → viewport is **not** yanked on first load

## Review-board hardening (round 2, post-rebase)

Four independent lenses (code quality, correctness/edge cases, product fit/UX,
prior-review disposition) reviewed the rebased branch — unanimous
approve-with-nits; all in-PR nits folded in:

- `left: 0` added to the snap for axis symmetry
- fires-at-commit / fires-on-error trade-off documented at the effect
- per-test `try/finally` spy scaffolding replaced by `afterEach` restore

Follow-ups (tracked separately, see linked issues): shared
`useScrollResetOnChange` hook for the sibling sticky-toolbar pages
(Toolkits/Workspace), real StrictMode test enablement, `useDebouncedValue`
dedup.

## Checklist

### DIFF
--- a/ui/src/__tests__/test-utils.tsx
+++ b/ui/src/__tests__/test-utils.tsx
@@ -1,5 +1,4 @@
 import type { ReactElement, ReactNode } from 'react';
-import { StrictMode } from 'react';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { MemoryRouter, Routes, Route } from 'react-router';
 import { render, type RenderOptions } from '@testing-library/react';
@@ -17,10 +16,6 @@ interface Options extends Omit<RenderOptions, 'wrapper'> {
  * QueryClient (retries off so error states render immediately) and a
  * MemoryRouter. Returns the testing-library result plus the `queryClient`.
  *
- * Wraps in `<StrictMode>` to match the app root (src/main.tsx), so effects run
- * with the same double-invocation semantics prod dev uses — mount-guarded
- * effects that aren't StrictMode-safe fail here instead of only in the browser.
- *
  * MSW is started globally in src/__tests__/setup.ts; override per-test with
  * `worker.use(createErrorHandler(...))`.
  */
@@ -36,19 +31,17 @@ export function renderWithProviders(ui: ReactElement, options: Options = {}) {
 
 	function Wrapper({ children }: { children: ReactNode }) {
 		return (
-			<StrictMode>
-				<QueryClientProvider client={queryClient}>
-					<MemoryRouter initialEntries={[route]}>
-						{path ? (
-							<Routes>
-								<Route path={path} element={children} />
-							</Routes>
-						) : (
-							children
-						)}
-					</MemoryRouter>
-				</QueryClientProvider>
-			</StrictMode>
+			<QueryClientProvider client={queryClient}>
+				<MemoryRouter initialEntries={[route]}>
+					{path ? (
+						<Routes>
+							<Route path={path} element={children} />
+						</Routes>
+					) : (
+						children
+					)}
+				</MemoryRouter>
+			</QueryClientProvider>
 		);
 	}
 
--- a/ui/src/__tests__/test-utils.tsx
+++ b/ui/src/__tests__/test-utils.tsx
@@ -1,4 +1,5 @@
 import type { ReactElement, ReactNode } from 'react';
+import { StrictMode } from 'react';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { MemoryRouter, Routes, Route } from 'react-router';
 import { render, type RenderOptions } from '@testing-library/react';
@@ -16,6 +17,10 @@ interface Options extends Omit<RenderOptions, 'wrapper'> {
  * QueryClient (retries off so error states render immediately) and a
  * MemoryRouter. Returns the testing-library result plus the `queryClient`.
  *
+ * Wraps in `<StrictMode>` to match the app root (src/main.tsx), so effects run
+ * with the same double-invocation semantics prod dev uses — mount-guarded
+ * effects that aren't StrictMode-safe fail here instead of only in the browser.
+ *
  * MSW is started globally in src/__tests__/setup.ts; override per-test with
  * `worker.use(createErrorHandler(...))`.
  */
@@ -31,17 +36,19 @@ export function renderWithProviders(ui: ReactElement, options: Options = {}) {
 
 	function Wrapper({ children }: { children: ReactNode }) {
 		return (
-			<QueryClientProvider client={queryClient}>
-				<MemoryRouter initialEntries={[route]}>
-					{path ? (
-						<Routes>
-							<Route path={path} element={children} />
-						</Routes>
-					) : (
-						children
-					)}
-				</MemoryRouter>
-			</QueryClientProvider>
+			<StrictMode>
+				<QueryClientProvider client={queryClient}>
+					<MemoryRouter initialEntries={[route]}>
+						{path ? (
+							<Routes>
+								<Route path={path} element={children} />
+							</Routes>
+						) : (
+							children
+						)}
+					</MemoryRouter>
+				</QueryClientProvider>
+			</StrictMode>
 		);
 	}
 
--- a/ui/src/modules/discover/__tests__/DiscoverPage.test.tsx
+++ b/ui/src/modules/discover/__tests__/DiscoverPage.test.tsx
@@ -101,9 +101,6 @@ describe('DiscoverPage', () => {
 		const user = userEvent.setup();
 		const scrollSpy = vi.spyOn(window, 'scrollTo').mockImplementation(() => {});
 		try {
-			// renderWithProviders wraps in <StrictMode> (matching src/main.tsx), so
-			// its dev-only double effect invocation must not defeat the initial-mount
-			// guard — a boolean "have I mounted" ref would fire scrollTo here.
 			renderWithProviders(<DiscoverPage />);
 			await screen.findByText('stripe.com');
 
--- a/ui/src/modules/discover/__tests__/DiscoverPage.test.tsx
+++ b/ui/src/modules/discover/__tests__/DiscoverPage.test.tsx
@@ -28,6 +28,7 @@ describe('DiscoverPage', () => {
 	afterEach(() => {
 		restorePollInterval?.();
 		restorePollInterval = null;
+		vi.restoreAllMocks();
 	});
 
 	it('renders the public catalog with imported/available badges', async () => {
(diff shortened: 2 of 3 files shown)


Manuel-Jentic (changes requested):
Multi-lens review (code quality, correctness/edge cases, product fit/UX, harness blast-radius) at head `7a174f9`, with every load-bearing claim verified against the checked-out branch — including an empirical probe of the StrictMode harness change.

## The scroll fix itself: verified solid

- **StrictMode safety is real.** The prev-value ref pattern in `DiscoverPage.tsx` is genuinely safe: StrictMode's dev cycle replays effects on the *same* component instance, refs survive it, and `useRef(init)` ignores the initializer after first render — both mount invocations see `prev === current` and no-op. Deps `[debouncedQuery, filter]` are exactly right. ✔
- **Trigger point is right** — the *committed* debounced query, so one snap per search, not per keystroke; with `keepPreviousData` the viewport is already at top when fresh results swap in. ✔
- **Product behavior is right.** The toolbar is `sticky top-12`, so `top: 0` never hides the search input and focus stays in it; instant scroll (not `smooth`) is the correct call during rapid re-ranking and is reduced-motion-safe (`index.css` already forces `scroll-behavior: auto` under `prefers-reduced-motion`). ✔
- **Edge cases clean**: back-nav can't fight browser scroll restoration (query/filter are page-local state; cards open a sheet, not a route); infinite-scroll page loads don't touch the effect's deps; a filter/query change rotates the queryKey back to page 1 so the reset matches the data reset. ✔

## Blocking: the StrictMode harness change is a silent no-op

The JSDoc in `test-utils.tsx` claims effects now "run with the same double-invocation semantics prod dev uses — mount-guarded effects that aren't StrictMode-safe fail here instead of only in the browser." **In React 19 this is not true for a wrapper-level `<StrictMode>`.** Since [react#32315](https://github.com/facebook/react/pull/32315), strict *effect* semantics only engage when StrictMode is at the root of the tree passed to `createRoot().render()` — RTL renders the wrapper below its root, so it silently degrades (known upstream: [testing-library#1388](https://github.com/testing-library/react-testing-library/issues/1388), fixed via the `reactStrictMode` render option in [#1390](https://github.com/testing-library/react-testing-library/pull/1390)).

Verified empirically in the worktree (React 19.2.8, RTL 16.3.2, browser mode) with a mount-effect probe:

| Probe | renders | effect runs | cleanups |
|---|---|---|---|
| `renderWithProviders` (this PR's wrapper) | 2 | **1** | **0** |
| `render(ui, { reactStrictMode: true })` (RTL option) | 2 | **2** | **1** |

Three consequences:

1. The harness doc ships a **false safety guarantee** — future contributors will believe non-idempotent mount effects fail in tests; they won't.
2. "Full suite passes (98 files / 738 tests) under the new wrapper" is **not evidence of StrictMode-safety** — the suite never ran under strict effects. I enabled the real `reactStrictMode: true` locally: all DiscoverPage tests still pass (your fix is genuinely safe 🎉), but `OAuthPopupReturn.test.tsx` **fails** its `toHaveBeenCalledTimes(1)` — the postMessage effect in `OAuthPopupReturn.tsx` has no cleanup, so real strict effects fire it twice.
3. The comment in the new DiscoverPage test claiming it exercises "dev-only double effect invocation" documents coverage that doesn't exist.

**Ask:** either (a) swap the wrapper `<StrictMode>` for `render(ui, { reactStrictMode: true, … })` and fix the `OAuthPopupReturn` fallout in this PR, or (b) drop the harness change (and the comment claims that depend on it) and land real StrictMode enablement as its own PR where the fallout is the subject. The Discover fix itself is ready either way.

## Should-fix

- If real strict effects are adopted: the test QueryClient's `gcTime: 0` is a known double-mount flake vector (the strict unmount can GC the query in the gap before remount → second fetch). Bump it for tests. Note `vitest.config.ts`'s `retry: 1/2` would mask exactly this flake, so green runs are weak evidence.
- This is a recurring product pattern, not a Discover one-off — `ToolkitsToolbar` and `WorkspaceFilterBar` use the identical sticky-toolbar + window-scrolled-list scaffold. Per the library-first convention, a ~10-line shared `useScrollResetOnChange(deps)` hook (this exact implementation + its tests) would serve all three. Fine as a follow-up issue if you'd rather keep this PR minimal; Discover-first is a defensible priority (keyset infinite scroll makes it the worst case — the DOM list stays long after a re-rank, unlike the client-filtered pages where the browser clamps scroll).

## Nits

- `test-utils.tsx` JSDoc: "the same double-invocation semantics **prod dev** uses" — garbled; StrictMode double-invocation is dev-build-only.
- `window.scrollTo({ top: 0, left: 0 })` for symmetry; horizontal offset is currently preserved (harmless today).
- The scroll fires at debounce-commit, before the new results land — right UX, but it also fires when the fetch errors. Worth a one-line comment acknowledging the trade-off.
- The two new tests duplicate the `try/finally { mockRestore() }` scaffolding; the file's `afterEach` could own it. Assertions are at-least-once (`toHaveBeenCalledWith`) rather than exactly-once — defensible given debounce timing on slow CI, but worth being a deliberate choice.
- Pre-existing, for a follow-up: `useDebouncedValue` is duplicated verbatim in `modules/discover/lib/` and `shared/hooks/`.

The fix is genuinely well-engineered — the ref pattern, the sticky-toolbar interaction, and the instant-scroll choice are all exactly right. It's only the harness rider that needs to either do what it says or come out. Happy to re-review quickly.

ritvik-jentic:
Thanks for the thorough review @Manuel-Jentic — you're right on the harness rider. I've dropped the StrictMode test change entirely (Option B) so this PR is just the scroll fix.

**What changed since `7a174f9`:**
- `test-utils.tsx`: removed the `<StrictMode>` wrapper from `renderWithProviders` and its docstring. As you showed, a wrapper-level `<StrictMode>` doesn't enable strict effect double-invocation in React 19 (it only engages at the `createRoot()` root, which RTL renders below), so it was a silent no-op shipping a false safety guarantee.
- `DiscoverPage.test.tsx`: removed the comment claiming the test exercises "dev-only double effect invocation" — without the wrapper it never did.

The Discover scroll fix itself is unchanged: the previous-value ref guard stands on its own correctness. Since real strict effects were never enabled, `OAuthPopupReturn.test.tsx` stays green (that's exactly the fallout we're avoiding pulling into this PR).

**Verified:** `DiscoverPage` + `OAuthPopupReturn` tests pass (29/29), lint clean on the changed files.

Ready for re-review 🙏

Manuel-Jentic:
Continuing this PR while Ritvik is away (same process as the #794/#791 batch).

**Since `496acb2`:**
- Rebased onto current main (`2ea0591`, clean — no conflicts) and verified the round-1 rework: `test-utils.tsx` has an **empty diff vs main**, and no StrictMode-coverage claim survives anywhere in the diff.
- Ran a 4-lens review board on the rebased branch (code quality, correctness/edge cases, product fit/UX, prior-review disposition) — **unanimous approve-with-nits, zero blockers**; the round-1 blocking item is confirmed fully resolved via option (b), and every claim in Ritvik's reply checked out true.
- Folded the remaining in-PR nits into `2213efe`: `left: 0` snap symmetry, the fires-at-commit/fires-on-error trade-off documented at the effect, and `afterEach`-owned spy restoration replacing the per-test `try/finally` scaffolding.
- Retitled `catalogue` → `catalog` (squash title consistency) and refreshed the body to the shipped diff (it still described the dropped harness change).
- Filed the follow-ups the two review rounds named: #894 (shared `useScrollResetOnChange` for Toolkits/Workspace parity), #895 (real StrictMode enablement via RTL `reactStrictMode` + `OAuthPopupReturn` fallout), #896 (`useDebouncedValue` dedup).

**Validation on `2213efe`:** `npm run lint` / `npm run build` clean; full browser-mode suite 104 files / 884 tests pass.

One heads-up for the squash body: the branch commit `ae7d2a0`'s "no-op StrictMode wrapper" wording is inaccurate (the wrapper was functional in dev-build browser mode; it was removed as review scope-trim, and it was this branch's own rider, not main code) — squash-merge hides the message, just don't carry the "no-op" claim into the squash body.

Issue #620, opened 2026-07-15
[ui] Can't rename an agent — display name stuck at "jentic-cli-default" though PATCH /agents/{id} exists


ash-jentic:
### What happened

My agent is named `jentic-cli-default` and there is no way in the UI to rename it. I'd like to call it something meaningful to me — e.g. "Claude Code agent" — so I can tell my agents apart at a glance. The Agents section lets me approve / deny / disable / archive, but not rename.

### This is a UI gap, not a missing capability

The control plane already supports it — `PATCH /agents/{agent_id}` ("Update Agent") exists in the API. So the backend can almost certainly rename an agent; the web UI simply exposes no control for it. (Worth confirming with eng which field carries the display name.)

Also unclear / worth deciding: can the **agent itself** rename itself? Today the agent token is read-only on `owner:agents:*` (no `agents:write`), so it cannot — rename is an operator action. That's probably correct, but the operator needs a way to do it, and right now there is none in the UI.

### Why it matters

The agent is a top-level object the user reads constantly (dashboard, events feed, toolkit bindings, access requests). A machine default like `jentic-cli-default` — and especially the raw `agnt_…` id — is exactly the ID-only, error-prone identity problem already raised in feedback#36 (internal tracker). Letting users name agents in human terms ("Claude Code agent", "finance bot") makes every downstream screen legible.

### Expected

An inline "Rename" / editable name on the agent's row or detail page in the Agents section, writing through `PATCH /agents/{agent_id}`.

### Environment
- Jentic One: live self-hosted instance (127.0.0.1:8000), 2026-07-09
- Surface: Web UI · Agents

---
*Migrated from the internal dogfooding tracker (feedback#53, originally filed 2026-07-10). Screenshots/attachments referenced in the original will be re-attached.*

Manuel-Jentic:
Delivered by the agents identity-console rebuild (#878): the agent detail page's Settings tab now carries an editable name/description form on the shared `IdentitySettingsCard` (PATCH /agents/{id}, dirty-field-only). The rename-dialog approach originally attached to PR #852 was dropped in favour of that rebuild; #852 continues as the friendly-API-names work (#631) only.

PR #852, opened 2026-07-28
feat: persist catalog identity (api_id) and title API surfaces from it

### LINKED ISSUE #631: [ui] APIs, credentials and toolkits are shown by machine tuple, not a friendly name
Across the app, APIs/credentials/toolkit rows display the raw identity tuple (e.g. `posthog-com/posthog-com-posthog-api`) instead of a human name like **PostHog**. It appears in the credential picker, the credential detail ("Edit credential") header and Name field, and toolkit bound-credential rows. Effects: hard to scan, hard to recognise what you're looking at, and the credential Name defaults to the tuple. Ask: show a friendly display name (and vendor logo where available), with the tuple demoted to a secondary/copyable detail. Found during a toolbox-first setup run (connecting many pre-imported APIs).

---
*Migrated from the internal dogfooding tracker (feedback#69, originally filed 2026-07-14). Screenshots/attachments referenced in the original will be re-attached.*

### LINKED ISSUE #910: Persist the catalog identity (api_id) at import and expose it on credential/binding/API DTOs
## Problem

Only the Discover surface can render a clean friendly API name (`Article Search`), because only the catalog DTO carries **`api_id`** (`nytimes.com/article_search`) — a token where the vendor and sub-API are separable. Every other surface (credential cards, pickers, workspace tiles, toolkit binding rows) receives the `vendor`/`name` tuple, where on real data the sub-API path is mangled into `api_name` with the vendor mirrored in (`posthog-com/posthog-com-posthog-api`, `posthog com posthog-api`, …).

PR #852 currently compensates in the UI with a ~300-line heuristic (`ui/src/shared/lib/api-display.ts`: vendor-prefix strip, mirrored-domain TLD peel, tuple peel) plus ~400 lines of edge-case pins. Review found three false-positive classes in two days (`Com Posthog Api`, `App Store Connect` truncation, vendor-doubled subtitles) — the signature of a parser fighting its data. The root cause is a data-model gap, not a display problem.

## Why it isn't a one-line passthrough

The `api_id` is genuinely **lost at the import boundary** today:

- It's parsed out of the external catalog manifest (`apis.json` include URLs) in `registry/services/catalog/manifest_builder.py` and never persisted.
- The registry `Api` row (`registry/core/schema/apis.py`) stores only `vendor`/`name`/`version`.
- Credentials copy `api_vendor`/`api_name` strings (`control/core/schema/credentials.py`), and the toolkit binding DTO re-exposes those.

## Proposal

1. **Persist the catalog identity on the `Api` row at import time** — an `api_id` (or `catalog_slug`) column, nullable for manually-imported/draft APIs that never had one.
2. **Carry it through the read DTOs**: `ApiResponse`, credential responses, `ToolkitCredentialBindingResponse` (and anything else exposing the vendor/name tuple as identity).
3. **Shrink the UI helper** to Discover's rule: split `api_id` on `/`, title-case the tail; fall back to humanising `vendor` when `api_id` is absent. The strip/peel machinery in `api-display.ts` gets deleted.

Alternative considered: a computed `display_title` on the DTOs instead of the raw `api_id`. Exposing `api_id` is preferred — it's the stable identity (useful beyond display: dedup, deep-links to Discover) and keeps presentation in the UI.

## Notes

- Backfill: existing rows can be back-filled where `(vendor, name)` unambiguously matches a current catalog entry; the rest stay `NULL` and use the vendor fallback.
- Blocks the thin redo of #852 (parked pending this).

ritvik-jentic:
## Summary

Rebuilt per review: instead of reverse-engineering friendly names from the slugified `vendor`/`name` tuple, the catalog identity slug (`api_id`, e.g. `nytimes.com/article_search`) is now **persisted end-to-end** and the UI titles from it directly. Closes #631, closes #910.

### Backend (#910)

- **Registry** — nullable `apis.catalog_api_id`, carried verbatim through the existing catalog-import path (the slug already reached the importer as `api_name` but was destroyed by slugification). Catalog imports set/backfill it; slugless re-imports (manual, overlay materialization) never clear it. Exposed on `ApiResponse`.
- **Control** — nullable `credentials.catalog_api_id`, accepted on the create request's `api` reference and copied verbatim (deliberately outside scope canonicalisation — it's display provenance, never resolution identity). Exposed on credential responses and toolkit-binding responses (which read straight off the ORM row).
- Two migrations (registry + control), OpenAPI + generated UI client refreshed.

### UI (#631)

- `api-display.ts` rule is now: explicit `display_name` → title from the persisted slug **exactly like Discover** (`Article Search`; bare domain `stripe.com` verbatim) → legacy tuple humanisation for pre-column rows.
- The mirrored-domain **TLD-peel heuristics are deleted** — the guesswork existed only because the slug was lost. A minimal exact-prefix vendor strip remains for legacy rows.
- The create-credential flow stores the picked slug on the credential (picker → `SelectedApi.apiId` → form → `api.catalog_api_id`), so new credentials render friendly everywhere immediately.
- Machine-identity subtitles (`apiIdentityTuple`) show the slug verbatim when recorded.
- Catalog picker rows now title identically to Discover (no more `Github` here vs `github.com` there).

### Notes

- Old rows are NULL and fall back to the previous tuple rendering; they backfill on their next catalog re-import (APIs) or naturally as credentials are recreated.
- The toolkit clear-description fix and actor-directory invalidation from the earlier revision of this PR are retained.

## Testing

- Backend: `pytest tests/unit tests/arch` (2498 passed), ruff + mypy clean. New unit tests pin the upsert backfill/preserve semantics and verbatim slug carriage through `load_specification`; web tests cover the credential round-trip (`catalog_api_id` on create echo/read/list) and the import-source payload.
- UI: full vitest suite (1022 passed), tsc + eslint + prettier clean. Tests updated to the Discover-consistent titling; new tests pin slug-first behaviour in all three helpers and the create-body carriage.


Manuel-Jentic (changes requested):
Multi-lens review (code quality, adversarial heuristic testing, architecture/conventions, UX/a11y, data-flow correctness) at head `f59b4ad`. Every load-bearing finding below was verified directly against the branch and the backend code, not assumed.

## Verified clean

- Module boundaries hold: all cross-module reuse goes through `shared/*`; Discover's `adapters.ts` now *consumes* `titleFromApiId` from the shared module — no drifting private copy. ✔
- PATCH-returns-row contract verified against both backend routers; `setQueryData` shapes match on both sides (`agentToEntity` / raw `ToolkitResponse`). ✔
- Dialog hardening is genuinely good: open-before-load seeding, seeded-snapshot diffing (immune to background refetches clobbering a draft), pending-gated dismissal on all three paths (Cancel / Escape / backdrop), inline server errors with `mutation.reset()` on re-open. ✔
- Diff-vs-seeded patch logic handles whitespace correctly; unedited saves are a true no-op. ✔
- Dead-code removal complete (`truncateTitle`, `ToolkitDetailBody` sheet-Edit path incl. all unused imports). No security issues. ✔

## Blocking

**1. Clearing a toolkit description silently no-ops.**
The dialog normalizes a cleared description to `null` (`EditNameDescriptionDialog.tsx` patch build) and the toolkit host forwards it verbatim. The agent path is fine (`PATCH /agents/{id}` uses `model_dump(exclude_unset=True)`, so explicit `null` clears). But the toolkit repo skips `None` fields (`src/jentic_one/control/repos/toolkit_repo.py:113` — `if description is not None:`), so the server returns 200 with the **old** description, the dialog closes as "success", and `setQueryData` re-seeds the stale text. The user's deletion is silently discarded. The toolkit tests don't catch it because their PATCH mocks ignore `description`. Fix: map `description: null → ''` in the toolkit host (the repo honours `''`), and add a clear-description test that asserts the round-tripped value.

**2. The branch pre-dates #797 and its `useUpdateToolkit` rewrite drops the `toolkitNameRoot` invalidation.**
Merge-base is `a9c5307`, before #797 (`ccd9441`) landed. On `main`, `useUpdateToolkit.onSuccess` invalidates `[...sharedQueryKeys.toolkitNameRoot, toolkitId]` so the Agents module's per-bound-row name cache (`useToolkitName`, 5-min staleTime) refreshes on rename. This branch's rewritten `onSuccess` doesn't — `toolkitNameRoot` doesn't exist anywhere on the branch (`rg toolkitNameRoot ui/src` → empty). Rename is *exactly* the operation this PR adds a UI for: as merged, renaming a toolkit leaves stale labels on every agent detail page's "Bound toolkits" card for up to 5 minutes. Both sides rewrote the same function, so this will conflict — please rebase onto main and deliberately re-add the `toolkitNameRoot` invalidation alongside your new `setQueryData`/`dashboardRoot` logic, then re-verify.

## Should-fix

1. **Agent rename never invalidates the actor directory.** `useUpdateAgent` invalidates lists/`agentsRoot`/`dashboardRoot` but not `actorDirectoryKey` (`['actor-directory']`, 5-min staleTime, no focus refetch). Every `ActorLabel` — monitor, toolkit audit, access requests, and the "Registered by"/"Approved by" grid on the very page being renamed — shows the old name for up to 5 minutes. Add the invalidation (exported from `@/shared/hooks`, no boundary issue).
2. **Filters don't match the friendly names they display.** `ApiPicker`'s haystack is raw `display_name/description/vendor/name/host`; `CredentialPicker` filters on raw `name/vendor/apiName/provider`. A user reading "Posthog Api" who types "posthog api" gets zero results. Add the derived display label to each haystack — otherwise the feature undermines its own goal.
3. **Heuristic edge cases** (adversarially traced through `api-display.ts`):
   - A test pins `'Com Posthog Api'` as *expected* for `{vendor:'posthog', name:'posthog com posthog-api'}` — the strip orphans the `com` token; that's the bug class the strip exists to prevent, shifted one token right.
   - `name === vendor` exactly (realistic mirror shape) escapes both the strip and the generic-name gate: renders `Posthog Com`, while `name:''` renders `Posthog.Com` — same API, two casings. Treat a fully-consumed prefix as generic → vendor fallback.
   - Genuinely dotted multi-label domains lose all dots: `bbc.co.uk` → `Bbc Co Uk` (the `uk` tail fails the TLD allowlist, and interior dots always split to spaces). If input `hadDot`, preserve its dots wholesale.
   - Vendor-only APIs render differently across surfaces: Discover `posthog.com` (raw), picker `Posthog.Com`, workspace tile `Posthog Api`. The umbrella case (`Article Search`) does line up — worth stating in the PR body that parity is scoped to umbrella sub-APIs.
4. **`EditNameDescriptionDialog` contradicts `dialog-state-lifecycle.md` as written** — the rule mandates "reset on successful commit, persist between dismissals" for plain edit forms and lists clear-on-close/re-seed-per-open as anti-patterns; the dialog deliberately does both. Since this is a shared `shared/ui` primitive that hard-codes the policy for every future consumer, either follow the rule or amend it in the rules repo first.
5. **Toolkit description cap mismatch**: shared `DESCRIPTION_MAX_LENGTH = 1024` matches the agent column (`String(1024)`) but the toolkit column is `String(1000)` and `ToolkitUpdateRequest.description` carries no `max_length` — a 1001–1024-char description passes client + request validation, then overflows on length-enforcing databases. Parameterise per entity (or align the backend in a follow-up).
6. **Validation fires on the unseeded draft**: with the dialog open before the entity loads, the Name field shows "Name can't be empty." before the user has touched anything. Gate error display on seeding/touched.
7. `api-display.ts` and its test reference `docs/plans/issue-631-friendly-api-names.md` as the decision log — the file doesn't exist on the branch.
8. Ephemeral review-round numbering (`#1`…`#15`) is baked into comments across ~8 files; future readers can't resolve them (same feedback as #797 rev-1 — those were scrubbed there).

## Nits

- `validators.ts` hardcodes "255"/"1024" in error copy instead of interpolating the constants; `AgentCreateSheet`/`ServiceAccountCreateSheet` still carry inline `maxLength` for the same caps.
- `useUpdateAgent`'s variables type allows `name: null`, which the dialog can never emit and the backend 422s/mis-maps; tighten to `name?: string`.
- `entityMissing` is dead on `AgentDetailPage` (the page early-returns before the dialog can see a missing entity) — and behavior diverges from the toolkit page where it's live.
- Edit pill: consider routing through `PageHeader`'s existing `actions` slot per `page-scaffold.md` ("title + subtitle + actions"), or bless the new prop in the rule body.
- `PageHeader.test.tsx` asserts Tailwind class names (`self-center`) — brittle implementation pinning.
- The separator regex in `api-display.ts:137` is written twice; the JSDoc at :121 lists fewer separators than the code accepts.
- Missing heuristic tests: `name === vendor` mirror, dotted country TLDs, strip-leaves-TLD-only remainder (`vendor:'posthog', name:'posthog-com'` → renders `Com`), idempotency pins, uppercase input, cross-helper consistency pin on one identity.

The engineering quality here is high — the dialog lifecycle, patch semantics, and test discipline are all above the bar, and the PR body's honest "why not api_id everywhere" analysis is exactly the right kind of writeup. The two blockers are a silent data-loss bug and a cross-PR cache regression that the merge conflict will force you to resolve anyway; the filter-haystack gap is the one should-fix I'd weigh nearly as heavily since it undercuts the feature's stated goal. Happy to re-review quickly.

Manuel-Jentic:
Re-review round addressing the CHANGES_REQUESTED review at `f59b4ad`, plus a fresh merge of `main` (through #905) and a two-lens review board on the result. The scope changed materially since the review: the rename-dialog half (#620/#635) was superseded by `main`'s console rebuilds (#878 + toolkit tabbed console, shared `IdentitySettingsCard`) and dropped — the PR is now purely the #631 friendly-names work (~26 files, +1000/−76). Body and title updated to match.

**Blocking**

1. *Toolkit description clear silently no-ops* — *fixed.* The Settings form now wires a cleared description as `''` (the repo honours `''`, ignores `null`), with a regression test whose MSW mock mirrors the real repo's `null`-skip so it fails under the old `|| null` code. The agent console keeps `main`'s `null` semantics — verified its backend honours explicit `null` (`model_dump(exclude_unset=True)` → repo `setattr` verbatim), so the two paths genuinely differ and the comments say why.
2. *Missing `toolkitNameRoot` invalidation (pre-#797 rebase)* — *fixed.* Branch is merged up to current `main`; `useUpdateToolkit.onSuccess` invalidates `[...sharedQueryKeys.toolkitNameRoot, toolkitId]` alongside the list invalidation, with a comment pinning why `useSetToolkitActive` deliberately doesn't touch that root.

**Should-fix**

1. *Actor directory staleness on agent rename* — *fixed*: `useUpdateAgent` invalidates `sharedQueryKeys.actorDirectoryRoot`.
2. *Filters don't match displayed names* — *fixed*: both pickers include the derived friendly label in their haystacks (tested by typing the on-screen name).
3. *Heuristic edge cases* — all four addressed: the orphaned-TLD strip now peels the vendor's mirrored domain form (`posthog com posthog-api` → `Posthog Api`), gated on the vendor not already carrying its TLD so `apple-com/…-app-store-connect` keeps `App Store Connect`; `name === vendor` falls through to one vendor rendering; genuinely dotted domains keep their dots (`bbc.co.uk` → `Bbc.Co.Uk`); the umbrella-scoped-parity caveat is now stated in the PR body.
4./5./6. *Dialog-lifecycle rule, description-cap mismatch, unseeded-draft validation* — *moot*: `EditNameDescriptionDialog` and `validators.ts` are gone with the rename half; identity editing rides `main`'s `IdentitySettingsCard`.
7./8. *Phantom plan-doc reference, ephemeral review ordinals* — *scrubbed* (including a second pass that caught `(#6)`/`(#14)` leftovers in test titles).

**Nits** — the `AgentPatch` `name?: string` tightening and dead `entityMissing` went away with the dropped half; the separator regex is now defined once (`PREFIX_SEPARATORS`); the missing heuristic pins (TLD-only remainder, uppercase, idempotency, cross-helper agreement on one identity) are in.

**Review board on the new state** found and fixed three more: catalog picker rows keep the full `api_id` as the mono subtitle (two entries that humanise identically stayed distinguishable pre-PR; now they do again), `apiRefDisplayName` peels a tuple-shaped `name` like the binding-row helper (same backing column), and raw-tuple subtitles go through a shared `apiIdentityTuple` so a tuple-shaped `api_name` can't render the vendor twice.

`tsc`/`eslint` clean, `vitest run` 1018/1018.

Manuel-Jentic:
**Parking this PR** after a scope discussion: the reason `api-display.ts` needs ~300 lines of heuristics (vendor-prefix strip, mirrored-domain TLD peel, tuple peel) plus ~400 lines of edge-case pins is that the UI is reverse-engineering an identity the backend already had and dropped at the import boundary — the catalog `api_id` is parsed from the manifest but never persisted, so credential/binding/API DTOs only carry the mangled `vendor`/`name` tuple. Three false-positive classes surfaced in two review rounds; that's the approach fighting its data.

Filed #910 to fix the root cause: persist `api_id` on the `Api` row at import, expose it on the read DTOs, and let every surface use Discover's 10-line rule (with a plain vendor humanise as the `NULL` fallback). Once that lands, this PR gets redone thin — the wiring, filter-haystack, subtitle-convention, and test work here carries over; the strip/peel machinery gets deleted.

Independently useful pieces that shouldn't wait for #910 and can be cherry-picked into small PRs if wanted: the toolkit clear-description fix (+ regression test) and the actor-directory invalidation on agent rename.

Manuel-Jentic:
Un-parking: rebuilt per the discussion — #910 is now implemented *in this PR* (persisted `catalog_api_id` on `apis` + `credentials`, exposed on API/credential/binding DTOs), and the UI heuristics that motivated parking are deleted in favour of titling from the persisted slug. See the updated PR description for the full breakdown.

PR #853, opened 2026-07-28
feat(cli): run local coding agents as a dedicated unix user

DavidAtJentic:
## Summary

Closes the local credential-boundary gap for locally-run coding agents: today a same-user agent can read the operator's `~/.jentic` keyset, db files, browser session, and provider credentials. This adds a dedicated-Unix-user isolation model plus a per-session process-confinement layer, and the CLI commands to drive its full lifecycle — provision, launch, and tear down — folded into onboarding so the agent runs as its own account behind a real OS boundary.

## Changes

- **`jentic run <agent>`** — launches a coding agent as its own Unix user, with per-directory access granted via a traverse-walk + rwx-leaf ACL model (no home-wide sweep), plus provider-aware LLM config seeding. Sessions are wrapped in `sandbox-exec` (macOS) / `bwrap` (Linux), which denies every human-home root and re-opens only the agent home and granted dirs; the launch **errors closed** when confinement is unavailable. `--grant` / `--revoke` / `--list-grants` manage access standalone.
- **Agent-user creation folded into `jentic bootstrap` / `jenticctl wizard`** — sudo-last opt-in that creates the account, seeds config/provider files, writes the agent's identity into its own `~/.jentic`, optionally installs a narrowly-scoped passwordless-launch sudoers rule, and offers to start a session.
- **`jentic reset [profile]`** — inverse teardown: scope follows the argument (named profile vs. full clean slate), removes ACLs, sudoers, account, agent identity, and config; agent home preserved-and-re-owned by default, deletable on explicit confirmation.
- Operator-editable inputs (account name, home, grant paths) are validated at the source and re-checked before every privileged command; grant paths are canonicalized (symlink-resolved) before classification; banned paths (`~/.ssh`, `~/.jentic`, system trees, home roots) are un-grantable with no escape hatch.
- New `config`/`profile` support (shared agent-account entry, atomic locked config writes, profile delete, `profile view` access map) and a consolidated design-doc set under `docs/security/local-agent/`.

## Testing

- `go build`, `go vet`, `go test ./...` pass (all 22 packages), `golangci-lint` clean; `ui/public/cli-reference.json` regenerated and in sync.
- Reviewer-driven live boundary testing on macOS 15.6 against the real privileged command builders: account create → grant → probe boundary (denied: operator home, `~/.ssh`, `~/.jentic`, sibling traversal) → revoke → full teardown.
- All review merge-blockers verified closed at head (see review thread); the two remaining review residuals (fail-open userns probe, agent-writable `~/.local/bin` exec route) are fixed in the final commits.

## Checklist


Manuel-Jentic (changes requested):
## Review

**Verification done locally**

- `go build`, `go vet`, `go test ./...` all pass on the branch
- `ui/public/cli-reference.json` is in sync with the regenerated `cmd/clidocs` output
- Commits follow Conventional Commits with scopes and carry DCO sign-offs
- `run`/`reset` are registered additively on the `jentic` (API) tree only, with a test guarding they don't leak onto `jenticctl`

The overall design is strong — sudo-last posture, survey→confirm→act for `reset`, the "reference not copy" identity model, and tests that pin the load-bearing ACL permission strings. The design doc is excellent. The findings below are mostly consistent with the PR's own "not yet fully hardened" caveat.

### 1. `DangerReason` only exact-matches sensitive dirs — descendants bypass the gate (high)

`cli/internal/localagent/localagent.go` (~L833): the sensitive-dir check is `abs == filepath.Join(home, d)`, so `jentic run claude ~/.ssh/foo` (or `--grant ~/.jentic/profiles`) is **not** flagged dangerous. Worse, granting a descendant makes Layer 1 stamp an **execute traverse ACE on `~/.ssh` itself** — precisely the ancestor the model is supposed to keep closed. Same gap for paths *inside* another user's home (`/Users/bob/x` passes; only the home root trips `isDirectChild`). These should be prefix/subtree checks (like `IsUnderHome`), not equality.

### 2. Unvalidated `agentUser` interpolated into root shell scripts (medium)

`CopyBinaryCmd` / `CopyConfigCmd` splice `agentUser` **unquoted** into a `sudo sh -c` script via `$(eval echo ~<agentUser>)`. `agentUser` comes from `--agent-user` / config, so it's the operator injecting into their own sudo (not a privilege escalation), but it's a fragile pattern — and the unquoted `dest` also breaks on homes with spaces. Suggest validating the account name once at entry (POSIX username charset, e.g. `^[a-z_][a-z0-9._-]*$` in `resolveAgentUser` and the bootstrap form) and resolving the home in Go (`user.Lookup`) instead of `eval`.

### 3. `AgentACLPresent` substring match can false-positive and abort reset (medium)

`strings.Contains(out, "user:"+agentUser)` — `user:alice` matches an ACE for `alice-local-agent`. A false positive makes `buildResetSteps` emit a revoke for an ACE that isn't there; macOS `chmod -a` errors on a missing entry, and revoke steps are **not** `BestEffort`, so the teardown aborts. Anchor the match (`user:<name> ` on macOS, `user:<name>:` on Linux). Related: `RemoveSudoersCmd`'s `grep -v <name>` treats the name as a regex *and* substring — resetting `alice-agent` would also drop lines for `alice-agent2`.

### 4. Mid-failure in account setup leaves an orphaned account invisible to `reset` (medium)

In `setupAgentUser`, `recordAgentAccount` runs only **after** `createAgentAccount` fully succeeds. If account creation succeeds but a later step in the same function fails (home reclaim grant, config seeding), no `local_agents` entry is written — and `jentic reset <agent>` then refuses with "no configured local agent", leaving a Unix account jentic created but won't tear down. Recording the entry immediately after the account exists (before seeding) would keep reset's "config records what still needs cleaning" invariant.

### 5. `AccountCreated` is written but never read (low)

Its doc comment says `jentic run` and teardown paths key off it, but nothing in the codebase reads the field. Either wire it up or soften the comment.

### 6. Smaller nits

- `AncestorChain(home, home)` returns `["/", "/Users"]` — granting the home root itself (possible after the typed dangerous confirm) would probe/grant traverse on `/` and `/Users`. Harmless in practice, but returning nil for `leaf == home` seems right.
- The hand-rolled bubble sort "to avoid importing sort" appears twice (`localagent.Known`, `resetTargets`) — `slices.Sort` is stdlib.
- `offerAgentSession` failure at the very end returns an error from `bootstrapE` after everything already succeeded — probably better reported as a warning.
- Per the agent-commands rule's spirit, `--list-grants` could emit JSON off-TTY at some point; fine to defer since `run` is operator-facing.

### Verdict

Approve-with-changes for a WIP: #1 is worth fixing before merge since it undercuts the exact boundary the PR exists to build; #2–#4 are solid hardening follow-ups that fit the PR's own "more work to fully lock down" note, though #3/#4 affect `reset`'s reliability, which this PR presents as complete.

Manuel-Jentic:
## Review addendum — additional findings

Follow-up to my earlier review; these came out of a second pass and are not covered there.

### 1. Symlink bypass of the danger check and the grant itself (high)

`resolveWorkingDir` / `runGrantDir` (`cli/internal/cmd/run.go` ~L304, ~L530) only do `filepath.Abs` + `Clean` before calling `DangerReason` and granting — there is no `filepath.EvalSymlinks`. A symlink at a benign-looking path pointing at `~/.ssh` (or `~` itself) classifies as safe, and `chmod -R +a` / `setfacl -R` on that path applies the rwx ACL to the **target**. Since the agent can create symlinks inside its granted workspaces, this is a plausible re-entry into exactly the boundary the PR closes: the operator re-runs `jentic run claude <workspace subdir>` on a planted link and grants the agent its own way back in.

Suggest resolving with `EvalSymlinks` before classifying, granting, and recording in `granted_dirs` (and double-checking how `chmod -R` / `setfacl -R` treat symlinks *inside* the granted subtree on both platforms).

### 2. Named `jentic reset <agent>` can delete a Unix account shared by other agents (medium)

The default account is per-**operator** (`<operator>-local-agent`), not per-agent, so two `local_agents` entries will normally share one Unix account. `execAgentReset` unconditionally deletes `plan.user` (and strips its sudoers lines) without checking whether another configured entry still references the same user. Today the registry only has `claude`, but the config schema and `reset_test.go` already model multiple agents. Suggest skipping account/sudoers deletion when another entry references the user — and saying so in the printed plan.

### 3. `--force --delete-home` can leave the agent identity behind (low)

`buildResetSteps` skips the separate identity-removal step when `deleteHome` is true, relying on the home `rm -rf` — but that step is `BestEffort`. If the rm partially fails (the macOS SIP/TCC case it exists for), the documented "identity gone" guarantee silently doesn't hold, and a later re-bootstrap over the surviving home can find the stale registration. Making the identity step unconditional is harmless (the code comment already says so) and closes the hole.

### 4. Doc/code mismatch on account creation in `jentic run`

The design doc (step 1 of the `jentic run` section) says "if the account doesn't exist, **offer to create it** — the same account step bootstrap/wizard run"; the code errors and points at `jentic bootstrap` instead (`run.go` ~L109). One of the two should change.

### 5. Bootstrap wiring questions

- The agent-user step is gated on `!opts.skipSkill`, so `bootstrap --skip-skill` silently never offers isolation. If intentional, worth a comment/doc note; a dedicated gate seems more correct.
- After an isolated bootstrap, `SetDefaultProfile(idPaths, …)` writes the default profile into the **agent's** `~/.jentic` only — the operator's own config keeps no profile, so the operator's bare `jentic` commands fall back to the built-in default. Deliberate per the reference model, but worth confirming that's the intended operator experience.
- `config.yaml` previously had a single writer (`jentic profile use`); this PR adds `run`/`reset`/`bootstrap` as writers via the local-agent entries. Fine as a design change, but the state-directory contract/conventions should be updated to match in the same breath.

### 6. Residual worth adding to the docs' accepted-gaps list

The launch attaches the operator's TTY directly to the agent process (`LaunchCmd` wires `os.Stdin/out/err`). Same-TTY attacks — TIOCSTI keystroke injection on platforms/kernels where it's still enabled, and terminal escape-sequence tricks — mean the OS-user boundary isn't complete without a pty intermediary. Consistent with the PR's "not yet fully hardened" caveat; I'd just list it explicitly alongside the other accepted residuals in `docs/security/analysis/local-agent-isolation.md`.

Manuel-Jentic:
## Live end-to-end testing — two confirmed boundary leaks

I rebuilt both binaries from this branch's head and drove the **real** `localagent` privileged command builders (`CreateAccountCmds`, `TraverseGrantCmd`, `LeafGrantCmd`, `LeafRevokeCmd`, the full `reset` teardown set) through a complete lifecycle on macOS 15.6 — create account → `chmod 700 ~` → grant a project dir → probe the boundary → revoke → tear everything down. So the `sudo` operations below are exactly the ones the CLI issues, not a paraphrase.

**What worked as designed:** account creation, `chmod 700` on the operator home, the 5-level traverse chain + recursive rwx leaf (`DirAccess` flipped false→true), agent read **and** write (create+rename+delete — the macOS `add_subdirectory`/`delete_child` fix genuinely works), revoke (true→false), and a clean idempotent teardown (post-revoke leaf-revoke printed `No ACL present` and continued). Teardown left zero residual ACEs on the whole ancestor chain, deleted the account/home/sudoers, and restored the home perms. Nice.

Two boundary problems came out of the live run:

### A. Traverse-walk exposes world-readable `~/.jentic` — the credential store itself (high)

After granting a project dir *inside* `~/Desktop/…`, Layer 1 opened execute-traverse on every ancestor up to and including `~`. I then probed as the agent user:

```
agent can list ~/.ssh (expect FALSE):           false   ok (0700 saves it)
agent can read operator home (expect FALSE):    false   ok (700 home)
agent can traverse to ~/.jentic (expect FALSE): TRUE    LEAK
```

`~/.jentic` is created mode **0755**, so once `~` is traversable the agent can `ls ~/.jentic` (and descend into anything world-readable within) despite never being granted it. This is precisely the "world-readable dir reachable via a traversed ancestor" residual the design doc lists as accepted — **but the dir in question is the operator's own jentic keyset/tokens/registration**, i.e. the exact credential store this whole feature exists to wall off. I'd treat this one as not-acceptable-as-is rather than a documented residual. Cheap mitigations: `chmod 700 ~/.jentic` at creation, and/or the doc's own "strip world bits on the traversed ancestors" narrow-traverse option.

### B. Symlink bypass — confirmed live, not just at the classify layer (high)

A symlink planted inside the agent's *own* granted workspace, pointing at `~/.ssh`:

```
DangerReason(.../proj/escape -> ~/.ssh) = ""            (classified SAFE)
$ sudo chmod -R +a# 0 user:<agent> allow <macLeafACE> .../proj/escape
after granting the symlink, agent can list ~/.ssh THROUGH it: TRUE
```

`resolveWorkingDir`/`runGrantDir` classify and grant on the `Abs`+`Clean` path with no `EvalSymlinks`, and the real `LeafGrantCmd` (`chmod -R +a`) follows the link, handing the agent recursive rwx into `~/.ssh`. Since the agent can create symlinks in its workspace, this is a realistic re-entry into the boundary (operator re-runs `jentic run <workspace-subdir>` over a planted link). This is the same issue as Addendum #1, now reproduced with the PR's actual grant command rather than inferred. Fix: `EvalSymlinks` before classifying, granting, and recording; and confirm how `chmod -R`/`setfacl -R` treat symlinks *within* the granted subtree on both platforms.

### Note

Testing had to go through the raw `localagent` builders because `jentic bootstrap` (the only account-creation entry point) is a fully interactive TUI that also registers against the control plane — I couldn't drive real account creation via the shipped UX headlessly. Everything else (the pure `run`/`reset` logic, plan previews, `--force` guards, drift detection) I exercised through the built binaries directly and it all behaved correctly.

Manuel-Jentic:
## Teardown-reliability edge cases — three confirmed (no sudo needed)

Follow-up probing the multi-agent / name-collision paths in `reset`. These are logic bugs I could prove against the real functions without creating accounts or touching ACLs, so they're cheap to verify and fix.

### 1. `AgentACLPresent` substring match false-positives across name prefixes (confirms review comment #3)

`localagent.go:408` is `strings.Contains(out, "user:"+agentUser)`. I stamped a real ACE `user:manuel` on a temp dir and asked `AgentACLPresent` about `manue` (a prefix):

```
real ACE owner:           user:manuel
probed (prefix) agent:    user:manue
AgentACLPresent(prefix) = true      <-- want false
```

Impact in `reset`: `surveyReset` marks such an ACL `present`, so `buildResetSteps` emits a revoke for an ACE that isn't actually the agent's. macOS `chmod -a` errors on a missing entry, and the leaf/traverse revoke steps are **not** `BestEffort`, so `execAgentReset` aborts the whole teardown. Anchor the needle: `user:<name> ` (macOS `ls -lde`) / `user:<name>:` (Linux `getfacl`).

### 2. `reset <agent>` deletes a Unix account a second configured agent still shares (confirms Addendum #2)

The default account is per-**operator** (`<operator>-local-agent`), so two `local_agents` entries normally map to one Unix user. Surveying a teardown of only `claude` while `cursor` shares the account:

```
two agents share Unix user: op-local-agent
tearing down ONLY 'claude' produces:
  - remove sudoers drop-in (grep -v op-local-agent): sudo sh -c ... grep -v 'op-local-agent' ...
  - delete the Unix account op-local-agent: sudo dscl . -delete /Users/op-local-agent
```

`buildResetSteps`/`execAgentReset` target `plan.user` with no lookup against other entries, so `reset claude` deletes the account (and sudoers lines) out from under the still-configured `cursor`. Guard: skip account + sudoers removal when another `local_agents` entry references the same user, and say so in the plan.

### 3. `RemoveSudoersCmd` over-matches on substring (related to #2)

`RemoveSudoersCmd` filters with `grep -v <name>` (unanchored, and the name is a regex). Local demo with three drop-in lines, tearing down only `x-agent`:

```
before:                          after `grep -v 'x-agent'`:
manuel ALL=(x-agent)  ...        manuel ALL=(y-agent) ...
manuel ALL=(x-agent2) ...
manuel ALL=(y-agent)  ...
```

`x-agent2`'s line is collaterally removed. Anchor the match to the exact `(<name>)` run-as spec (and treat the name as literal, not a regex).

---

**Testing status summary.** Between the earlier live e2e run and these, I've now exercised: account create + `chmod 700` home, traverse+leaf grant, agent read/write, the `~/.jentic` traverse exposure, the live symlink bypass, revoke, full teardown + home restore, all `run`/`reset` pure logic + `--force` guards + plan drift detection, and these three teardown edge cases. The one path I did **not** drive is `jentic bootstrap`'s interactive TUI + control-plane registration end-to-end — that's largely orchestration over primitives already covered here, so I'd rate its residual risk low.

Manuel-Jentic:
Did a fair bit of hands-on testing on this over the morning (macOS 15.6) — happy path all works nicely, the ACL write-fix in particular. My two real worries are the boundary leaks: the symlink one and the `~/.jentic` traverse exposure both let the agent back at operator credentials, so I'd want those closed before merge rather than filed as accepted residuals. The teardown edge cases (shared account, substring matches) are lower stakes but worth a pass since reset is presented as complete.

Didn't drive the `bootstrap` TUI + registration end-to-end — it's mostly orchestration over primitives I already exercised, so I'm comfortable leaving that as-is unless you'd like me to sit down and click through it. Nice work overall; the design doc made this easy to review.

Manuel-Jentic:
## Second-pass review (against the 74-commit head) + a correction to my earlier report

Re-reviewed after the confinement refactor. Big picture: the new `sandbox-exec`/`bwrap` layer, the error-closed launch, and turning banned paths un-grantable materially strengthen this. Rechecked my earlier findings and re-tested the boundary live. Two things to correct/add.

### Correction — the symlink case is NOT a sandbox bypass (I was wrong earlier)

In my earlier addendum + live-testing comment I claimed a symlink planted in the agent's workspace pointing at `~/.ssh`, once granted, would expose `~/.ssh` **through the session**. I re-tested that against real `sandbox-exec` on macOS 15.6 and that claim is **wrong**: Seatbelt evaluates the *resolved real path*, so `(allow file* (subpath <link>))` does not follow the link into the denied `/Users` subtree. Reading `~/.ssh` through a granted symlink was **denied (`Operation not permitted`)**, and `~/.ssh` stayed denied even when the granted symlink pointed at a re-allowed sibling:

```
# profile denies /Users, re-allows only <workspace>/work/escape (a symlink -> secret-target sibling)
sandbox-exec ... cat <workspace>/work/escape/creds.txt   -> Operation not permitted
sandbox-exec ... ls  ~/.ssh                               -> denied
# control: symlink -> /tmp (outside denied roots) reads fine, as expected
```

So the residual is narrower than I said: the **ACL grant** still follows the symlink at the filesystem layer (`chmod -R` on a link stamps the target), which only matters on a **non-confined** path — i.e. the new `runSameUser` fallback, or any future code path that reaches a launch without `ConfinementAvailable`. That makes it defense-in-depth (resolve with `EvalSymlinks` before classifying/granting/recording so the persisted grant and the sandbox re-allow both refer to the real path), not the live boundary break I originally reported. Apologies for the overstatement.

Note `Classify` still runs on `filepath.Abs`+`Clean` with no `EvalSymlinks` (`run.go` resolveWorkingDir/runGrantDir), so the grant of a symlinked path is still accepted — worth the `EvalSymlinks` for the fallback path and for grant hygiene.

### New — committed `cli-reference.json` is stale; CI (`make check`) will fail

`TestCommittedCLIReferenceUpToDate` fails on this head: the checked-in `ui/public/cli-reference.json` is out of date with the command tree despite the "regenerate cli-reference.json" commit. Missing/stale entries: `jentic profile view`, the `run` group retitle (`Find and run operations` -> `Local agent client`) + the `-- <agent-args>` synopsis/help, the `run --profile` flag, and the rewritten `reset [profile]` help/example. It is the ONLY failing test in the suite (everything else, including the new `confine_test.go`, passes). Fix: `make cli-reference` and commit.

### Rechecked findings

- `DangerReason` descendant bypass — **fixed** (subtree/prefix match now).
- `~/.jentic` traverse exposure — **mitigated** by the sandbox (verified: `~` denied regardless of grants).
- reset abort when a leaf is also a traverse ancestor — **fixed** (and revoke steps are best-effort now).
- shared-account deletion (my Addendum #2) — **obsolete**, single shared account removes the premise.
- `AgentACLPresent` uses unanchored `strings.Contains(out, "user:"+agentUser)` — **still present** (lower impact now that revoke is best-effort, but a prefix name still mis-detects). Anchor to `user:<name> ` / `user:<name>:`.
- `RemoveSudoersCmd` filters with unanchored `grep -v <name>` (name also treated as a regex) — **still present**; anchor to the exact `(<name>)` run-as spec, literal.

Still digging into the `--`-forwarding parser, `identity_target.go`, `profile view`, and the Linux `bwrap` path — will follow up.

Manuel-Jentic:
## Full multi-pass review (8 parallel reviewers, one per slice) — findings against head `46eafbb`

Ran a fresh full-PR review, one reviewer per area (confine / localagent core / run / reset / bootstrap+agentuser / profile+identity+config / cli-wiring / docs). Findings below are deduped and each was verified against the code at head; where I could reproduce an issue mechanically I say so. **First, a retraction.**

### Retraction — the stale `cli-reference.json` finding is WRONG; it's fixed at head

In my previous comment I said `TestCommittedCLIReferenceUpToDate` fails because `ui/public/cli-reference.json` is stale. That was against an older extraction. At head `46eafbb` the committed JSON already contains `profile view`, the `run` retitle + `-- <agent-args>` synopsis, the `run --profile` flag, `reset [profile]`, and the correct doc path — regenerated in `b1370626` ("regenerate cli-reference.json for the refactored help text"). **The test passes at head. Please disregard that finding.** Apologies for the churn.

---

### 🔴 Critical

**C1 — Root command injection via the agent-account name (`localagent.go:719`, `CopyBinaryCmd`; and `:769`, `CopyConfigCmd`).**
Both build a `sudo sh -c` script that splices the account name **unquoted** into a command substitution:
```
dest := "$(eval echo ~" + agentUser + ")/.local/bin"   // agentUser NOT quoted here
```
The account name is an operator-editable free-text form field (`agentuser_form.go:26-30`, only `notEmptyField` validation). A name like `$(command)` or `x; …;` executes arbitrary commands **as root** (the `eval` compounds it with double expansion). The `//nolint:gosec` on line 722 asserts the value is "shell-quoted" — it is quoted in the later `chown` occurrence but NOT in the `$(eval echo ~…)` occurrence. **Reproduced mechanically**: building the exact `CopyBinaryCmd` script with a payload name and running it (without sudo) executed the injected command. Under the real path it runs as root.
Fix: resolve the agent home in Go (`user.Lookup(agentUser).HomeDir`) and pass argv instead of `sh -c`; never `eval echo ~$name`. And validate the name (below).

### 🟠 High

**H1 — sudoers-content injection via the account name (`localagent.go:524`, `SudoersRule`).** `operator + " ALL=(" + agentUser + ") NOPASSWD: " + shell` — the name is shell-quoted for the *shell* in `InstallSudoersCmd` but unescaped in the *sudoers content*. A crafted name (`x) ALL=(ALL:ALL) NOPASSWD: ALL #`) yields a syntactically valid passwordless-root rule that `visudo -cf` (syntax-only) accepts. Gated on the operator typing a malicious name, but the impact is root.

**H2 — Setup reuses (and re-owns) any pre-existing Unix account of the target name (`agentuser.go:165-217`).** `createAgentAccount` treats *any* existing account of that name as "ours" (`UserExists` = `id -u`), then unconditionally runs `ReclaimAgentHomeCmd` (`sudo chown -Rf <name> <homeDir>`), `GrantOperatorHomeCmd` (recursive operator ACL), installs the passwordless sudoers rule, and seeds config into that home. If the name collides with a real human/service user, setup recursively re-owns and ACL-opens *their* home. No check that the account is jentic-managed.

**H3 — `reset --delete-home` / re-own act on an un-revalidated, operator-editable home path (`reset.go` → `localagent.go:447` `ReownHomeCmd`, `:480` `DeleteHomeCmd`).** `home_dir` comes from config (originally the free-text form field, only non-empty-validated) and is fed straight to `sudo rm -rf <homeDir>` / `sudo chown -Rf <op> <homeDir>` with no `/Users/Shared`|`/opt` allowlist and no `lstat`/symlink guard. A typo/hand-edit/legacy value (`/Users/alice`, `/`) targets the operator's own tree; a symlinked home is followed by `rm -rf`. The `//nolint:gosec` "homeDir is a resolved path" is false — it's never re-resolved on teardown.

**Common root cause for C1/H1/H2/H3:** `agentUser` and `homeDir` are unvalidated operator-editable inputs treated as trusted throughout. **Single highest-value fix:** a shared `ValidateAgentUser` (`^[a-z_][a-z0-9_-]*$`, length-bounded) + `ValidateHomeDir` (absolute, cleaned, under `/Users/Shared` on darwin / `/opt` on linux, reject `..`/symlink/human-home roots), enforced in the form `Validate()` **and** re-checked in `localagent` before any command is built. Add `--` separators to the `chown`/`rm`/`setfacl`/`dscl` argv builders as defense-in-depth.

**H4 — Copied profile secrets can land group/world-readable (`identity_target.go:185`, `copyFile`; `:158`, `copyTree`).** `os.OpenFile(dst, …, 0600)` / `MkdirAll(dst, 0700)` are masked by umask; under a permissive umask (`002`/`007`/`000`, common in some dev/CI envs) the copied `agent.key`/`tokens.json` land `0640`/`0660`. The store's own `writeFileAtomic` (`store.go:217`) already `Chmod`s explicitly — `copyFile`/`copyTree` should too. Fix: explicit `os.Chmod(dst, perm)` after copy.

**H5 — `translateOperatorProfile` deletes the operator original without a post-copy read-back (`identity_target.go:135-144`).** The move of an unrecoverable Ed25519 key + registration `Delete()`s the source after `copyTree`; a copy error aborts before delete (good), but there's no fsync and no verification the key is readable at the destination before the irreversible delete. Fix: re-open dst, assert key present + `LoadMeta` succeeds, then delete.

### 🟡 Medium

- **M1 — `runSameUser` is a silent security downgrade (`run.go:162,282`).** When no agent account exists, `jentic run` launches the agent **unconfined, as the operator, with full filesystem access** — no warning. Emit a prominent one-time warning before this fallback.
- **M2 — Linux `--dev-bind / /` parity gap (`confine.go:327`).** Only `/Users` and `/home` are tmpfs-masked; operator secrets outside a human-home root (home at `/opt`, `/var/lib/…`, group-readable configs) stay visible/writable — no macOS counterpart. Either document the invariant (this layer only closes the in-home sibling leak) or prefer `--ro-bind / /` + explicit RW rebinds.
- **M3 — Linux exec-route `--ro-bind` breaks under a masked home (`confine.go:338-340`).** A PATH exec dir under `/home` (e.g. Linuxbrew's `/home/linuxbrew/.linuxbrew/bin`) is wiped by `--tmpfs /home`, then `--ro-bind` of the now-missing source fails the whole launch. Bind such routes from the host before the tmpfs, or exclude+assert.
- **M4 — `config.Save` is not atomic (`config/file.go:207`).** Direct `os.WriteFile` onto `config.yaml`; a crash mid-write can drop `agent_account`/`granted_dirs` (the authoritative grant record). Mirror the store's temp+rename+chmod.
- **M5 — `README.md` command-surface table omits the new `client` group** (`run`/`reset`/`skill`) and several `jentic` commands.
- **M6 — docs still describe the obsolete `chmod 700 ~` model** (`filesystem-access-model.md:397` "pristine 700-closed state" contradicts the doc's own "we no longer chmod 700 ~" thesis; also present-tense "700 home" at :241/:270). And the `runSameUser` unconfined fallback isn't flagged in the residual-risks/README as reintroducing full same-user exposure.

### 🟢 Low / Nits

- `classifyGrantStderr` (`run.go:591`) matches the English `"No such file or directory"` — fails closed under non-C locales (spurious hard failure, not a hole). Run grants with `LC_ALL=C`.
- `Classify` uses `filepath.Abs`+`Clean`, not `EvalSymlinks` (`localagent.go:990`) — symlink-into-sensitive lexical bypass; bounded by the sandbox + `! -type l` stamp. Defense-in-depth: `EvalSymlinks` before classify/grant/record/revoke so the persisted grant and the sandbox re-allow key off the same real path.
- macOS case-insensitivity: ban prefix checks are case-sensitive, so `~/.SSH` bypasses the HardBan on APFS/HFS+ (`localagent.go:1003,1014`). Use `EqualFold` on darwin.
- `.config` is HardBanned wholesale (`localagent.go:1050`) — blocks legit `~/.config/<project>` grants; and the list has dead/redundant entries (`.gcloud` doesn't exist; `.config/google-chrome` redundant under `.config`).
- `RemoveAgentIdentityCmd` (`sudo rm -rf <configDir>`) runs as a fatal step before sudoers/account removal (`reset.go`), so a failure strands the higher-risk leftovers — make it best-effort or reorder after account delete.
- `AgentACLPresent` unanchored `Contains("user:"+name)` and `RemoveSudoersCmd` unanchored `grep -v <name>` (name as regex) — already noted; low impact under the single shared account, but anchor to exact principal / full rule line.
- `handOffToAgent` `chown -R` (no `-h`) — mitigated by the tool's no-follow default + `copyTree` skipping symlinks; add a clarifying comment.
- sessionPaths prefers the operator copy on a same-name tie (`profile.go:164`), relying on translation having removed it (best-effort) — consider preferring the agent ref when the account is enabled.

### ✅ Verified correct (broad)
`--` arg-forwarding parser (both forms, `--` twice, `=`-joined, empty args); `launchAgent` error-closed contract (only unconfined path is `runSameUser`, M1); exit-code propagation; banned-path grant has no escape hatch under any flag combo; SBPL last-match ordering + ancestor `(literal)` vs grant `(subpath)`; `sbplPath` escaping (matches the known-good Bazel fix); Linux sibling-hiding via tmpfs; sudoers file-integrity (mktemp + `visudo -cf` + 0440, idempotent); `run`/`reset`/`skill` correctly on `jentic` and absent from `jenticctl`; access.go exit codes 2/3/4 + the 7-line deletion is a safe refactor; store `Delete()` path safety; store secret perms (atomic + explicit chmod); run-as access map sourced from the same `SessionAccess` the sandbox uses (no drift); `AncestorChain`/`IsUnderHome`/`isDirectChild`; the plan docs are correctly status-tagged/superseded.

Prioritise **C1 → H1/H2/H3 (all fixed by the shared name/home validator) → H4/H5**. Happy to dig deeper on any of these.

Manuel-Jentic:
## Third review pass — 6 cross-cutting reviewers (data-flow, teardown-atomicity, adversarial-sandbox, concurrency, tests, history)

This pass approached the PR by *lens* rather than by file, to catch cross-file issues the earlier per-file passes missed. Head unchanged (`46eafbb`). New findings below; I verified the significant ones against the code (and note one earlier claim I'm **correcting**). Build + full test suite **pass** at head, including `TestCommittedCLIReferenceUpToDate` (re-confirming my retraction).

### 🔴 New — SBPL profile injection voids the sandbox (macOS)

**`sbplPath` (`confine.go:306`) escapes only `\` and `"`, not newlines.** The SBPL profile is line-oriented (`SandboxProfile` writes one rule per `\n`) over an `(allow default)` base with last-match-wins. A path containing a newline — the agent **home** or a **granted dir** — emitted into `(allow file* (subpath "…"))` terminates the rule early; bytes after the `\n` are parsed as new top-level rules. An injected `(allow file* (subpath "/Users"))` / re-emitted `(allow default)` re-opens the human-home deny the confinement exists to enforce — silently defeating the PR's core, "required / error-closed" control. macOS filenames may contain newlines, so this is reachable via a crafted `--grant`/path argument or a subdir an agent creates inside an already-granted workspace, which is then recorded verbatim. (Linux `bwrapArgs` is NOT affected — each path is a `shellQuote`'d argv token.) Fix: reject control chars (`\n`/`\r`/NUL) in any path before it's recorded as home/grant, and make `sbplPath` refuse newlines (SBPL has no line-continuation). Add a `sbplPath` newline test.

### 🔴 New — macOS ACL builders inject into the ACE grammar (distinct from the earlier shell-injection)

`TraverseGrantCmd`/`LeafGrantCmd`/`LeafRevokeCmd` build `"user:" + agentUser + " allow execute"` (and `macLeafACE`) as a single **argv** token (`localagent.go:346,381`). No shell injection — but `agentUser` is spliced into the **ACL-entry grammar** unvalidated, so a name like `x allow read,write,file_inherit,directory_inherit` re-parses the ACE and turns an intended execute-only traverse into a full inherited read/write grant on the operator's home ancestors. Same root cause as the earlier injection findings (no name validation), different sink that argv-quoting does not protect.

### 🟠 New — concurrent `config.Save` → unrevokable dangling ACL on the operator's home

`config.Save` (`config/file.go:216`) is a bare `os.WriteFile` with no lock and no atomic temp+rename; every mutation is `Load → mutate snapshot → Save`. Two concurrent `jentic run <dir>` (or a wizard's workspace multiselect held open while a `run` grants elsewhere) lose one `AddGrantedDir` to last-writer-wins — **the ACL is stamped on the operator's real home but the dir is dropped from `granted_dirs`.** `reset` builds teardown solely from `GrantedDirs`, so that ACE is **never revoked** and survives account deletion (a dangling named-user grant on `~/Documents`-style paths). `recordAgentAccount`/`offerWorkspaceGrants` have the same clobber window. Also non-atomic `Save` can truncate the authoritative config on crash. Fix: a single `config.Mutate(paths, fn)` helper that flocks, reloads under the lock, and writes atomically — routes `grantDir`/`runRevoke`/`recordAgentAccount`/reset's clears through one correct path.

### 🟠 New — network egress is unrestricted, so any readable secret is exfiltratable (macOS)

The SBPL profile is `(allow default)` with only `file*`/`file-write*` denies — **no `network*` rule**. Under the PR's own compromised/prompt-injected-agent threat model, everything the agent may read — the seeded `~/.aws`/`~/.config/gcloud`/`~/.claude.json` provider creds (`CopyConfigCmd` explicitly copies these into the agent home), the granted working dir's `.env`/tokens, world-readable `/tmp` — can be shipped out in one HTTP request. Confinement is filesystem-only; that's a reasonable scope but it undercuts the "the agent can only touch granted dirs" mental model. Fix: at minimum document loudly that egress is unrestricted and seeded creds are exfiltratable; ideally `(deny network*)` + a narrow allowlist for the control-plane loopback + configured provider endpoints (Linux: `bwrap --unshare-net` + proxy).

### 🟡 New — Medium

- **Agent login shell runs UNSANDBOXED before `exec` (`confine.go:69-80`, `localagent.go:600`).** Launch is `sudo -u agent -H bash -lc '… && exec sandbox-exec …'`. `bash -l` sources the agent-owned `~/.bash_profile`/`.bashrc` **before** the `exec sandbox-exec`, so agent-controlled shell-init runs as the agent uid with no confinement on every launch — a guaranteed unsandboxed foothold. Fix: wrap the shell itself (`sudo -u agent sandbox-exec -p <profile> /bin/bash -lc …`) or use a non-login shell with an explicit env.
- **`~/.local/bin` holds the exec'd agent binary but is NOT a read-only exec route (`confine.go:280`, PATH-prepended `localagent.go:628`).** The sandbox write-denies `/usr/bin` etc. but not the agent-writable `~/.local/bin` where `claude` actually lives — violating the stated "a compromised agent can't rewrite the binaries jentic run executes." Fix: add the agent-home bin dir on the launch PATH to `execRouteDirs`/read-only routes.
- **`GOOGLE_APPLICATION_CREDENTIALS` → root `cp -R` of an arbitrary path (`localagent.go:822`).** For Vertex, the raw creds path from `~/.claude/settings.json` is appended to provider config paths with no under-home constraint, then `CopyConfigCmd` copies it as root and chowns it to the agent; an out-of-home value (`/etc/…`, a symlink) is copied across the boundary (and produces a `//etc/…` dest). Fix: require provider paths resolve under the operator home; skip otherwise.
- **`--agent-user` flag reaches every privileged sink unvalidated (`run.go:316`)** — an independent entry point into the injection sinks even after the setup form is fixed; validate it too.
- **Setup records the account AFTER the privileged create steps (`agentuser.go:144`).** A crash/kill between `createAgentAccount` and `recordAgentAccount` leaves a real Unix account + sudoers + ACLs with **no config record**, so `reset` reports "nothing to reset." Fix: write a provisional record before the first privileged step; have reset reconcile via `UserExists`.
- **Shared sudoers drop-in read-modify-install race (`localagent.go:538`)** — concurrent install/remove for two agent names can drop or resurrect a passwordless rule. Fix: `flock` around the cat→edit→install critical section.

### Correction to a claim from the concurrency/adversarial reviewers (verified myself)

The reviewers suggested `LeafGrantCmd`'s `find <dir> ! -type l` would **follow a top-level symlink** and stamp ACLs on the real target (a HardBan bypass). **I empirically tested this on macOS (BSD find) and it's false**: `find <symlink> ! -type l` treats the arg as the link, excludes it, and does **not** descend into the target — the grant harmlessly no-ops on a symlinked leaf. So there is no macOS ACL-stamp-through-symlink bypass. The *real* residual is narrower and still worth fixing: `Classify` is lexical-only (no `EvalSymlinks`), so a symlinked *path* bypasses the ban *check*; and on **Linux** `setfacl -R` on a symlinked dir may follow it (I couldn't verify — no Linux box). Fix remains: `EvalSymlinks` + re-`Classify` the resolved path at grant time, and reject a symlink leaf.

### Teardown residue (new orderings)

- **`granted_dirs` is pruned all-or-nothing** (only at `ClearAgentAccount`, `reset.go:451`) — a partial/interrupted reset leaves every grant recorded while some ACLs were revoked; prune per-successful-revoke.
- **`AgentACLPresent` matches the account NAME (`localagent.go:437`)**, but once `DeleteAccountCmd` removes the record, macOS renders the ACE by numeric uid — so a re-run's name probe returns false and **skips a dangling ACE on the operator's home**. Capture the uid before deletion, or always revoke recorded `granted_dirs` regardless of the name probe.
- **macOS `chmod -a "user:<name> allow execute"` (`TraverseRevokeCmd`) only removes an ACE whose perms match exactly** — a widened/mismatched ancestor ACE survives revoke (Linux `setfacl -x` removes by name regardless). 
- **Identity dir can survive a failed `--delete-home`** (identity-removal step is emitted only when home is kept, but home delete is best-effort). Always emit it.

### Test gaps (guarding Critical invariants)
No test for: name/home validation (**it doesn't exist**), banned-path-under-any-flag (`decideDirGrant` untested), error-closed launch (`ConfinementAvailable`/`launchAgent` gate untested), symlink-in-grant, copied-secret 0600 under loose umask, config atomicity. And two tests pass while the property is broken: `AgentACLPresent` (substring, bot vs robot) and `RemoveSudoersCmd` (`grep -v` substring strips another agent's line) — both need two-agent fixtures.

### The one fix that closes the most
A shared **`ValidateAgentUser`** (`^[a-z_][a-z0-9_-]{0,31}$`) + **`ValidateHomeDir`** (absolute, cleaned, under `/Users/Shared`|`/opt`, no `..`, **no control chars**), enforced in the form `Validate()` AND re-checked in `localagent` before any command is built — closes the root injection, the sudoers injection, the ACE-grammar injection, the SBPL-newline injection, and the unvalidated-`chown`/`rm` targets. Pair it with: resolve homes in Go (`user.Lookup`, drop `eval echo ~`), a locked+atomic `config.Mutate`, `EvalSymlinks` on grant paths, and a `(deny network*)` decision. Given the WIP status and ~8.8k additions of root-privileged builders, I'd treat these as merge blockers.

Manuel-Jentic:
## Fourth review pass — launch/env, sandbox scope, and identity-handoff (all net-new)

Three more reviewers, on ground the earlier passes didn't cover: the **launch/env/PATH** path, the **non-file scope** of the sandbox, and the **token/identity hand-off**. Everything below is new (deduped against my prior comments); I verified each load-bearing claim against the code at head `46eafbb`, and I ran a couple of harmless macOS sandbox probes. Two of these independently defeat the credential boundary before any file ACL/SBPL rule matters.

### 🔴 Unconfined-launch hijack via the agent's own `~/.local/bin` on PATH

The confined launch (`confine.go:156`) is `exec sandbox-exec -p '<profile>' claude <args>` inside `sudo -u <agent> -H bash -lc …`. **Both `sandbox-exec` and the agent binary are bare names** (`binary := desc.Binary`, `run.go:150`) resolved by the agent's login-shell PATH — and `EnsureLocalBinOnPathCmd` **prepends** the agent-writable `$HOME/.local/bin` (`localagent.go:628`), which is **not** in `execRouteDirs` (`confine.go:280,287`). Even if it were, the read-only-exec-route rule lives *inside* the profile that `sandbox-exec` applies — it can't protect `sandbox-exec` itself. A compromised agent drops `~/.local/bin/sandbox-exec`; the operator's next `jentic run` resolves it there and runs it **completely unconfined** with the operator's cached sudo credential (the wrapper just ignores `-p` and execs the real agent). This is the launcher's *own* tool resolution, so it breaks confinement before the sandbox is ever established — strictly worse than the earlier "`~/.local/bin` isn't read-only inside the box" note. Fix: hard-code `/usr/bin/sandbox-exec` (absolute) and resolve the agent binary to an absolute path from the descriptor; never let the agent's PATH pick the confinement tool.

### 🔴 `chown -R` in the identity hand-off follows agent-planted symlinks → local priv-esc

`ChownToAgentCmd` = `sudo chown -R <agent> <dir>` (`localagent.go:473`) and `ReclaimAgentHomeCmd` = `sudo chown -Rf <agent> <home>` (`localagent.go:464`) — **no `-h`/`-P`/`--no-dereference`**. The dir (`~/.jentic/...`) is inside the **agent-owned, agent-writable** home, and the hand-off reclaims/writes/chowns while the agent account exists. Between steps the agent pre-creates `~/.jentic` (or a leaf) as a **symlink to `/etc/sudoers.d`** (or any root-owned path); the operator's root `chown -R` then descends the link and chowns the target to the agent, who rewrites its own sudoers rule to `(root)`. Classic recursive-chown-through-symlink privilege escalation on exactly the multi-tenant boundary this PR creates. Fix: write the identity into an **operator-owned staging dir** and move it in atomically; use `chown -Rh`/`--no-dereference` and refuse to traverse symlinks; verify every component of `<configDir>` is a real dir owned by the expected uid before writing/chowning.

### 🟠 Operator environment (incl. `SSH_AUTH_SOCK`) leaks into the agent via `sudo`

The launch (`agentCmdContext`, `localagent.go:616`) sets **no `cmd.Env` and no `-E`** — env policy is left entirely to the host's `sudo` `env_keep`. `-H` only fixes `HOME`. Variables operators routinely export — `AWS_*`, `ANTHROPIC_API_KEY`/`OPENAI_API_KEY`, `GITHUB_TOKEN`/`GH_TOKEN`, and especially a forwarded **`SSH_AUTH_SOCK`** — survive on stock configs and land in the agent process. `SSH_AUTH_SOCK` is the worst: it hands a prompt-injected agent the operator's *live ssh-agent* (key use) even though `~/.ssh` is HardBanned and sandbox-denied — the file boundary is bypassed by the forwarded socket. The code comment claims "login shell so no operator environment leaks" (`run.go:48`) but nothing enforces it. Fix: set an explicit minimal allowlist `cmd.Env` for the sudo launch (or `sudo env -i` + re-add only `HOME`/`PATH`/`TERM`/`LANG`/`JENTIC_PROFILE`), and explicitly unset `SSH_AUTH_SOCK`.

### 🟠 macOS sandbox is filesystem-only: non-file operations and non-home paths stay open (verified live)

The profile is `(allow default)` + `file*`/`file-write*` denies on `/Users`+`/home` only — nothing else is constrained. I confirmed under the exact profile shape on macOS 15.6 (harmless probes, no operator secrets read):
- **`process-exec*` is open** — a sandboxed proc ran an arbitrary `/usr/bin/true`; a compromised agent can spawn any helper on the system.
- **`mach-lookup` is open** — `pbpaste` resolves, i.e. the pasteboard mach service (and others) are reachable; add clipboard scraping / IPC to running operator services to the agent's toolkit.
- **`/private/var/folders` (per-user caches/temp), `/tmp`, `/Library`, `/usr/local`, `/etc/passwd` are all read/write-or-read** — they sit *outside* the `/Users`+`/home` deny, so operator caches under `var/folders`, `/tmp` droppings, and `/Library/LaunchAgents`-style persistence are fully reachable.

Correction to keep this honest: I could **not** confirm cross-process argv/env reads — a sandboxed `ps -axww` returned 0 matches for a marker process's args vs 3 unsandboxed, so Seatbelt *does* restrict `kern.procargs2` even under `(allow default)`. I'm **not** claiming operator-process argv disclosure. Fix: add explicit `(deny process-exec* …)` (allowlist the agent's real toolchain), `(deny mach-lookup …)` for pasteboard/keychain/etc., and extend the file deny to `/private/var/folders/<op>`, `/Library`, `/tmp` (or invert to a deny-by-default file model).

### 🟠 Linux bwrap re-exposes the whole rootfs and shares host namespaces

`bwrapArgs` does `--dev-bind / /` and then tmpfs-masks only `/Users`+`/home` (`confine.go` ~315-330). Everything else — `/root`, `/tmp`, `/run/user/<uid>` (D-Bus/Wayland/gcloud sockets), `/etc` — stays read/write through the root bind. And there's no `--unshare-pid/ipc/net` and `/proc` is the host's, so the agent sees operator processes and can reach abstract-domain sockets (X11/D-Bus) and loopback services. (`--die-with-parent` *is* correctly set — good.) Fix: build the mount set allowlist-style (bind only what's needed) rather than root-bind-then-mask, and add `--unshare-all` + `--share-net` only if egress is intended.

### 🟠 Provider credentials are copied to the agent verbatim, and `cp -R` (as root) follows symlinks

`CopyConfigCmd` (`localagent.go:768-780`) runs `sudo sh -c "… cp -R <src> <dest> …"` to seed `~/.claude(.json)`, `~/.aws`, `~/.config/gcloud`, `GOOGLE_APPLICATION_CREDENTIALS` into the agent home. Two problems: (1) these are the operator's **real, un-scoped provider secrets** handed to a separate (adversarial) account — the agent-scoped DCR token this PR correctly mints then sits right next to fully-privileged provider keys, defeating the boundary for those; (2) `cp -R` (not `-RP`) **follows source symlinks while running as root** reading the 0700 operator home, so a symlink under `~/.aws` → `/etc/shadow` (or any out-of-home target) is dereferenced and copied into the agent-readable home. Fix: prefer the LLM-proxy/agent-scoped-credential path over copying; for any copy that remains use `cp -RP`/`rsync --safe-links` and reject sources whose symlinks escape the operator home.

### 🟡 Medium — new correctness/lifecycle gaps

- **macOS launch has no `--die-with-parent` analog & no process-group kill (`run.go:748`).** The launch is `sudo → bash -lc → sandbox-exec → agent` via `cmd.Run()` with no `Setpgid`; on SIGKILL/terminal-close/crash of `jentic run`, the sandboxed agent can reparent to launchd and **keep running detached** with its granted ACLs while the operator thinks the session ended. Linux is torn down by `--die-with-parent`; macOS isn't. Fix: `SysProcAttr{Setpgid:true}` + kill the group on context-cancel (Seatbelt has no parent-death signal).
- **`translateOperatorProfile` deletes the operator original before verifying the copy (`identity_target.go:133-146`).** `copyTree` **silently skips symlinks** (`:166`) and `copyFile` does no post-copy verify/fsync; a symlinked `tokens.json`/`agent.key` in the source is dropped, then the original is deleted → lost token cache / orphaned registration. And a *partial* copy leaves a dest that the next run treats as authoritative (the "already present" short-circuit at `:112`). Fix: copy to a temp sibling, fsync + verify all expected entries, atomic-rename, then delete; treat an unexpected symlink as an error, not a skip.
- **Grant "success" is inferred from stderr text, not the actual ACE (`run.go:568-605`).** `classifyGrantStderr` treats a non-zero grant as benign whenever every stderr line ends in `No such file or directory`; since the leaf grant walks the agent-influenced tree with `find … -exec chmod +a#`, churn/symlink games can make failures look benign while the intended leaf ACE was never stamped — "Granted." is reported without verifying the leaf. Fix: re-check the leaf's ACE directly after granting; only treat missing-child errors as benign once the leaf is confirmed stamped.
- **Approval isn't bound to the provisioned identity (`register.go:249-305`, `agentauth/session.go`).** `waitForApproval` accepts the mint result without a `/me` check that the token subject equals the approved `agent_id`; `--force` (`ResetRegistration`) re-registers a new `agent_id` while keeping the same key, and `aud` comes from the operator-config `base_url` — a tampered `base_url` mints against an attacker control-plane. Fix: after mint, assert `/me` `id == AgentID`, `type==agent`, `status==active`; pin/echo the exact `agent_id`+`base_url` being approved.
- **`ConfigDir` steering the identity write+chown is unvalidated (`identity_target.go:39-49`).** It's used verbatim as the root that gets written then `chown -R`'d (feeds the priv-esc above). Fix: require `ConfigDir == AgentConfigDir(HomeDir)`, a real non-symlink dir owned by the agent uid, before use.

### 🟡 Low — refresh token echoed
`register.go:159-160` prints `shorten(RefreshToken)`/`shorten(AccessToken)` to `a.Out`; `shorten` only masks strings >16 chars, so a short token prints in full, and even the masked refresh token is echoed into a terminal/log the agent or a co-tenant may read. Fix: never echo refresh tokens; mask with a min-length guard; gate any token echo behind `--show-secrets`.

### Checked and NOT issues (so they don't get re-raised)
`--`-arg forwarding quoting is solid (each arg independently single-quoted; no break-out of `bash -lc` or the `sandbox-exec`/`bwrap` argv in either parser form; Linux `bwrap --` stops flag-injection). `IsUnderHome` is correctly anchored (`home + separator`) — no sibling-prefix bug. Tokens go in the `Authorization` header, not argv/env. `SudoersRule` is `NOPASSWD: /bin/bash` runas the agent only (no root capability) and is `visudo -cf`-validated. `bwrap --die-with-parent` is set.

The two I'd gate merge on from this pass are the **unconfined `sandbox-exec` hijack** and the **`chown -R` symlink priv-esc** — each fully bypasses the boundary on its own. Combined with the unvalidated-name root injection from earlier passes, the privileged builders need absolute tool paths + no-dereference chown/cp + an explicit env allowlist before this is safe to run on a shared host.

Manuel-Jentic:
## Fifth review pass — TUI/reset-crash, HTTP/transport, and reset teardown-completeness (net-new)

Three more reviewers on ground the earlier passes hadn't touched. This pass was quieter — a lot of what I checked is actually *correct*, and I've listed those explicitly so they don't get re-litigated. New findings below, each verified against code at head `46eafbb`; severities reflect that most inputs here are operator-controlled, not remote.

### 🔴 A "clean slate" reset leaves seeded provider credentials in the (now operator-owned) home

`jentic run`/bootstrap can copy the operator's real provider secrets into the agent home — `CopyConfigCmd` seeds `~/.claude`, `~/.claude.json` (may embed the Anthropic API key), and `~/.aws` / `~/.config/gcloud` (`localagent.go:768,817,864`). But the **default** reset path (home kept) only runs `RemoveAgentIdentityCmd(<configDir>)` — the agent's `~/.jentic` — then `ReownHomeCmd` (`reset.go:601-623`). It **never removes the seeded `.aws`/`.claude`/`.config/gcloud`.** So a live API key / AWS SSO config the operator handed the agent survives every default "clean slate" reset, now sitting in an operator-owned dir, while the banner says the home was merely "KEPT and re-owned to you" (`reset.go:697`). Fix: in the keep path, `rm -rf` the known seeded provider dirs (mirroring `RemoveAgentIdentityCmd`), or at minimum have `printResetPlan` warn that the kept home may still hold seeded provider creds and that only `--delete-home` purges them.

### 🟠 `jentic reset` is blind to a config-less but OS-real account → orphaned sudoers with no removal path

`resetAll` calls the disk-scanning `surveyReset` **only** when `hasPlan` is true, and `hasPlan = hasAcct && acct.User != ""` reads purely from `config.yaml` (`reset.go:104-113`). Because `recordAgentAccount` runs *after* `CreateAccountCmds`/`InstallSudoersCmd` (`agentuser.go:136-144`), a crash/kill in that window leaves a real `<op>-local-agent` Unix user **and `/etc/sudoers.d/jentic-agent`** on disk with a `nil` config row → `jentic reset` prints "Nothing to reset" and there is **no CLI path to remove the orphaned passwordless-sudoers rule.** (This is the reset-side gap, distinct from the earlier "bootstrap records after privileged steps" note — the point here is reset has no live-OS fallback despite `surveyReset` already supporting the empty-account default-name case.) Fix: when `!hasPlan`, still probe `UserExists(DefaultUserName(operator))` + the sudoers drop-in and offer teardown of whatever's on disk.

### 🟠 No scheme/host validation on `base_url`; plaintext HTTP is the default

`httpx.New` (`httpx.go:46-51`) only trims the base URL — no `url.Parse`, no `https` enforcement, no host check — and the default is `http://127.0.0.1:8000` (`config/paths.go:21`), with a plaintext example in the test config. Every agent bearer token, refresh token, and the JWT-Bearer DCR assertion is therefore sent in **cleartext** to whatever `base_url` resolves to; a typo or a copied insecure example downgrades all agent-credential traffic with no warning, and there's no guard against `base_url` pointing at `169.254.169.254`/loopback/`file://`. Medium (operator-controlled config, not remote input), but it compounds the earlier "aud/base_url mints against attacker control-plane" point — plaintext makes passive interception trivial. Fix: in `httpx.New`, `url.Parse` and require `https` (allow `http` only for an explicit loopback/dev opt-in), reject non-host schemes, fail closed.

### 🟡 Medium/Low — new correctness gaps

- **Profile-name collision makes `reset <name>` unable to target an agent profile (Medium).** `findProfileRef` prefers the operator-owned profile on an ambiguous name (`profile.go:164-179`) and same-named operator+agent profiles are kept distinct (`profile.go:104,112`), but `reset` has no owner selector. So if both own a `work` profile, `jentic reset work` **always** deletes the operator's and can never reach the agent's; the "last agent profile → account teardown" offer is gated on `ref.owned()` (`reset.go:197`) so it's silently suppressed too. Fix: add `--agent` (thread into `findProfileRef`), or error on an ambiguous name and list both owners.
- **Declining passwordless on a reuse run never removes an existing sudoers rule (Medium).** `createAgentAccount` re-asserts `InstallSudoersCmd` only when `fields.passwordless` (`agentuser.go:243`); there's no "remove if declined" branch, so a drop-in installed on an earlier run persists after the operator says "No" on a later one. Fix: run `RemoveSudoersCmd` on the reuse path when `!fields.passwordless` so the toggle is authoritative both ways.
- **`reset <profile> --delete-home` silently no-ops on the common path (Low).** `--delete-home` is only consulted inside the last-agent-profile interactive-teardown branch (`reset.go:249`); for any non-last profile, and even for `--force` on the last (where `tearDownAccount` is forced false, `reset.go:227-231`), the flag is ignored with no warning. Fix: warn when `--delete-home` is passed to a profile reset that won't tear the account down.
- **`api_id` interpolated unescaped into the catalog request path (Low).** `catalogPath` (`catalogclient/client.go:251`) concatenates the CLI-supplied `api_id` raw (deliberately, for `{api_id:path}` vendors), while `apiclient.apiPath` (`apiclient/client.go:305`) correctly `PathEscape`s every segment; a `../`/`//host` value can steer the request path. Local-operator self-targeting, hence Low. Fix: percent-encode all but the intended `/`, or validate the `vendor/...` shape.
- **Aborting the operator multiselect surfaces a raw `user aborted` (Low).** `chooseAdapters`/`skill.go:228` returns `huh.ErrUserAborted` unwrapped up to `bootstrapE`, unlike the clean "Cancelled." handling elsewhere (`bootstrap.go:106`). Cosmetic; no state impact. Fix: treat `ErrUserAborted` there as a clean cancel.
- **Operator's inherited ACE on the kept home is never stripped on reset (Low/cosmetic).** `GrantOperatorHomeCmd` stamps recursive inherited `user:<operator> allow …` ACEs across the agent home (`localagent.go:206`); the keep-path `chown -Rf` preserves extended ACLs, so every file in the re-owned home carries a redundant explicit operator ACE the "your own home is untouched" framing (`reset.go:703`) doesn't mention. Not a privesc (operator granting itself); confuses later audits. Optional: strip during re-own, or note it.

### Doctrine note
This PR intentionally **removed the old `chmod 700 ~` step** (`agentuser.go:228`, `confine.go:31`) — so any "home left at 700 / never restored" concern is moot. The flip side: in-home confidentiality now rests *entirely* on the confinement layer, which the prior passes showed has an unconfined-launch hijack, non-file sandbox scope, and a bwrap rootfs re-expose. Removing the 700 fallback means those confinement gaps are now the **sole** barrier between the agent and the operator's home, with no defense-in-depth.

### Checked and CLEAN (so these aren't re-hunted)
**Transport/HTTP:** no `InsecureSkipVerify`, no custom transport, no `CheckRedirect` — Go's default redirect policy caps hops and **strips `Authorization` cross-host** (bearer isn't replayed to a redirect target); the auth header is set per-request, never persisted; response bodies are bounded (`io.LimitReader` 64 MiB); every client has a finite timeout; poll loops honor a deadline + `ctx.Done()`, back off with a cap, and never retry on hard errors or re-send on 401. **Crypto:** Ed25519 from `crypto/rand`, stored PKCS#8 at 0600, assertion carries `iat`/`exp` (2-min) + fresh `jti` per call; the CLI treats received tokens as opaque (no client-side JWT-validation gap). **TUI:** grant-confirm has no TOCTOU (same `abs` reused through Classify→decide→grant); `--force`/`--delete-home` bypasses are documented+intentional; no zero-value/dangerous-default confirm; typed confirms only `TrimSpace`. **Wiring:** `cli-reference.json` is in sync at head (both sync tests pass), `reset` is registered and captured, `skip()` correctly drops Hidden/help/completion (no hidden leak); account-delete is genuinely last (no uid→name loss); no PATH exports land on operator profiles.

Net for this pass: no new criticals, but the **provider-cred-survives-reset** and **reset-blind-to-orphaned-account** findings are real teardown holes, and the **plaintext-HTTP default** is worth closing. The earlier passes' criticals (unvalidated-name root injection, `sandbox-exec` hijack, `chown -R` symlink privesc) remain the merge blockers.

Manuel-Jentic:
## Sixth pass — composed exploit chains + doc/reality gaps (net-new; scope-checked)

Final pass: an adversarial *synthesis* (composing known gaps into worse chains), a docs-vs-code accuracy review, and a sweep of the remaining subsystems. Every link below is cited to code at head `46eafbb` and verified. **Scope note up front:** I also reviewed the `jenticctl update` installer path (unsigned-installer-exec, `repo`/`ref` from `install.json`, ldflags injection) — but this PR's only change to `update.go`/`manifest.go`/`paths.go`/`cmd/update.go` is a `jentic`→`jenticctl` **doc-comment rename**, so that whole surface is **pre-existing and out of scope for #853** (worth a separate issue against the update path, not this PR). Excluded from the findings below.

### 🔴 CHAIN 1 — Reset-surviving root persistence (the recovery guarantee is defeatable)

A single prompt-injection becomes a foothold that outlives a full `jentic reset`:
- **A —** the agent is launched by *bare name* `claude` (`run.go:150`), resolved from the agent-writable `~/.local/bin` (prepended to PATH, `localagent.go:628`). Inside the sandbox the agent's **own home is re-opened read-write** (`SessionAccess`, `localagent.go:110-121`) while only the *shared/system* bin dirs are held read-only (`execRouteDirs`, `localagent.go:287-302`) — **`~/.local/bin` is never in the read-only set.** The `SandboxProfile` comment claims "the binaries jentic run executes stay read-only … no sandbox self-escape" (`confine.go:258`), but that only covers the system routes, **not the agent-home route the binary actually resolves from**. So the confined agent overwrites its own `~/.local/bin/claude`.
- **B —** a default `reset` **re-owns but does not delete** the home (`reset.go:617-623`) and only removes the `~/.jentic` identity — never `~/.local/bin` or the PATH-export lines. The trojan persists on disk through the "clean slate."
- **C —** re-enabling isolation runs `chown -Rf <agent> <home>` (`ReclaimAgentHomeCmd`, `localagent.go:464`), handing the planted binary back to the agent; `ProbeBinary`/`BinaryOnPath` then sees `claude` "on PATH" and **skips reprovisioning** (`run.go:355`), and `launchAgent` execs it under `sudo -u agent`. No integrity check anywhere.

Net: inject once → overwrite own `~/.local/bin/claude` → survive a full reset → re-detonate on the next isolated launch. This is the composition the earlier passes missed — they flagged the bare-name `sandbox-exec` PATH resolution, not that the **agent's own binary route is writable and re-exec'd across a reset**. Fix: hold agent-home exec routes read-only in the sandbox too, and on reset scrub/quarantine home-local bin dirs + jentic PATH lines even when the home is kept (or refuse to reuse a kept home without re-provisioning the binary from a trusted source).

### 🟠 CHAIN 3 — Corrupt/truncated `config.yaml` silently downgrades to an unconfined same-user launch

`launchAgent` *does* error-close when confinement is unavailable (`run.go:725`) — but `runSameUser` is an **earlier, ungated** branch reached whenever `HasAgentUser()` is false (`run.go:157-162`), and it runs the agent **as the operator, no confinement, no ACLs, inheriting the operator's full environment** (`run.go:298-303`, `c.Env = append(os.Environ(), …)`). `HasAgentUser()` is a bare read of `config.yaml` (`file.go:245`), and `Save` is a non-atomic `os.WriteFile` (`file.go:216`). So a crash/concurrent write that truncates `config.yaml` (dropping the `agent_account` block) → `HasAgentUser()` false → the *next* `jentic run` silently launches unconfined as the operator. The earlier passes flagged the non-atomic save only as "lost grant → dangling ACL"; the **downgrade from confined-isolated to operator-privileged** is the materially worse composition. Fix: atomic `Save` (temp+fsync+rename), and distinguish "never set up" from "record present-but-damaged" — refuse to silently same-user-launch when an account/home for the derived name actually exists on the OS.

### 🟠 CHAIN 2 — Deliberately induced un-cleanable root orphan

Composing the record-after-privileged ordering (`agentuser.go:136-144` — Unix user + `NOPASSWD` sudoers at `:244` + seeded creds all run *before* `recordAgentAccount`) with reset keying off the config record (`resetAll` early-returns "Nothing to reset" on `!hasPlan`, `reset.go:126-129`): an attacker with operator code-exec can **SIGKILL between `InstallSudoersCmd` and `recordAgentAccount`** to manufacture a real account + passwordless-sudoers rule + seeded creds that `reset` cannot see or remove (the name is the deterministic `<operator>-local-agent`). This is the *induced* version of the reset-blindness I reported in pass five — here it's an attacker primitive, not just a crash artifact. Fix: write a "provisioning-in-progress" marker to config *before* the first privileged step, and have `reset` fall back to a live OS scan (account existence + sudoers grep for the derived name) when no record exists.

### 🟠 Docs & operator-facing banners promise protection the code doesn't deliver

The most dangerous doc gap is a **false "you're jailed" assurance**:
- `filesystem-access-model.md:38` is titled "**Layer 0 — default-deny is the confinement profile**"; line 45 says non-granted paths are "**invisible to the agent**"; the mechanism table (233/235) describes bwrap as "everything else on the host is *invisible*" and Seatbelt as "`(deny default)` … only permits the granted subpaths." **The shipped profile is `(allow default)` + `/Users`+`/home` deny** (`confine.go:216`) — I confirmed live that process-exec, mach-lookup (pasteboard), `/tmp`, `/Library`, `/private/var/folders`, `/etc/passwd`, and **network egress** all stay open. The doc describes the *rejected* design as if implemented. (Notably `confine.go`'s own header and `sandbox-exec-plan.md` are honest about the targeted-deny model — so align `filesystem-access-model.md` and the banners to *them*.)
- The operator-facing **"fully locked-down agent sessions"** error (`run.go:726`) and **"(confined)"** banner (`run.go:745`) overstate a session with open egress and non-home reads.
- `run.go:48` help claims the login shell means "**so no operator environment leaks**" — the exact property the `SSH_AUTH_SOCK`/env-leak finding (pass four) disproves.
- `filesystem-access-model.md:141` states "the binaries jentic run executes stay read-only" unconditionally, but only a hardcoded bin-dir list is protected (Chain 1 shows the agent-home route isn't).
- The reset banner "your own files … **NOT touched**" / "**re-owned to you**" (`reset.go:697,703`) is contradicted by the `chown -R` symlink issue — and the doc even *claims* the symlink-safe `find ! -type l` form is used for re-own while `ReownHomeCmd` doesn't. Align code to doc.

A false assurance is arguably worse than a documented gap — an operator runs untrusted agents believing exfiltration is blocked. Fix: rewrite the model prose + banners to state plainly that this is per-session *home* confinement, that network egress and non-home reads are **not** restricted, and mark `(deny default)`/"invisible" as the rejected option.

### 🟡 Medium / Low — new single issues

- **Linux userns availability probe fails OPEN on read error (`confine.go:360-365`).** `unprivilegedUserNSEnabled` returns `true` for *any* `os.ReadFile` error (masked `/proc`, LSM denial, unreadable knob), not just "knob absent" — so `ConfinementAvailable` can report available when it isn't, and the whole error-closed contract leans on this boolean. Fix: treat a read error as **unavailable** (fail closed), consistent with the design.
- **Workspace picker pre-selects everything from an agent-writable trust source (Medium).** `offerWorkspaceGrants` pre-checks all entries (`agentuser_workspaces.go:86`, "pre-select all") sourced from `~/.claude.json` `projects` (`localagent.go` `TrustedWorkspaces`). A pre-isolation same-user agent (or Chain 3's downgrade) can add a `projects` entry with `hasTrustDialogAccepted:true` for a broad/symlinked path; it then appears **pre-checked**, and an operator pressing Enter grants it — removing the friction the model relies on. Fix: default the picker to nothing selected (opt-in), and derive trust from a source the agent can't write pre-isolation.
- **`RemoveSudoersCmd` cross-account strip / dangling-rule reuse (Low, new angle on a known bug).** The `grep -v '<agentUser>'` (`localagent.go:554`) is an unanchored substring: resetting `bob-local-agent` also strips `bob-local-agent-2`'s `NOPASSWD` line, or leaves a dangling rule a future same-named account inherits. (Earlier I noted the substring bug generically; the cross-account and rule-reuse impacts are the concrete new angle.) Fix: `grep -vxF` the exact `SudoersRule(...)` whole line.

### Net
No new *criticals* in the subsystem sweep, and I explicitly **excluded the pre-existing update-path RCE surface** as out-of-scope. But **Chain 1** is arguably the most important finding of the whole review: it defeats the advertised "full clean slate" recovery, composing the writable agent-home exec route + no-home-delete reset + chown-back-on-reuse. Combined with the earlier merge-blockers (unvalidated-name root injection, unconfined `sandbox-exec` hijack, `chown -R` symlink privesc), the isolation model needs the exec-route/reset-scrub fix, an atomic+validated config with a fail-closed confinement probe, and the doc/banner language corrected to stop over-promising, before this is safe on a shared host.

Manuel-Jentic:
## Re-verification against the new head (`3701592`, +6 commits since `46eafbb`)

Six new commits landed after my previous passes — the confinement **prerequisite preflight/gate** work (`ab1f174`, `10922c5`, `7edfc3b`, `b703b76`) plus two doc/comment commits (`06ad46c`, `3701592`). I re-extracted this head, rebuilt, and re-checked every finding against the changed files. **Build + `go test ./internal/localagent ./internal/cmd` PASS.**

### What the new commits genuinely improve
`MissingPrereqs()`/`Prereq` + `ConfinementAvailable` (`confine.go:112`) is a clean, structured preflight: it checks `sandbox-exec`/`bwrap`/`setfacl`/`getfacl`/userns and prints exact per-distro install commands, `agentuser.go:121` gates account creation on it, and `run.go` surfaces the missing-prereq hints at the launch gate. This makes the error-closed contract more robust and much friendlier. Good change.

### What these commits do NOT fix — the earlier findings still stand at this head (verified against the new files)
- **Agent name/home validation — still absent.** The new gate is prerequisite-only; there is still no `ValidateAgentUser`/`ValidateHomeDir`, so the **root command-injection / sudoers / macOS-ACE / SBPL-newline injection** chain (my top merge-blocker) is unchanged.
- **Unconfined `sandbox-exec`/binary hijack — unchanged.** Launch still resolves `sandbox-exec` and the agent binary by bare name off the agent-writable `~/.local/bin` PATH.
- **`chown -R`/`cp -R` symlink privesc — unchanged** (`ReclaimAgentHomeCmd`/`ChownToAgentCmd`/`CopyConfigCmd` still no `-P`/`--no-dereference`).
- **Chain 1 (reset-surviving persistence) — unchanged**, and the new prereq-gate arguably makes **Chain 3 (unconfined downgrade) more reachable**: `agentUserPrereqGate` (`agentuser.go:184`) now *offers* "continue same-user right now," so the unconfined-as-operator `runSameUser` path (`run.go:298-303`, still reached on `!HasAgentUser()` before any confinement check, inheriting the operator's full env) is now a first-class branch. A corrupted/truncated `config.yaml` still silently routes into that same execution mode.
- **S1 — the Linux userns probe still fails OPEN on read error.** `unprivilegedUserNSEnabled` (`confine.go:490-495`) still `return true` on *any* `os.ReadFile` error, so the otherwise-stronger new gate can still report confinement available when `/proc` is masked/denied. This one is a small, clean fix inside the very code these commits refactored — treat a read error as unavailable.
- **Doc / banner false-assurances — NOT corrected.** The doc commits only fixed `jentic`→`jenticctl` command names. Verbatim at this head: `filesystem-access-model.md:38` still titled "default-deny is the confinement profile", line 45 "invisible to the agent", table rows 233/235 still describe bwrap/Seatbelt as `(deny default)`/"invisible"; `run.go:48` still claims the login shell means "so no operator environment leaks"; `run.go:727` still "fully locked-down agent sessions." These over-promise a jail the `(allow default)`+home-deny profile (open egress/process-exec/mach-lookup, readable `/tmp`//`Library`) does not provide.
- All the teardown-residue, config-atomicity, base_url-plaintext, provider-creds-survive-reset, and approval-binding findings are in files these 6 commits didn't touch — unchanged.

### Net
The preflight/gate work is a real improvement to *availability/UX* of confinement, but it does not close any of the substantive security findings from the six review passes. The merge-blockers remain: **(1)** unvalidated agent name/home → root injection, **(2)** unconfined `sandbox-exec`/binary hijack, **(3)** `chown -R` symlink privesc, **(4)** Chain-1 reset-surviving persistence — plus the fail-open userns probe (a quick win in the newly-refactored code) and the doc/banner language that still over-promises. Happy to re-review once these are addressed.

Manuel-Jentic:
## Re-verification at head `06d1160` (+20 commits since `3701592`) — merge-blockers resolved

Re-ran the full harness against the new head after the merge from `main` + the `fix(localagent/cli/config)` block. **`go build ./cmd/jentic` + `./cmd/jenticctl` both build; `go test ./...` passes (all 22 packages `ok`, incl. the new `validate`/`lock`/`update`/`install` tests).** I verified each fix in code, not just from the commit message.

### The four merge-blockers are closed
| Finding | Fix | Verified in code |
|---|---|---|
| Unvalidated agent name/home → root cmd / sudoers / ACE / SBPL injection | `1cb6b81` (`validate.go`) | `ValidateAgentUser`/`ValidateHomeDir`/`ValidateConfigDir`/`ValidateGrantPath`, wired at the form, `--agent-user` (`run.go:198`), setup incl. the **operator** name (`agentuser.go:251-254`), identity hand-off (`identity_target.go:93`), **and the reset teardown** (`reset.go:432-462`) before every `rm -rf`/`chown`. Defense-in-depth as recommended (validate at source + re-check before every privileged command). |
| `eval echo ~<name>` home resolution | `2496efd` | resolved in Go, no shell eval |
| SBPL newline injection | `c6b4c8a` | control chars rejected at source + stripped at the `sbplPath` sink |
| `sandbox-exec`/binary PATH hijack | `ff282e8` | `sandboxExecPath`/`bwrapPath` pinned to absolute, resolved in the operator process; outer shell now **non-login** (`bash -c`) so no agent rc runs before the wrapper takes hold; login shell runs *inside* the sandbox |
| `chown -R`/`cp -R` symlink privesc | `c824a2a`, `3f9035e` | symlink-safe hand-off/teardown; provider seeding constrained under the operator home |

### Other reported findings also fixed (verified)
- **Operator env / `SSH_AUTH_SOCK` leak** — `5b3d919`: `UnsetSensitiveEnvSnippet` scrubs SSH/GPG agent handles in-snippet regardless of sudoers `env_keep`.
- **Config race → dangling ACL / unconfined downgrade** — `fa92d7b` (`lock.go`): atomic writes + serialized mutations under a lock.
- **Reset-surviving persistence / touches-wrong-tree (Chains 1–2)** — `09d6c281`: `VerifyManagedHome` fails closed unless the account's *live* OS home equals the recorded managed `/opt`//`Users/Shared` path, so a name-collision or hand-edited `home_dir` can't steer a privileged reclaim/reown/delete.
- **Delete-before-verify data loss** — `bebc967`: pinned copied-profile perms + fsync before the irreversible delete.
- **Benign-stderr grant classifier** — `2ca8061`: forces C locale and verifies the ACE actually landed.
- **Substring ACL-presence / sudoers removal** — `2c205f9`: anchored against prefix siblings.
- **Refresh-token echo + approval-not-bound-to-identity** — `4f971c0`: stops echoing the refresh token; asserts the minted token's identity via `/me`.

### Two residuals remain (both minor now)
1. **Linux userns probe still fails OPEN.** `unprivilegedUserNSEnabled` (`confine.go:579-585`) still `return true` on *any* `os.ReadFile` error (masked/denied `/proc`), so the otherwise-stronger prereq gate can report confinement available when it isn't. Given the whole model is error-closed, a read error should be treated as **unavailable**. One-line fix in code you just refactored.
2. **`~/.local/bin` still not in the read-only exec set** (`execRouteDirs`, `confine.go:475`), so a compromised agent can overwrite its own launched `claude`. With the wrapper-path pinning + `VerifyManagedHome` now in place, this is no longer a sandbox escape or reset-surviving root persistence — it's *confined* re-execution of a self-modified binary. **Severity dropped from Critical to Low / defense-in-depth**; worth adding the home-local bin dir to the read-only routes, but not a blocker.

Net: the security posture is materially where it needs to be — nice work turning these around. I'd still land the two residuals (both small) before merge, and I'm filing a separate note on operator coverage (`jentic run` currently supports only `claude`; a plan to reach parity with the skill-supported operators is coming).

ren-jentic:
Pushed the two remaining review residuals (verified in code + covered by new tests):

- **Userns probe now fails closed** (`f0d98b6`): `unprivilegedUserNSEnabled` treats only a clean not-exist of the Debian sysctl knob as "enabled" (mainline/RHEL kernels); any other read error (masked/denied `/proc`) is now "disabled", matching the error-closed contract of the rest of the confinement layer. Knob path extracted to a var; `TestUnprivilegedUserNSProbe` covers absent / 0 / 1 / unreadable.
- **Agent `~/.local/bin` is now a read-only exec route** (`f0d98b6`): `execRouteDirs` takes the agent home and includes `<home>/.local/bin`, so both `SandboxProfile` (write-deny emitted after the home re-allow, last-match-wins) and `bwrapArgs` (ro-bind after the home bind) pin the launched binary read-only. `TestAgentLocalBinIsReadOnlyExecRoute` asserts presence *and* ordering on both platforms; `jentic profile view` picks it up for free via `SessionAccess`.

Also `5fe9a65`: the banned-path warning no longer renders the refused path via `strings.ToUpper` (misrepresented paths on case-sensitive filesystems).

Housekeeping: retitled the PR to conventional format for the squash merge and refreshed the description (dropped the stale "locks the operator's home" claim — the shipped design deliberately no longer chmods the operator home to 700 — and filled in the Testing section). Build, vet, full test suite, and golangci-lint all pass locally at the new head.

ren-jentic:
Live-testing the bootstrap → launch flow on macOS surfaced a launch-blocking bug, fixed in `5fd4659`:

**Every confined launch failed with `sudo: bash: command not found`.** The launch hands sudo the curated `launchEnv` — which deliberately carries no `PATH` — but the sudo argv named the shell as a bare `bash`, which sudo resolves against *its* environment's PATH. macOS's default sudoers sets no `secure_path` to fall back on, so the confined launch (the only path using the stripped env) always died; the probe/grant sudo paths masked the bug because they inherit the operator's full environment. Both argv builders now exec the shell by the absolute `agentLaunchShell` (`/bin/bash`) — no resolution needed, and it's the exact command the sudoers NOPASSWD rule is scoped to. `TestConfineLaunchCmdUsesAbsoluteShell` pins the argv shape and the no-PATH precondition that makes it load-bearing.

PR #857, opened 2026-07-28
fix(auth): repair expired-token login race and add sliding web sessions

### LINKED ISSUE #608: [surface] token times out on local host
### What's broken

Token times out on the screen and it's not clear you need to refresh 

### Steps to reproduce

1.
2.
3.

### Expected

_No response_

### Actual (paste exact output/errors)

```shell

```

### Environment

- Jentic One version/commit:
- OS:

---
*Migrated from the internal dogfooding tracker (feedback#20, originally filed 2026-07-02). Screenshots/attachments referenced in the original will be re-attached.*

### LINKED ISSUE #610: [surface] Need to try things a couple of times before responsive - eg. Login
### What happened

I don't have a screenshot - but when my local host UI is there but logged out I need to try 2 / 3 times before the login works. I submit the correct info and it just refreshes to a blank screen - I continue trying until it works. 

### Where (command or page)

_No response_

### Why it was a pain point

Wasting the time of a user - confusing to know if it is the wrong login or the product is simply not working. Breaks trust that this product is reliable. 

### What you expected instead

The login to always work without issue. 

### Environment

- Jentic One version/commit:
- OS: Mac
- Install method (install.sh / source / Docker): Docker

---
*Migrated from the internal dogfooding tracker (feedback#29, originally filed 2026-07-08). Screenshots/attachments referenced in the original will be re-attached.*

### LINKED ISSUE #813: 401 on dashboard URLs returns raw JSON instead of redirecting to login
## What happened

When a user's session expires and they visit a protected dashboard URL (e.g. an access-request approval page), the response is raw JSON:

```json
{"type":"unauthorized","status":401,"title":"Unauthorized","detail":"Authentication required","instance":"/access-requests/areq_6a675e7c62d4db00882747b2"}
```

## Expected behavior

Browser requests to protected UI pages should **redirect to the login screen** (302 → `/login?next=...`) when the session is expired or missing, not return a JSON API error body.

This is standard web app behavior — the user should land on login, authenticate, and be redirected back to the page they were trying to reach.

## Impact

The operator is left staring at a JSON blob with no obvious way forward. They have to manually navigate to the dashboard root and find the access request themselves. Given that this is the approval step (the critical human-in-the-loop moment), this friction directly blocks agent workflows.

## Steps to reproduce

1. Install Jentic One and log into the dashboard
2. Wait for the session to expire (or clear cookies)
3. Visit `http://127.0.0.1:8000/access-requests/<any_request_id>` in a browser
4. Observe: raw 401 JSON instead of a login redirect

## Screenshot

401 response in browser

Manuel-Jentic:
Closes #610. Closes #608. Closes #813.

## TL;DR (human)

Logging in with an expired token in `localStorage` showed a blank screen and only worked after 2–3 retries (#610). The cause was a UI race: requests still carrying the **old** expired JWT got 401s *after* login stored the **new** token, and the error handler wiped the new one. This PR fixes that race and gives web sessions a real lifecycle: silent renewal while you're using the app, a hard 12-hour cap, a friendly "session expired" notice, browser 401s that land on the login page instead of raw JSON (#813), and DB cleanup of long-expired token rows.

**Verified end-to-end**: Bugbot and a dedicated security review both came back clean, and the full session story was smoke-tested against the real backend + built SPA with a headless browser (see [the smoke-test comment](https://github.com/jentic/jentic-one/pull/857#issuecomment-5115776721)).

## What you experience, scenario by scenario

**Scenario 1 — You log in and keep working (all day).**
Nothing visible changes. Behind the scenes, when your 1-hour token reaches ~80 % of its life (~48 min), the UI silently calls `POST /auth/refresh` and swaps in a fresh token. This repeats as long as the tab is open, so you never get logged out mid-work. Each renewal also re-reads your permissions from the DB — so a permission change or account deactivation takes effect within the hour, not "whenever you next log in".

**Scenario 2 — You leave the tab open overnight (past the 12-hour window).**
Renewal only works for 12 hours after your original password login (`admin.auth.session_ttl_seconds`). Past that, the next refresh is refused (401 `session_expired`), the UI signs you out, and the login page says *"Your session expired — please sign in again."* You sign in and you're back. This cap is what makes a stolen token useless after at most `session_ttl + jwt_ttl`.

**Scenario 3 — Your laptop was asleep / the tab was in the background.**
Browsers throttle timers in hidden tabs, so the scheduled renewal may be missed. The moment the tab becomes visible again, a `visibilitychange` listener catches up: if the session window is still open you're silently renewed; if not, you get the expired notice. Either way — no blank screen.

**Scenario 4 — You come back tomorrow with an expired token in `localStorage` (the original #610 bug).**
Before: blank screen, works after 2–3 retries — login stored the new token, but in-flight requests still signed with the old dead token got 401s and *wiped the new token*. Now: adopting a session clears in-flight query state first, so a late 401 from an old request can never clobber the new session. Login works on the **first attempt** (regression-tested in UI tests and in the browser smoke test).

**Scenario 5 — You paste a dashboard URL into the browser while logged out (#813).**
Before: raw JSON Problem Details in your face. Now: anonymous HTML `GET` navigations that hit a 401 redirect to `/app` (the login page). Strictly guarded — only when the SPA is mounted, only `GET`, only HTML `Accept`, only when **no** credential was presented — so curl, API clients, and requests with a bad token still get proper RFC 9457 JSON.

**Scenario 6 — Agent/platform tokens (not web sessions).**
`POST /auth/refresh` refuses opaque `at_` tokens and non-user actors — it's for human web sessions only. Separately, the background worker sweep now purges access/refresh token DB rows that have been expired for over 7 days (the `delete_expired` repo methods existed but were never called), so dead rows stop accumulating.

**Scenario 7 — Mixed versions during rollout.**
Tokens minted before this change (no `actor_type`/`auth_time` claims): still valid for normal use, but **not refreshable** — refresh fails closed on tokens that don't explicitly declare those claims, so pre-upgrade sessions simply end at their natural JWT expiry and the user signs in once more (a one-time cost). Old UI + new backend: unaffected (refresh is additive). New UI + old backend: refresh 404s once — no retry loop — and the session ends at natural expiry, exactly like today.

## Config (both optional, documented in `config/README.md`)

```yaml
admin:
  auth:
    jwt_ttl_seconds: 3600       # single JWT lifetime (existing)
    session_ttl_seconds: 43200  # NEW: absolute session window for refresh
```

Enforced invariant: `session_ttl_seconds >= jwt_ttl_seconds` (config validator fails closed). Note the last refresh just inside the window still mints a full-TTL token, so a session's hard end is at most `session_ttl + jwt_ttl` — documented trade-off.

## For agents / reviewers (machine-oriented map)

| Concern | Files |
|---|---|
| Login race fix + refresh scheduling | `ui/src/shared/auth/AuthContext.tsx` |
| Token/expiry/notice storage | `ui/src/shared/api/token-store.ts`, `ui/src/shared/api/index.ts` |
| Expired notice UI | `ui/src/shared/auth/LoginPage.tsx` |
| Refresh endpoint | `src/jentic_one/admin/web/routers/auth.py` |
| Refresh logic + `auth_time` minting + audit | `src/jentic_one/admin/services/auth_service.py`, `invite_service.py` |
| Errors | `admin/services/errors.py` (`SessionExpiredError`, parametrized `InvalidCredentialsError`), `admin/web/errors.py` (401 `session_expired`) |
| Config + invariant | `shared/config.py` (`AdminAuthConfig.session_ttl_seconds`, validator) |
| 401→/app redirect | `shared/web/app_factory.py`, `shared/web/static.py` (`spa_mounted` flag) |
| Token row sweep | `shared/jobs/worker.py` |
| OpenAPI/docs | `shared/web/openapi_meta.py`, `endpoint_scopes.py`, `openapi/control/control.openapi.yaml`, `docs/reference/*` (regenerated) |

Invariants to preserve when touching this later:
- `auth_time` must survive every re-mint; only `login`/`bootstrap`/invite-redemption/change-own-password may reset it (each requires a fresh credential proof).
- Refresh **fails closed**: a token must explicitly carry `actor_type: "user"` and `auth_time` — missing or unrecognised values are refused (401), never defaulted. All login-JWT mint paths set both claims.
- `POST /auth/refresh` accepts the **raw** bearer JWT (claims outside the `Identity` contract); opaque `at_` tokens and non-user actors are refused; each refresh writes an `AuditAction.REFRESH` audit row.
- The 401→`/app` redirect must stay limited to: SPA mounted + `GET` + HTML `Accept` + no credential presented.

## Verification

- [x] `make lint` (ruff, format, mypy) clean
- [x] Backend: unit (`test_config.py`, `test_app_factory.py`), web (`tests/web/admin/test_auth.py`), service integration (`test_auth_service.py`), worker sweep (`tests/integration/shared/test_worker_token_retention.py`) — all passing
- [x] UI: `tsc`, ESLint, vitest — including new tests for the #610 regression, proactive refresh success/refusal, and token-store expiry/notice semantics
- [x] **Bugbot review: no bugs found.** Security review: no issues found (HS256 pinned, no token-confusion path with the connect-state JWT, `auth_time` not resettable via refresh, no open redirect, sweep can't touch live/revoked-but-unexpired rows)
- [x] **Manual smoke test, time-compressed** (30 s JWT / 70 s window standing in for 1 h / 12 h) against the real backend + built SPA via headless Chromium: first-try login → silent renewal at ~80 % TTL (old token proven dead server-side with a 401 probe while the tab stayed signed in across a reload) → auto sign-out + expired notice past the window → first-try re-login. Full timeline + server-log trail (`login_success` → `session_refreshed` ×2 → `session_expired`) in [this comment](https://github.com/jentic/jentic-one/pull/857#issuecomment-5115776721)

> Merge with **squash + merge** (repo convention — PR title becomes the squash commit).


Manuel-Jentic:
## Manual smoke test — session lifecycle verified end-to-end

Ran the full session story against the **real backend + real built SPA** (no mocks), time-compressed so the hours become seconds: `admin.auth.jwt_ttl_seconds: 30` (stands in for the 1 h default) and `session_ttl_seconds: 70` (stands in for 12 h), on a fresh SQLite instance driven by headless Chromium (Playwright).

Timeline (all assertions passed):

| t | Event | Verified |
|---|---|---|
| 0 s | First-run setup → dashboard | Login works on the **first attempt** |
| 24.8 s | Silent refresh at ~80 % of TTL | Token rotated in `localStorage`, expiry extended; server logged `session_refreshed` |
| 33.5 s | Past the original token's 30 s lifetime | Original token probe returns **401** (genuinely dead server-side) while the open tab is **still signed in, even across a full page reload** |
| 74.7 s | Past the 70 s absolute window | Next refresh refused (`session_expired`), user signed out automatically, login page shows *"Your session expired — please sign in again."*, token cleared from storage |
| 74.8 s | Re-login | Succeeds on the **first attempt** — the #610 retry dance is gone |

Server-side log evidence: `login_success` → `session_refreshed` ×2 → `session_expired` refusal, plus one `AuditAction.REFRESH` audit row per refresh.

This is the time-compressed equivalent of: *log in, leave the tab open past 1 h → session silently renews; keep it open past the 12 h window → expired notice.*

Manuel-Jentic:
**Review follow-up (fail-closed refresh)** — addressed in f8ff8fd/5f520fe: `refresh` previously defaulted a missing `actor_type` claim to `USER`. Now login JWTs carry an explicit `actor_type: "user"` claim, and refresh refuses any token that doesn't explicitly declare both `actor_type: "user"` and `auth_time` — no defaults, no `iat` fallback. Side benefits: an unrecognised `actor_type` value can no longer raise `ValueError` (500) — it's a clean 401 — and pre-upgrade tokens simply expire at their natural TTL (one extra sign-in after deploy). New tests cover missing/unknown `actor_type`, missing `auth_time`, and the legacy-token refusal.

Issue #845, opened 2026-07-28
access-requests: add a cheap pending-count endpoint (GET /access-requests/count)


Manuel-Jentic:
## Context

The webapp's nav badge (`usePendingAccessRequestCount`) and the enterprise admin console's Requests tab both need "how many requests are pending" — and today the only way to get it is to fetch a full page (`GET /access-requests?status=pending&limit=50`) and count the rows client-side. That's a full envelope (items included) serialized for a number, and it silently caps at the page size.

## Proposal

A dedicated count endpoint, e.g.:

```
GET /access-requests/count?status=pending
→ { "count": 12 }
```

- Same visibility filter as the list (`build_access_filters`), so it's caller-scoped for members and org-wide for `org:admin`.
- `status` optional; probably worth supporting the same values as the list filter.
- SQL `COUNT(*)` instead of hydrating rows + items.

## Consumers ready to switch

- OSS nav badge (`ui/src/shared/app/rail/usePendingAccessRequestCount.ts`) — currently fetches a limit-50 page.
- Enterprise admin console Requests tab (jentic-one-enterprise, `ui/src/modules/admin/`) — could show true per-segment counts, which server-side cursor pagination currently makes impossible.

(Recorded in the enterprise `docs/admin-console-todo.md` §9.)

Manuel-Jentic:
Closing after building and then deliberately removing this in PR #858: the nav badge merges several approximate sources (the agents-tab badge page-fetches with an N+ floor, and the dashboard bell renders actual rows so it needs a page anyway), so an exact count for just the access-request slice doesn't improve anything user-visible. Queues big enough to trip the 50-cap indicate a larger operational problem than an inexact badge. The PR branch history (commits a2a1334/2f4c83a, removed in e0f672f) documents a working COUNT + group_by=status design should a real breakdown consumer (e.g. console per-segment counts) ever justify it.

PR #858, opened 2026-07-28
feat(access-requests): filer-owner enrichment, widened UI type, shared queue helpers

### LINKED ISSUE #846: access-requests: enrich list/get responses with filer-owner display info (avoid client-side roster joins)
## Context

`GET /access-requests` responses carry `filer_owner_id` / `created_by` as raw ids. Any surface that wants to show *who* is behind a request (the enterprise admin console's org-wide Requests queue renders an Owner column; the OSS dashboard queue shows the actor only) has to join those ids against a separately-fetched users roster client-side.

That works for the admin console (it already loads `/users` for its Users tab), but:

- it forces every consumer to have `users:read`-level visibility just to label a row;
- N ids resolve against a full roster fetch, not a targeted lookup;
- ordinary (non-admin) users can't resolve owner ids at all, so the same UI can't be reused down-scoped.

## Proposal

Enrich the access-request response envelope with a display block, e.g.:

```json
{
  "filer_owner_id": "usr_…",
  "filer_owner": { "id": "usr_…", "email": "owner@example.com" }
}
```

(or a `display_names` sidecar map on the page envelope, mirroring how other joins could be batched). Server-side it's one join against a table the service already reads.

## Consumers ready to switch

- Enterprise admin console Requests tab (jentic-one-enterprise, `ui/src/modules/admin/tabs/RequestsTab.tsx`) — currently resolves via the cached `/users` query (`UserLabel`).

(Recorded in the enterprise `docs/admin-console-todo.md` §9.)

### LINKED ISSUE #847: ui: AccessRequest type omits wire fields (created_by, filer_owner_id, approve_url) — widen the hand-rolled shared/lib type
## Context

The backend's `AccessRequestResponse` serializes `created_by: str`, `filer_owner_id: str | None`, and `approve_url: str` on every list/get item (`control/web/schemas/access_requests.py`), and the generated client model agrees (`ui/src/shared/api/generated/models/AccessRequestResponse.ts`).

But the hand-rolled `AccessRequest` interface in `ui/src/shared/lib/accessRequests.ts` omits all three, so any consumer that needs them has to widen the type locally. The enterprise admin console does exactly that (`AdminAccessRequest extends AccessRequest` adding the three fields, in jentic-one-enterprise `ui/src/modules/admin/api/types.ts`) — a copy that will drift.

## Proposal

Add the fields to the shared `AccessRequest`:

```ts
export interface AccessRequest {
  // …existing…
  /** The principal that created the row (the agent itself for self-filed requests). */
  created_by: string;
  /** The filer's human owner (null for ownerless actors). */
  filer_owner_id?: string | null;
  /** Deep link into the webapp's decision surface. */
  approve_url: string;
}
```

They're already on the wire, so no backend change — just type truth. Then delete the enterprise widening.

Related: while touching this file, consider also exporting the page-local helpers duplicated across queue surfaces (tracked separately).

(Recorded in the enterprise `docs/admin-console-todo.md` §9.)

### LINKED ISSUE #848: ui: export shared access-request summary + status-variant helpers — three page-local forks exist
## Context

Three surfaces now hand-roll the same access-request row presentation, each page-local and therefore un-importable by the others:

- `ui/src/modules/dashboard/pages/AccessRequestsPage.tsx` — `summarize()` + a status→`Badge` variant map;
- `ui/src/shared/app/rail/ActorAccessRequestsCard.tsx` — the same pair, forked;
- the enterprise admin console's Requests tab (jentic-one-enterprise `ui/src/modules/admin/tabs/RequestsTab.tsx`) — a third fork, extended (`partially_approved`/`expired` as `warning`, explicit `withdrawn`).

The copies already disagree: OSS renders the rarer statuses (`partially_approved`, `expired`, `withdrawn`) as the `default` (primary-tinted) badge, while the console renders two of them as warnings — so the same request wears different colors on different surfaces. And when a new status lands, three files need the same edit.

## Proposal

Export from `ui/src/shared/lib/accessRequests.ts` (next to the existing `itemTargetLabel`/`scopeLabel`):

```ts
/** "toolkit · bind — github/github@v3 +2 more" — the one-line queue-row summary. */
export function summarizeAccessRequest(request: AccessRequest): string;

/** Wire status → shared/ui Badge variant, covering all six statuses incl. derived `expired`. */
export const ACCESS_REQUEST_STATUS_VARIANT: Record<string, BadgeVariant>;
```

…then delete the three copies. Deciding the canonical tones for the rarer statuses (neutral vs warning) is part of this issue.

(Recorded in the enterprise `docs/admin-console-todo.md` §9.)

Manuel-Jentic:
## Summary

Lands three of the four access-request follow-ups filed from the enterprise admin-console work, so API consumers (including the enterprise console) can drop their client-side workarounds:

- **`filer_owner` enrichment** (#846) — list/get now batch-resolve `filer_owner_id` against the admin `users` table (one query per page, mirroring `_resolve_names`) and return `{id, email, display_name}`. Falls back to `created_by` when `filer_owner_id` is null (legacy rows) — the same fallback consumers render. Labelling only, never authorization; ids that aren't users stay `null`.
- **Widened UI `AccessRequest` type** (#847) — `created_by`, `filer_owner_id`, `filer_owner`, and `approve_url` are now on the shared hand-rolled type; `filed_at`/`expires_at` corrected to required (the wire always carries them).
- **Shared queue helpers** (#848) — `summarizeAccessRequest` + `ACCESS_REQUEST_STATUS_VARIANT` exported from `shared/lib` (with vitest coverage); the duplicated copies in the dashboard queue page and the actor card are deleted.

**Deliberately NOT landed: the count endpoint (#845).** The nav badge merges several approximate sources (the agents-tab badge and the dashboard bell both page-fetch — the bell renders actual rows), so an exact count for just the access-request slice buys nothing; pending queues large enough to trip the 50-cap signal a bigger operational problem than an inexact badge. Built, tested, then removed — the history documents the design if a real breakdown consumer ever materializes.

### Deliberate behavior notes (explicit sign-offs)

- **Owner email/name exposure:** any principal that can already read a request now sees its owner's email + display name without `users:read`. Scoping is untouched (members still only see own/owned requests), so the practical disclosure is "an agent learns its own owner's label" — intended, so consoles can label rows for non-admin viewers.
- **OSS badge-tone change:** `partially_approved`/`expired` now render as `warning` (previously `default`) on the OSS queue surfaces — converging on the enterprise console's mapping.

### Reviews

Four independent review passes (Bugbot, security, architecture/rules, feasibility/API-design): no blockers; both should-fixes (created_by fallback coverage, shared-helper vitest coverage) and all nits addressed in the final commit.

Closes #846, closes #847, closes #848.

## Test plan

- [x] `make openapi` + `npm run codegen` — spec and generated client committed, parity check green
- [x] Web tests: owner enrichment on list+get, created_by fallback for null owner ids, null degradation for unresolvable ids
- [x] Full backend unit + arch suites green (2,283 passed; endpoint reference regenerated)
- [x] UI typecheck, eslint/prettier, and vitest suite green (784 passed, incl. new shared-helper tests)
- [x] Rebased on the #859 dashboard rebuild


Manuel-Jentic:
Rebased onto the dashboard rebuild (#859) and extended the count endpoint with `group_by=status`: one `GROUP BY` returns the whole per-status breakdown (`{count, by_status}`), so segment consumers don't poll once per status. The plain `{count}` badge shape is unchanged (`by_status` omitted unless requested). Post-#859 note: the new action-inbox bell renders actual rows so it correctly keeps fetching a page; the count endpoint's consumers are the nav/bottom-bar badge (exact count, no more "50+") and per-status segments.

Issue #862, opened 2026-07-29
Auth gate defaults a missing actor_type claim to USER — should fail closed


Manuel-Jentic:
## Context

Spun out of the review of #857 (sliding web sessions). During that review the question was raised for the refresh endpoint:

> feels like we should maybe default to something other than user if we can't find the actor type? Just to have it fail closed

#857 fixed this **for `POST /auth/refresh`** (it now requires an explicit `actor_type: "user"` claim and refuses anything else) and made all login-JWT mint paths embed the claim explicitly. But the same fail-open default exists **platform-wide** in the shared auth gate, which was out of scope for that PR.

## Current behavior

`verify_token` in `src/jentic_one/shared/auth/verify.py` resolves the actor type of every JWT-authenticated request like this:

```python
actor_type = ActorType(claims.get("actor_type", ActorType.USER))
```

Two problems, in decreasing severity:

1. **Fail-open default.** A validly-signed JWT with no `actor_type` claim is treated as a USER. Today this is only reachable by tokens we minted ourselves under `admin.auth.jwt_secret` (historically without the claim), so there is no known concrete exploit — but the posture is wrong: an auth gate should refuse to guess, especially since USER is the *most* privileged interpretation (agents get parent-scoped permissions, service accounts get their own).
2. **Junk values 500 instead of 401.** `ActorType("anything-unknown")` raises `ValueError`, which surfaces as an unhandled 500 rather than a clean `invalid_credentials` 401. (Same bug was fixed in the refresh endpoint in #857.)

`Identity.actor_type` (`src/jentic_one/shared/auth/identity.py`) also defaults to `ActorType.USER`, which invites the same pattern in future call sites.

`verify_token` is used by the admin web app, the auth app, the broker setup, and the shared web deps — i.e. this affects every JWT-authenticated endpoint, which is why it deserves its own change with its own rollout plan rather than riding along on #857.

## Proposed fix

1. After #857 ships (all newly-minted login JWTs carry an explicit `actor_type: "user"`), tighten `verify_token` to **require** the claim: missing → reject as invalid credentials (401), unknown value → reject as invalid credentials (401, never `ValueError`/500).
2. Consider removing the `ActorType.USER` default from `Identity.actor_type` so new code must state the actor type explicitly.
3. Add tests: missing claim rejected, junk claim rejected with 401 (not 500), all real mint paths still verify.

## Rollout constraint

[...]

PR #865, opened 2026-07-29
fix(broker): reject unknown actor_type in JWT validator, log missing claim

### LINKED ISSUE #864: Broker JWT validator defaults a missing actor_type claim to AGENT — apply the same fail-closed treatment as #862
## Context

Follow-up spun out of the reviews of #863 (which closed #862 by making the shared `verify_token` gate fail closed on missing/unknown `actor_type` JWT claims). The broker surface has its **own, independent** JWT validation path that still guesses:

```python
# src/jentic_one/broker/services/auth/token_validation.py (JwtTokenValidator.validate)
actor_type_raw = claims.get(Claim.ACTOR_TYPE, ActorType.AGENT.value)
...
actor_type=ActorType(str(actor_type_raw)),
```

Two parallels with the fixed gate:

1. **Fail-open default**: a validly-signed JWT without an `actor_type` claim is assumed to be an AGENT. Unlike the shared gate's old USER default this is the *least*-privileged interpretation, so the severity is lower — but it is still an untrusted boundary inferring a claim instead of requiring it.
2. **Bare `ValueError` on unknown values**: `ActorType(str(actor_type_raw))` raises `ValueError` on unrecognised strings instead of a typed rejection (same class of bug fixed in #862; check how the broker edge maps validator exceptions before deciding whether this can surface as a 5xx).

## Proposed fix

Mirror #863 in `JwtTokenValidator.validate` (and `TrustedIssuerVerifier` if it shares the claim parsing):

- Missing `Claim.ACTOR_TYPE` → typed rejection (whatever error the broker edge maps to 401), no default.
- Unknown value → same typed rejection, never a bare `ValueError`.
- Log the refusal cause (`jwt_actor_type_missing` / `jwt_actor_type_unknown`) like `shared/auth/verify.py` now does, keeping the wire response uniform.
- Audit the broker-side token producers first: confirm every JWT minted for broker consumption embeds `actor_type` before flipping the default (same precondition #857 established for the admin-secret tokens ahead of #862/#863).

## References

- #862 / #863 — the shared-gate equivalent (shipped)
- Flagged independently by two reviews of #863 (feasibility and architecture) as the last JWT boundary in the codebase that guesses a missing claim.

Manuel-Jentic:
Closes #864. Follow-up to #857 / #863.

## TL;DR (humans)

The broker's own JWT gate (`JwtTokenValidator`) had the same pattern flagged in #862: it guessed when a token's `actor_type` claim was absent or garbage. This PR makes it behave deliberately:

- **Unknown `actor_type` value** (e.g. `"superadmin"`, `"AGENT"`, `5`): **refused**. Same 401 as any other bad token, and the real reason is logged server-side with the issuer.
- **Missing `actor_type` claim**: still accepted as **AGENT** (the least-privileged interpretation) — but now **logged once per issuer/sub**, so operators can see which issuers omit the claim.

## Why not fail closed on a missing claim too? (the audit)

Issue #864 asked for a producer audit before flipping the default. The audit changed the answer:

| Gate | Producers | Verdict |
|---|---|---|
| Shared admin gate (#863) | Only our own minting paths — all carry `actor_type` since #857 | Fail closed on missing claim ✅ |
| Broker JWT gate (this PR) | **External trusted issuers** (JWKS path, `broker.jwt_verification.trusted_issuers`) + dev HS256 secret | Missing claim is a live external contract — hard-failing would break externally-issued agent tokens |

There is no published claims contract requiring `actor_type` from trusted issuers, existing tests mint broker JWTs with only `sub`/`exp`/`scopes`, and no internal jentic-one code mints JWTs for the broker path (login JWTs use a different secret and gate). So the missing-claim default stays — explicit, *least*-privileged (AGENT), and observable. Once `jwt_actor_type_missing` telemetry shows all real issuers send the claim, #867 flips this to a hard refusal.

An unrecognised *value* is different: that's a malformed token, not a contract gap, so it's refused outright (previously the enum constructor's bare `ValueError` escaped by accident; the web layer's catch-all still mapped it to 401, but the rejection is now typed-in-intent, deliberate, and logged).

## Behaviour matrix (agents: this is the contract)

| Token state | Before | After |
|---|---|---|
| `actor_type: "agent"` / `"user"` / `"service_account"` | honoured | honoured (unchanged) |
| `actor_type` absent | silent AGENT default | AGENT default + `jwt_actor_type_missing` warning (`iss`, `sub`, `outcome="defaulted_to_agent"`, deduped per issuer/sub) |
| `actor_type: <junk>` (unknown string, wrong case, empty, int, list) | accidental `ValueError` → catch-all 401 | deliberate `ValueError("jwt_actor_type_unknown")` → 401 + `jwt_actor_type_unknown` warning (`iss`, `sub`, value truncated to 64 chars) |
| Opaque / API-key / toolkit-key paths | untouched | untouched (DB rows carry `actor_type`) |

The `jwt_actor_type_missing` / `jwt_actor_type_unknown` event names match the shared gate's (#863), so one query aggregates both gates; the broker's missing-claim event is distinguished by `outcome`.

## Changes

- `src/jentic_one/broker/services/auth/token_validation.py` — explicit missing/unknown handling, structlog warnings with issuer context, per-(iss, sub) dedup of the missing-claim warning
- `tests/unit/broker/test_token_form_selection.py` — missing claim defaults to AGENT, explicit claim honoured, parametrized rejection over junk values, log-dedup assertion via `capture_logs`

## Reviews (all six passes done, all findings addressed)

- **Bugbot**: no bugs.
- **Security**: no issues — independently confirmed broker authorization is entirely scope/PBAC-based, so the AGENT default confers no privileges anywhere in the broker.
- **Feasibility**: sound — traced every `validate()` caller to a uniform 401 (never a 500) and confirmed the AGENT default is genuinely required. Fixed: warning dedup, claim-value truncation, naming alignment, junk-value test matrix.
- **Product fit**: fixed the one pre-merge must-have — the missing-claim warning now carries `iss` so operators can identify the IdP. Breadcrumb for the future hard flip filed as #867.
- **Architecture/rules**: compliant (layering, logging style, tests, commit conventions). Log event unified with the shared gate's name per its suggestion. `ValueError` kept for consistency with the file's existing contract (`deps.py` maps `ValueError` → 401); typed-error migration filed as #866. Gate-divergence rationale documented in the internal rules repo.
- **Code quality**: approve with nits — all applied (error-string/log-event alignment, non-string claim tests, truncation, `validate` docstring).

Follow-up issues filed from review findings: #866 (typed errors for the broker JWT path), #867 (flip missing-claim default to refusal once telemetry allows), #868 (pre-existing: JWT can claim an actor_type with no binding behind it).

## Rollout

No user-visible change for any well-formed token. No config, schema, or API changes. Externally-issued JWTs without the claim keep working (now with an operator-visible log trail).

## Verification

- `make lint` (ruff + mypy): clean
- `tests/unit/broker` + `tests/integration/broker`: 529 passed (2 pre-existing streaming-record failures unrelated to this diff, present on main-equivalent tree)

### DIFF
--- a/src/jentic_one/broker/services/auth/token_validation.py
+++ b/src/jentic_one/broker/services/auth/token_validation.py
@@ -22,10 +22,13 @@ from enum import StrEnum
 from typing import Protocol
 
 import jwt
+import structlog
 
 from jentic_one.broker.core.token_validation import CachedTokenValidator
 from jentic_one.shared.auth.identity import Identity
 from jentic_one.shared.models import ActorType
+
+logger = structlog.get_logger(__name__)
 
 # Algorithms we accept for the self-contained-JWT path. HS256 only for the
 # minimal PR-A2 verifier; §08 widens/locks this down (and adds asymmetric/JWKS).
@@ -97,13 +100,32 @@ class JwtTokenValidator:
         if not isinstance(sub, str) or not isinstance(exp, int | float):
             raise ValueError("jwt_missing_required_claims")
 
-        actor_type_raw = claims.get(Claim.ACTOR_TYPE, ActorType.AGENT.value)
+        actor_type_raw = claims.get(Claim.ACTOR_TYPE)
+        if actor_type_raw is None:
+            # Unlike the shared admin gate (#862), the broker's JWT producers
+            # include *external* trusted issuers with no published claims
+            # contract requiring actor_type, so a missing claim keeps the
+            # least-privileged AGENT default. Log it so operators can see
+            # which issuers omit the claim before any future hard refusal.
+            logger.warning("jwt_actor_type_missing_defaulting_to_agent", sub=sub)
+            actor_type = ActorType.AGENT
+        else:
+            try:
+                actor_type = ActorType(str(actor_type_raw))
+            except ValueError:
+                # An unrecognised value is a malformed token, not a contract
+                # gap — refuse it deliberately (#864) with the same error
+                # shape as any other invalid JWT instead of letting the enum
+                # constructor's ValueError escape by accident.
+                logger.warning("jwt_actor_type_unknown", sub=sub, actor_type=str(actor_type_raw))
+                raise ValueError("jwt_unknown_actor_type") from None
+
         scopes_raw = claims.get(Claim.SCOPES, [])
         permissions = [str(s) for s in scopes_raw] if isinstance(scopes_raw, list) else []
 
         return Identity(
             sub=sub,
-            actor_type=ActorType(str(actor_type_raw)),
+            actor_type=actor_type,
             permissions=permissions,
             expires_at=datetime.fromtimestamp(float(exp), tz=UTC),
             active=True,
--- a/src/jentic_one/broker/services/auth/token_validation.py
+++ b/src/jentic_one/broker/services/auth/token_validation.py
@@ -51,8 +51,16 @@ class Claim(StrEnum):
 
     SUB = "sub"
     EXP = "exp"
+    ISS = "iss"
     ACTOR_TYPE = "actor_type"
     SCOPES = "scopes"
+
+
+# Dedup for the missing-actor_type warning: one line per (iss, sub) instead of
+# one per request, so a steady issuer that omits the claim doesn't flood the
+# logs. Bounded — reset wholesale rather than tracking recency.
+_MISSING_ACTOR_TYPE_SEEN: set[tuple[str, str]] = set()
+_MISSING_ACTOR_TYPE_SEEN_MAX = 1024
 
 
 def looks_like_jwt(token: str) -> bool:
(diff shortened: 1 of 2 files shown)


Manuel-Jentic:
Closing without merging — direction change after review discussion.

This PR kept an AGENT default for JWTs with no `actor_type` claim (logged, telemetry-gated flip in #867). Decision: we should not support claim-less identity tokens at all — silently assigning an actor type, even the least-privileged one, is the wrong contract. The broker gate should fail closed on a missing `actor_type`, same as the shared admin gate (#862/#863).

#864 stays open as the tracking issue and now carries the audit findings and the requirements for the fail-closed implementation (issuer claims contract, e2e test updates, breaking-change rollout). The unknown-value rejection and log-context work on this branch (`fix/864-broker-fail-closed-actor-type`) is salvageable for that implementation. The broker JWT path's silent-refusal observability gap is split into its own issue.

Issue #604, opened 2026-07-15
[surface] NYT API doesn't appear in search when New York Times is searched


ash-jentic:
### What happened

NYT API doesn't appear in search when New York Times is searched 
Search is picking up exact correct name only 
And also showing loads of APIs even though I imagine they are totally unrelated. 

*(screenshot from the original report — to be re-attached)*

### Environment

- Jentic One version/commit:
- OS:
- Install method (install.sh / source / Docker):

---
*Migrated from the internal dogfooding tracker (feedback#15, originally filed 2026-07-02). Screenshots/attachments referenced in the original will be re-attached.*

ritvik-jentic:
Partially addressed by [PR #872](https://github.com/jentic/jentic-one/pull/872) (`fix(catalog): rank whole-word api_id matches above substring matches`).

That PR fixes the **general relevance/ordering** complaint here — a query word that matches a whole `api_id` token now outranks an incidental substring hit, and it stops flooding the results with unrelated entries at the top. The result *set* is unchanged (recall is identical); only ordering improves.

What it deliberately does **not** fix: the literal **"New York Times" → `nytimes.com`** case. The words "new"/"york" don't appear anywhere in the catalog data (id, vendor, path, or the NYT sub-API titles), so resolving that specific query would need an alias map or per-spec title fetches — both out of scope for this change.

Note: this was also worked on locally in a separate working directory; flagging here so we don't duplicate the effort. Suggest we track the alias-map / synonym idea as a follow-up if the NYT-by-friendly-name case is still wanted.

Manuel-Jentic:
Closing as a duplicate in favor of #899, which tracks the remaining work canonically.

The ranking half of this report was fixed by [PR #872](https://github.com/jentic/jentic-one/pull/872) (whole-word `api_id` matches rank above substring hits), and a live check on v0.28.0 today confirms the reported symptom is gone: `jentic catalog search "new york times"` now surfaces `nytimes.com/times_tags` first (because "times" is a whole token of the id).

What remains is exactly #899's ask: a proper alias/synonym map so human names resolve even when *no* query word appears in the `api_id` ("New York Times" → `nytimes`, "Google Analytics" → `analyticsdata`). Today's scorer (`score_entry()` in `registry/services/catalog/manifest_builder.py`) matches `api_id` tokens only — the NYT case works by lexical luck, not by design. Consolidating there.

PR #872, opened 2026-07-29
fix(catalog): rank whole-word api_id matches above substring matches

ritvik-jentic:
## Summary

Catalog search (`GET /catalog?q=…`) ranked poorly: an exact-name match tied with an incidental substring hit and fell back to **alphabetical** order, so searching `stripe` surfaced `soundstripe.com` above `stripe.com`, and `box` surfaced `Mapbox` above `box.com`. The old scorer awarded a flat `1.0` per query token that appeared *anywhere* as a substring of the `api_id`, so for a single-word query every match tied and the order collapsed to the `api_id` alphabetical tie-break.

This regrades match **quality** without ever sacrificing match **coverage**: the score encodes `(coverage, whole-token hits)` as a single integer — `coverage * (len(q_tokens)+1) + whole` — so covering more of the query always outranks covering less of it (exactly the old scorer's ordering), and *within* a coverage class, whole-token matches (query "stripe" IS the id token "stripe") outrank substring-only matches ("box" buried in "mapbox"). Across the real ~6,300-entry public catalog this fixes the top result for hundreds of single-word queries (`stripe`, `box`, `google`, `openai`, `mysql`, `nasa`, `plex`, …) with **zero ordering contradictions vs the old scorer** (empirically verified). **Which** entries appear is unchanged — only the order — so recall is identical (guardrail test).

Because this scorer backs the shared `GET /catalog` endpoint, the ranking improvement lands on **three surfaces**: the Discover page, the credentials "Choose an API" picker's catalog results, and the CLI's `jentic catalog search` (which the agent skill routes discovery through). (The picker's separate "in your workspace" list uses a client-side substring filter over `GET /apis` and is unaffected.)

**Key decisions**

- **Coverage dominates quality.** An earlier revision used flat 3:1 points, which let one whole-token hit (3) beat full two-token substring coverage (2) — burying `googleapis.com` behind ~978 `*-api` entries for "google api". The review board caught it on real data; the encoding above makes that class of regression structurally impossible (the quality digit can't carry into the coverage digit).
- **Scoped to the `api_id` only.** Indexing the manifest's human titles was rejected for this fix — and note it would **not** fix the "New York Times" case anyway (no NYT sub-API title contains "New York"). The durable fix is an alias/synonym map — tracked in #899.
- **The literal "New York Times" → nytimes.com case is NOT durably fixed.** Amusingly, `nytimes.com/times_tags` now ranks #1 for that exact query — but only because that one sub-API id contains the standalone token "times" (tokenisation luck, fragile against renames). Hence **Related to #604**, not Closes: the friendly-name capability #604 asks for is #899's job.
- **Recall is a hard guardrail.** The membership filter (`s > 0`) is untouched, so the result set is exactly as before; only scoring/ordering changed — locked in by a dedicated test.
- **Integer scores** give clean, exact tier ordering and cursor round-trips. The keyset-cursor decode boundary stays `float | None` so any historical cursor value still round-trips; non-finite values (`NaN`/`Infinity`) are now rejected at decode (400) since we only ever mint finite ints.
- **No cursor versioning.** Paginating a catalog search *across a backend restart* may show one empty page (old cursor vs new scores); it self-heals on the next search/refresh — negligible on a self-hosted single-operator tool, so skipped.

## Related issue

Related to #604 (ranking/relevance half; the friendly-name/alias half is #899)

## Changes

- `src/jentic_one/registry/services/catalog/manifest_builder.py`: coverage-first integer scoring (see above); query tokens deduplicated (order-preserving) — duplicates scaled every entry uniformly at double the scan cost.
- `src/jentic_one/registry/web/routers/catalog.py`: `q` capped at `max_length=500` (parity with the preview endpoint) — the scorer is synchronous CPU work in an async handler and an unbounded `q` was a cheap event-loop-blocking primitive. Spec artifacts regenerated.
- `src/jentic_one/shared/pagination.py`: `decode_catalog_cursor` rejects non-finite scores.
- Tests: whole-word-beats-substring pins (#604 cases), coverage-dominance pin ("google api"), mutual-exclusivity pin (whole+substring on the same id counts once), duplicate-token dedupe, empty `api_id`, recall guardrail, non-finite-cursor rejection, keyset-paging walks updated to the new scores.

## Review-board hardening

Four independent lenses (correctness+security, code quality+architecture, product fit, ranking/relevance QA with real-data probes) reviewed the rebased branch. The ranking-QA lens found the 3:1 multi-word regression (fixed as above); the other lenses' findings are folded in: `q` length cap, cursor NaN/Infinity rejection, mutual-exclusivity test, dedupe, docstring corrections. Verified across lenses: recall equivalence (proven), keyset paging never skips/repeats incl. adversarial cursors, no injection surface, endpoint authenticated, cursor opaque to SPA/CLI, correct architectural seam (catalog scoring is not part of the operations SearchStrategy registry).

## Testing

- `ruff` + `ruff format` + `mypy` clean; detect-secrets clean
- `pytest tests/unit/registry tests/unit/shared` — 760 passed
- `pytest tests/arch` — 262 passed (after `make openapi` regen; zero further drift)
- `make test-integration-sqlite` — 541 passed
- Empirical probe over the real 6,310-include catalog: all headline single-word queries verified fixed; multi-word queries ("google api", "drop box", "sound stripe", "new york times") verified regression-free; recall identical old-vs-new for every probed query; zero coverage-ordering contradictions vs the old scorer
- `tests/web/registry/**` (Postgres) — relies on CI

## Checklist

- [x] Docs updated where relevant (ranking lands in the changelog via the `fix(catalog)` commit; no user-facing doc describes ordering semantics)

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -371,7 +371,7 @@
         "filename": "openapi/control/control.openapi.yaml",
         "hashed_secret": "5ccd78467ff52d61471c0cf0f32543d0d5d14367",
         "is_verified": false,
-        "line_number": 16326
+        "line_number": 16327
       }
     ],
     "release-please-config.json": [
@@ -1070,7 +1070,7 @@
         "filename": "tests/integration/registry/ingest/test_ingestor_openapi.py",
         "hashed_secret": "90bd1b48e958257948487b90bee080ba5ed00caa",
         "is_verified": false,
-        "line_number": 69
+        "line_number": 72
       }
     ],
     "tests/integration/shared/auth/test_verify.py": [
@@ -1618,5 +1618,5 @@
       }
     ]
   },
-  "generated_at": "2026-07-31T08:59:32Z"
+  "generated_at": "2026-07-31T10:31:03Z"
 }
--- a/openapi/control/control.openapi.yaml
+++ b/openapi/control/control.openapi.yaml
@@ -9750,7 +9750,8 @@ paths:
         required: false
         schema:
           anyOf:
-          - type: string
+          - maxLength: 500
+            type: string
           - type: 'null'
           title: Q
       - in: query
--- a/src/jentic_one/registry/services/catalog/manifest_builder.py
+++ b/src/jentic_one/registry/services/catalog/manifest_builder.py
@@ -134,39 +134,56 @@ def _tokenise(text: str) -> list[str]:
     return re.findall(r"[a-z0-9]+", (text or "").lower())
 
 
-#: Points for a query token that equals a whole ``api_id`` token (e.g. query
-#: "stripe" vs the id token "stripe" in ``stripe.com``) — the strongest signal.
-_WHOLE_TOKEN_POINTS = 3
-#: Points for a query token that is only a substring of an ``api_id`` token
-#: (e.g. query "box" buried inside "mapbox") — a weak, incidental match.
-_SUBSTRING_POINTS = 1
+#: Encoding of the two ranking signals into one sortable integer (the cursor
+#: carries a single number). ``coverage`` — how many query tokens matched at
+#: all — is the dominant signal and preserves the pre-existing recall-fraction
+#: ordering exactly: matching more of the query always outranks matching less
+#: of it, no matter how "strong" the individual hits are (so a multi-word query
+#: like "google api" still ranks ``googleapis.com`` — both words covered —
+#: above the hundreds of ids that merely contain the whole word "api").
+#: ``whole`` — how many of those matches are whole ``api_id`` tokens rather
+#: than substrings buried in longer tokens — only breaks ties *within* a
+#: coverage class (the #604 fix: query "stripe" ranks ``stripe.com``, a whole
+#: token, above ``soundstripe.com``, a substring).
+#:
+#: ``score = coverage * (len(q_tokens) + 1) + whole`` is order-isomorphic to
+#: the tuple ``(coverage, whole)`` because ``whole <= coverage <= len(q_tokens)
+#: < len(q_tokens) + 1`` — the quality digit can never carry into the coverage
+#: digit. Scores are only ever compared within one query (the cursor is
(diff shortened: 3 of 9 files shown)


Manuel-Jentic:
Continuing this PR while Ritvik is away (same process as #850).

**Since `920f638`:**
- Rebased onto current main (clean) and ran a 4-lens review board (correctness+security, code quality+architecture, product fit, ranking/relevance QA) with **empirical probes over the real public catalog** (6,310 includes / 5,980 unique ids).
- The ranking-QA lens found a real multi-word regression in the 3:1 point scheme: one whole-token hit (3) beat full two-token substring coverage (2), so **"google api" buried `googleapis.com` at rank ~979** behind the `*-api` entries, and "drop box" / "sound stripe" inverted. Fixed in `2ae5a2a` with coverage-first scoring (`score = coverage * (len(q_tokens)+1) + whole`) — a strict refinement of the original scorer's ordering, so multi-word regressions of that class are structurally impossible. Re-probed on the real catalog: every single-word win kept, all three regressions gone, recall identical, zero ordering contradictions vs old.
- Also folded in from the board: `q` capped at `max_length=500` on `GET /catalog` (unbounded q = ~3s of event-loop-blocking CPU, measured), `decode_catalog_cursor` now rejects `NaN`/`Infinity` (crafted-cursor hardening), query-token dedupe, a mutual-exclusivity test pin, and spec artifact regen.
- **`Closes #604` → `Related to #604`**: the PR body itself called the headline "New York Times" case out of scope, so auto-closing contradicted it. (Fun fact from the probe: `nytimes.com/times_tags` now ranks #1 for that exact query — but only by tokenisation luck, one sub-API id happens to contain the whole token "times".) The durable friendly-name capability is now tracked in #899 (alias/synonym map, with corrected data on the title-indexing option).

**Validation on `2ae5a2a`:** ruff/mypy/secrets clean; 760 unit + 262 arch + 541 integration-sqlite passed; real-catalog probe attached to the PR body's Testing section.

PR #873, opened 2026-07-29
feat(ui): rail day separators, proactive failure surfacing, and monitor event drill-in

### LINKED ISSUE #617: [surface] Critical issue flagged - not easy to find
### What happened

Critical issue being shown in 'monitor' section - I cannot click anywhere to further explore. 

I check 'critical' in telemetry section and nothing appears: 

I select 'error' in the telemetry section and one error surfaces (I am unsure if this is related to the 'critical' issue)

Can see where the issue was but can't click anywhere to explore this further 

overall it's likely this error is small but it is identifying itself as critical with no path for me to understand this or easily resolve it. I think the error was simply me asking for an operation that was not in the scope of the API so it failed - I asked for a weather forecast for dublin but openweathermap only does realtime info. 

agent transcript: 

can you tell me what the weather in dublin will be this weekend?
  can you tell me how many repos I have access to in the Jentic github repo?

⏺ Let me tackle both. For the weekend forecast I need a forecast operation (the imported OpenWeatherMap only has current weather), and for GitHub
  I'll query the org. Let me check what's available.

⏺ Bash(echo "=== does imported OpenWeatherMap have a forecast op? ==="
      jentic apis operations openweathermap-org/openweathermap-org/2.5 --json 2>&1 | python3 -c "impor…)
  ⎿  === does imported OpenWeatherMap have a forecast op? ===
     op_2297f12018a13bffd4f48a56fb0ddb0e1b755a0c GET /weather
     op_ac1a9286135f3d2b331e019a16e02f3bbdb3cade GET /box/city
     … +4 lines (ctrl+o to expand)

⏺ Bash(jentic apis operations github-com/github-com-api-github-com/1.1.4 --json 2>&1 | python3 -c "import sys,json; d=json.load(sys.stdin);
      ops=d.get('data',[]);
      for x…)
  ⎿  (No output)

⏺ Two issues to sort. Let me search GitHub's ops more thoroughly for the org-repos endpoint, and test whether Open-Meteo (the only forecast API)
  is reachable.

⏺ Bash(jentic search "list organization repositories" --limit 8 --json 2>&1 | python3 -c "import sys,json; d=json.load(sys.stdin);
      [print(x['operation_id'], x['method'…)
  ⎿  op_8d5708a7a52d7c333aaf55f7cdb52a0f73488102 GET https://api.github.com/orgs/{org}/repos
     op_176960bd305dc14d42713eea43ab4d4e5d848f45 GET https://api.github.com/repos/{owner}/{repo}/actions/organization-secrets
     op_af299bf3b8972a3b9281b46a2ff31bf034fd4732 GET https://api.github.com/repos/{owner}/{repo}/actions/organization-variables
     … +3 lines (ctrl+o to expand)

⏺ Bash(echo "=== can I reach Open-Meteo (forecast)? Dublin coords ==="
      jentic execute "op_27f0088a8a9bc1f3e0a7049492a8296cae9986a5" \…)
  ⎿  === can I reach Open-Meteo (forecast)? Dublin coords ===
     http 403 - no_toolkit_binding

⏺ Let me handle the GitHub question (that one I can do) and count repos in the jentic org.

⏺ Bash(jentic execute "op_8d5708a7a52d7c333aaf55f7cdb52a0f73488102" \
        --path org=jentic --query per_page=100 --query type=all \…)
  ⎿  http 200
     repos returned on page 1: 100
       jentic/api-recommender (private)
     … +99 lines (ctrl+o to expand)

⏺ Page 1 returned exactly 100 repos — that means there are almost certainly more. Let me fetch page 2 to get the full count.

⏺ Bash(for pg in 2 3; do
      jentic execute "op_8d5708a7a52d7c333aaf55f7cdb52a0f73488102" \…)
  ⎿  page 2: 27 repos (3 public)
     page 3: 0 repos (0 public)

⏺ Both answered — one fully, one blocked. Here's where each stands.

  GitHub: repos you can access in the jentic org ✅

  You have access to 127 repositories in the jentic organization (100 on page 1 + 27 on page 2, page 3 empty).

  Breakdown by visibility:
  - ~19 public (e.g. jentic-public-apis, jentic-sdks, standard-agent, arazzo-engine, jentic-one, jentic-docs…)
  - ~108 private — you can see these because your PAT is authenticated as a member of the org (api-catalogue-content, jentic-cloud, core,
  runtime, the infrastructure-* set, etc.)

  The fact that private repos show up confirms your token has org membership and the repo scope working correctly.

  Dublin weekend forecast ❌ (blocked — and it's a real limitation)

  I couldn't get this one. Two problems stacked:

  1. The OpenWeatherMap API you set up only has current weather — the imported v2.5 spec has no multi-day forecast operation (just /weather,
  /find, /group, /box/city). So it can tell you Dublin right now, but not the weekend.
  2. The only forecast API in the registry is Open-Meteo (keyless, 7-day forecast) — but it returned 403 no_toolkit_binding. Open-Meteo isn't
  served by the toolkit I'm bound to, so I can't call it.

  To get the weekend forecast, pick one:

  - Easiest: add Open-Meteo to your toolkit (tk_6a46796331fb0ab4d9e747fc). It needs no API key — if the dashboard insists on a credential, a
  dummy/empty one will do. Then I can pull a real 7-day Dublin forecast immediately (no approval needed, since it's your already-bound toolkit).
  - Alternative: OpenWeatherMap does offer a 5-day/3-hour forecast endpoint, but it's not in the imported v2.5 spec, so I'd need a spec that
  includes it — more work than just adding Open-Meteo.

  Want me to give you the exact API reference to add Open-Meteo, then fetch Dublin's weekend? I can have the forecast for you seconds after it's
  in the toolkit.

✻ Churned for 1m 6s

❯ do i need an api key for openmetro?

⏺ No — Open-Meteo needs no API key. I verified this earlier in the session: I inspected the forecast operation and it reported auth: None. Its
  free tier is completely open for non-commercial use — no signup, no token, nothing to obtain.

  The only reason my Dublin forecast call failed was 403 no_toolkit_binding — not a credential problem. Open-Meteo simply isn't in the toolkit
  I'm bound to yet.

  The one wrinkle

  Even though Open-Meteo needs no key, Jentic's model still requires the API to be served by your toolkit. And as you've hit before, the
  dashboard may not let you add an API without attaching some credential. If that happens:

  - Add a dummy/placeholder credential (any value) just to satisfy the "bind a credential" requirement. Open-Meteo ignores it — the request
  succeeds regardless, because it doesn't check auth.

  So the steps are:
  1. Add open-meteo-com / open-meteo-com to your toolkit (tk_6a46796331fb0ab4d9e747fc).
  2. If it demands a credential, attach a throwaway one.

  Then tell me and I'll pull Dublin's weekend forecast immediately — no access request, no real key. Want to go ahead?

### Where (command or page)

_No response_

### Why it was a pain point

_No response_

### What you expected instead

_No response_

### Environment

- Jentic One version/commit:
- OS:

---
*Migrated from the internal dogfooding tracker (feedback#40, originally filed 2026-07-08). Screenshots/attachments referenced in the original will be re-attached.*

### LINKED ISSUE #671: [ui] Failed executions surface only as a feed entry — no proactive notification/badge (silent failure risk for unattended agents)
### What happened

An execution failed (`Execution failed: Upstream returned 429`, 15:21:35) and the **only** indication was an event sitting in the sidebar events feed with an **Acknowledge** button. There was **no proactive signal** of the failure — no notification, no unread/error badge, no toast, no email. I only discovered it hours later because I happened to open the sidebar for an unrelated reason.

### Where (command or page)

Sidebar → events feed (Execution event, `type: error`). The `CRITICAL / ERROR / WARNING` filters exist, but they only help *after* you've decided to go looking.

### Why it was a pain point

The whole point of the product is agents executing on your behalf — often unattended. A failure you can only find by manually scanning a feed is a failure you will usually miss. The **Acknowledge** button implies a triage workflow exists, but there's nothing to *bring you into* that workflow — you have to already be watching. Concretely for a scheduled/unattended agent (e.g. a nightly job): if the run fails (expired token, upstream 429, a bad credential), the work silently doesn't happen and the operator has no way to know until they next happen to glance at the sidebar. That's the difference between "my digest is late" and "my digest has been broken for a week and I didn't notice."

### What you expected instead

Some proactive surfacing of ERROR/CRITICAL executions, at least one of:
- a persistent unread/error **badge or count** on the sidebar/app icon that stays until acknowledged,
- an in-app **toast/notification** when an execution fails while the app is open,
- an optional **push channel** (desktop notification, email, or webhook) for failures — essential for unattended/scheduled runs where no one is watching the UI.

At minimum, failed executions should be visually unmissable (a persistent error indicator) rather than one line among many in a scrollable feed.

### Environment
- Jentic One version/commit: `main` @ `4098999`
- OS: macOS (Darwin 25.5.0)
- Install method: `install.sh` (docker mode, postgres)

---
_Migrated from internal Jentic One dogfooding feedback (original ref #33). Suggested classification: type:pain-point | area surface:ui | severity:major. Labels omitted per CONTRIBUTING; the intake assistant applies them._

### LINKED ISSUE #705: [ui] Monitor's Agent rail shows time-of-day only — on multi-day windows, days are indistinguishable and the ordering looks broken
## What happened

On **Monitor → Events** with the window set to **All**, the Agent rail (right-hand live feed) renders every event with a bare `HH:MM:SS` timestamp — no date, no day separators. The main event list on the left uses relative times ("10m ago", "4d ago"), so it doesn't have this problem; the rail does.

With a window spanning several days, this has two effects:

**1. Events from different days are indistinguishable.** In my rail right now, `Credential cred_*** 14:04:31` sits directly below entries stamped `10:20:03`. The `14:04:31` events are `credential.expired` events from **four days ago** (stored `2026-07-16T13:04:31Z`); the `10:20:03` ones are from **today**. Nothing visually separates them.

**2. The list looks mis-sorted.** The rail is correctly ordered newest-first by actual datetime, but with dates hidden the displayed times read (top to bottom): `10:20:03 → 10:19:38 → 14:09:23 → 14:08:46 → 14:04:31`. Times go down, then jump up — it looks out of order unless you already know there's an invisible day boundary between `10:19:38` (today) and `14:09:23` (last Thursday).

## Why this matters

This bit me during a real incident investigation. While debugging what looked like credential loss (see #665), I read the `14:04:31` credential events in the rail as *today at 14:04* and folded them into today's incident timeline — they were actually last Thursday's OAuth-setup noise. An event feed is exactly the surface people scan while investigating something going wrong; a timestamp that silently drops the date turns a monitoring aid into a source of wrong conclusions.

## Suggestion

Any of these would fix it (roughly in order of preference):

1. **Day separators in the rail** ("Today", "Thu 16 Jul") when the window spans more than one day — keeps the compact `HH:MM:SS` while making boundaries visible.
2. **Relative timestamps** ("10m", "4d") matching the main list's convention, with the full datetime on hover.
3. At minimum, a **full datetime tooltip** on hover so the date is discoverable at all.

_(Screenshots showing the mixed-day rail to be attached in a comment.)_

## Environment

- Jentic One local Docker install, `jentic-one@main` (commit 6131e3e), macOS
- Monitor → Events, Window = All, "Needs action" filter; Agent rail expanded
- Timezone UTC+1 (Ireland) — rail renders stored UTC as local time-of-day

ritvik-jentic:
## Summary

Three related observability improvements around the agent rail and the Monitor Events tab:

1. **Date visibility in the rail (#705)** — day separators ("Today", DST-safe "Yesterday", weekday+date) between feed days, and an instant full-date/time tooltip on every row timestamp (new shared portaled `Tooltip` primitive with keyboard focus support, Escape dismissal, and viewport-edge clamping).
2. **Proactive failure surfacing (#671)** — a persistent unacknowledged-failure count pill on the rail header (window-scoped to the loaded feed, labelled "in recent activity"), and error/critical events always toast — even when the toast scope is `off` (the option is labelled "Off (failures still toast)" so the override is disclosed). Assertive screen-reader announcements coalesce under failure bursts.
3. **Critical-event drill-in (#617)** — rail deep-links open the Monitor sheet (fixes a real param-vocabulary mismatch on main: the rail emitted `trace`/`execution`/`job` while Monitor read `trace_id`/`execution_id`/`job_id`, so the sheet never opened), Events rows drill to trace/execution, and an independent severity filter (critical and error separately toggleable).

## Related issues

Closes #705
Closes #671
Closes #617

## Review-board hardening (2026-07-31)

A 3-lens review board (semantic drift vs main's dashboard/agents/toolkits rebuilds, UI/UX + a11y, product fit + hygiene) returned unanimous approve-with-nits; fixes applied on top:

- **Cross-surface acknowledgment sync**: the rail's acknowledge now invalidates the Monitor events root + dashboard root, and Monitor's acknowledge flips the rail's in-memory copy via a provider-optional stream hook — so the failure pill and the Events tab can't show contradictory acked states (the SSE watermark poll never re-delivers an event on an ack flip).
- **Tooltip a11y (WCAG 1.4.13)**: Escape dismissal (keyboard- and hover-opened), viewport-edge clamping (the rail hugs the right edge — worst case for a centered bubble), `aria-describedby` appended rather than clobbered, `whitespace-normal` default.
- **Honest labels**: pill count is window-scoped and says so; toast "Off" discloses the failure override.
- **ARIA correctness**: day separators drop the prohibited `aria-label` on `role="presentation"`.
- **Toast hygiene**: assertive-announcement coalescing under bursts; bounded dismissed-id memory.
- Monitor events query key now derives from the shared cross-module root; Events drill-in reuses the shared HAL-link parser; POSIX locale ids fixed in the vitest env.

## Testing

- `eslint`, `tsc --noEmit` clean; 958 unit tests pass (browser-mode vitest incl. axe checks); 24 mocked Playwright e2e pass.
- New/updated tests: DST-boundary "Yesterday" (real CDP timezone override), malformed-timestamp separators, failure-pill show/clear + label, dismissed-toast persistence across scope flips, TTL re-toast, Tooltip keyboard focus + `aria-describedby` stability + interactive-child tab stops + **Escape dismissal** + **describedby append**, severity-filter independence + URL hydration, drill-in id resolution from `_links`.

## Checklist

### DIFF
--- a/ui/src/modules/monitor/__tests__/MonitorPage.test.tsx
+++ b/ui/src/modules/monitor/__tests__/MonitorPage.test.tsx
@@ -503,4 +503,158 @@ describe('Monitor inter-linking', () => {
 		// The 24h window is sent as the SSE `since` lower-bound.
 		expect(streamUrl!.searchParams.get('since')).toBeTruthy();
 	});
+
+	it('Events row → clicking a failed event opens its trace detail sheet (#617)', async () => {
+		const user = userEvent.setup();
+		renderMonitor('/app/monitor?tab=events');
+
+		// evt_1 is a failed execution event carrying trace_bbbbbbbb; the row must
+		// be clickable through to the trace sheet (the whole point of #617).
+		const row = await screen.findByRole('link', {
+			name: /Open execution for Execution failed: github-api/,
+		});
+		await user.click(row);
+
+		await waitFor(() => {
+			expect(currentParams().get('trace_id')).toBe('trace_bbbbbbbb');
+		});
+		// The trace sheet opens for that trace.
+		expect(await screen.findByRole('dialog')).toBeInTheDocument();
+	});
+
+	it('Events row → an event with no execution/trace is not clickable (#617)', async () => {
+		// evt_2 (import.completed) carries only trace_aaaaaaaa via its top-level
+		// trace_id — so it IS clickable; assert the acknowledged/info evt_2 with a
+		// stripped trace is inert by pointing the events feed at a payload with no
+		// trace at all.
+		worker.use(
+			http.get('/events', () =>
+				HttpResponse.json({
+					data: [
+						{
+							_links: { self: '/events/evt_x' },
+							acknowledged: false,
+							acknowledged_at: null,
+							acknowledged_by: null,
+							created_at: new Date().toISOString(),
+							data: {},
+							detail: 'A configuration warning with no execution.',
+							event_id: 'evt_x',
+							requires_action: false,
+							severity: 'warning',
+							summary: 'Config drift detected',
+							trace_id: null,
+							type: 'config.drift',
+						},
+					],
+					has_more: false,
+					next_cursor: null,
+				}),
+			),
+		);
+		renderMonitor('/app/monitor?tab=events');
+
+		await screen.findByText('Config drift detected');
+		// No clickable row was rendered for an event with no drill-in id.
+		expect(
+			screen.queryByRole('link', { name: /Open execution for Config drift detected/ }),
+		).not.toBeInTheDocument();
+	});
+
+	it('rail-style deep-link (execution_id) on the Events tab opens the sheet (#617)', async () => {
+		// The rail's "View execution" now emits the underscore vocabulary; landing
+		// on the Executions tab with it must open the sheet.
+		renderMonitor('/app/monitor?tab=executions&execution_id=exec_2');
+		expect(await screen.findByRole('dialog')).toBeInTheDocument();
+	});
+});
+
+/**
+ * Events-tab severity filter (#617). The Events tab previously had no way to
+ * filter by severity — a "critical" flag elsewhere led to a filter that showed
+ * nothing. The backend honours a repeatable `severity=` param; these lock in
+ * that the chips drive it and that error/critical stay independent values.
+ */
+describe('Monitor Events severity filter', () => {
+	beforeEach(() => {
+		setToken('mock-access-token');
+		worker.use(...monitorHandlers);
+	});
+
+	function currentParams() {
+		return new URLSearchParams(screen.getByTestId('location-search').textContent ?? '');
+	}
+
+	it('selecting Error narrows the feed and writes the severity param', async () => {
+		const user = userEvent.setup();
+		renderMonitor('/app/monitor?tab=events');
+
+		// Both fixture events render first (error + info).
+		await screen.findByText('Execution failed: github-api');
+		await screen.findByText('Import completed');
+
+		await user.click(screen.getByRole('button', { name: 'Error' }));
+
+		// The URL carries severity=error and the info event drops out.
+		await waitFor(() => {
+			expect(currentParams().get('severity')).toBe('error');
+		});
+		await waitFor(() => {
+			expect(screen.queryByText('Import completed')).not.toBeInTheDocument();
+		});
+		expect(screen.getAllByText('Execution failed: github-api').length).toBeGreaterThanOrEqual(
+			1,
+		);
+	});
+
+	it('critical and error are independent chips (selecting Critical hides an error event) (#617)', async () => {
+		const user = userEvent.setup();
+		renderMonitor('/app/monitor?tab=events');
+		await screen.findByText('Execution failed: github-api');
+
+		// The fixture's failure is severity "error"; selecting ONLY Critical must
+		// therefore hide it — proving the chips don't silently coalesce (the exact
+		// confusion #617 reported). Both remain selectable together for "failures".
+		await user.click(screen.getByRole('button', { name: 'Critical' }));
+		await waitFor(() => {
+			expect(currentParams().get('severity')).toBe('critical');
+		});
+		await waitFor(() => {
+			expect(screen.queryByText('Execution failed: github-api')).not.toBeInTheDocument();
+		});
+
+		// Adding Error back brings the failure in; the param preserves canonical order.
+		await user.click(screen.getByRole('button', { name: 'Error' }));
+		await waitFor(() => {
+			expect(currentParams().get('severity')).toBe('critical,error');
+		});
+		await waitFor(() => {
+			expect(
+				screen.getAllByText('Execution failed: github-api').length,
+			).toBeGreaterThanOrEqual(1);
+		});
+	});
+
+	it('hydrates the selected severity from the URL and clears it', async () => {
+		const user = userEvent.setup();
+		renderMonitor('/app/monitor?tab=events&severity=info');
+
+		// Only the info event survives an initial severity=info deep-link.
+		await screen.findByText('Import completed');
+		expect(screen.queryByText('Execution failed: github-api')).not.toBeInTheDocument();
+		expect(screen.getByRole('button', { name: 'Info' })).toHaveAttribute(
+			'aria-pressed',
+			'true',
+		);
+
+		await user.click(screen.getByRole('button', { name: 'Clear' }));
+		await waitFor(() => {
+			expect(currentParams().get('severity')).toBeNull();
+		});
+		await waitFor(() => {
+			expect(
+				screen.getAllByText('Execution failed: github-api').length,
+			).toBeGreaterThanOrEqual(1);
+		});
+	});
 });
--- a/ui/src/modules/monitor/__tests__/MonitorPage.test.tsx
+++ b/ui/src/modules/monitor/__tests__/MonitorPage.test.tsx
@@ -561,6 +561,55 @@ describe('Monitor inter-linking', () => {
 		).not.toBeInTheDocument();
 	});
 
+	it('Events row → a failure with only a _links.execution (no trace) opens by execution_id (#617)', async () => {
+		// The exact real-world case that regressed: the backend surfaces the
+		// linked execution ONLY as `_links.execution` (= /executions/{id}) — it is
+		// NOT stamped into `data` and there is no usable trace. The row must still
+		// be clickable and open the sheet by execution_id.
+		const user = userEvent.setup();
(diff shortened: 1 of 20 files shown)


Manuel-Jentic:
Handover status (taking this over from Ritvik):

- Rebased onto current `main` (clean — no conflicts despite main's dashboard/agents/toolkits rebuilds; all integration points re-verified semantically, not just textually).
- Ran a 3-lens review board (semantic drift, UI/UX + a11y, product fit + hygiene): unanimous approve-with-nits, no blockers. The drift lens confirmed the deep-link vocabulary, sheet props, toast path, and DST logic all line up with post-rebuild main, and that the rail→Monitor param mismatch this PR fixes is real on main.
- Hardening applied on top (`2c8537b`), highlights:
  - **Cross-surface ack sync** — rail ack ↔ Monitor Events tab ↔ failure pill can no longer show contradictory states (invalidations + a provider-optional stream hook; the SSE watermark poll never re-delivers on an ack flip).
  - **Tooltip WCAG 1.4.13** — Escape dismissal + viewport-edge clamping (the rail's right-edge timestamps were the worst case), describedby append, wrap-friendly default.
  - **Honest labels** — pill count scoped "in recent activity" (it reads the loaded feed window); toast "Off" now says "(failures still toast)" instead of silently overriding the operator's setting.
  - ARIA fix on day separators (prohibited aria-label on role="presentation"), assertive-toast coalescing under failure bursts, bounded dismissed-id memory, shared HAL-link parser reuse, POSIX locale ids in vitest env, plan-item refs stripped.
- DCO sign-offs added to all commits; PR title/body corrected (was "Two improvements… Both land on the rail" while closing three issues).
- Filed #907: #617's critical-vs-error classification mismatch is an upstream severity-assignment question a UI filter can't fix; #873 delivers the discoverability half.
- Local: eslint + tsc clean, 958 unit (incl. axe), 24 e2e — all green.

PR #879, opened 2026-07-30
fix(web): serve SPA shell for dotted deep-link navigations

### LINKED ISSUE #647: [ui] Deep links into the app 404 as raw JSON — navigating directly to /app/workspace/{vendor}/{api}/{version} returns {"detail":"Not Found"}
### What happened

Navigating directly to an in-app URL (e.g. `/app/workspace/posthog-com/posthog-api/1.0` — a URL the app itself puts in the address bar when you browse there) returns a raw JSON `{"detail":"Not Found"}` instead of the app. Client-side routes work only when reached by clicking through from `/app/workspace`; the server doesn't serve the SPA shell for sub-routes.

### Why it matters beyond annoyance

Links are how work gets shared and resumed: an operator bookmarking an API detail page, a teammate pasting "look at this toolkit" into Slack, and — critically — **the `approve_url` pattern the CLI prints for access requests** all depend on deep links resolving. If direct navigation 404s, every emitted/shared URL into the app is broken, including the one the agent hands the human for the core approval flow.

### Suggested fix
Standard SPA catch-all: serve the app shell for all `/app/*` paths and let the client router resolve (keep JSON 404s for the API namespace only).

### Environment
Jentic One self-hosted, 2026-07-14. Repro: paste any `/app/workspace/{vendor}/{api}/{version}` URL into a fresh tab while logged in.

---
*Migrated from the internal dogfooding tracker (feedback#85, originally filed 2026-07-14). Screenshots/attachments referenced in the original will be re-attached.*

ritvik-jentic:
## Summary

Fixes a backend-only bug where browser navigations to a versioned API deep-link
(e.g. `/app/workspace/<vendor>/<name>/1.0`) returned raw `{"detail":"Not Found"}`
404 instead of booting the SPA. FastAPI's `app.frontend(fallback="auto")` treats
any `/app/*` path whose final segment has a file extension as a static-file
request, and API version segments contain dots (`1.0`, `2.1.3`), so the route was
mis-classified as a missing file. Bookmarks and shared/copied links to versioned
API pages broke for anyone opening them fresh. (The CLI-emitted `approve_url` is
an API-namespace path handled by the #813 login redirect, not this shim.)

## Related issue

Closes #647

## Changes

- Add `_SpaNavigationFallbackMiddleware`, an innermost ASGI shim that swaps a
  404 for the SPA shell (200 + `index.html`) **only** for a genuine GET/HEAD
  browser navigation under `/app/`; everything else passes through untouched so
  `app.frontend()` stays authoritative for real files (304/206/Range/symlinks).
- Decide asset-vs-route by MIME type (recognized static-file extension via
  `mimetypes`), not a hardcoded prefix list, so a missing/renamed asset anywhere
  in the bundle (`assets/…` or a mount-root file like `broker-openapi.json`)
  still 404s rather than silently booting the app.
- Register the shim as the innermost user middleware so the rescued shell flows
  back out through request-id/telemetry (carries `x-request-id`, logged as the
  real 200, not the inner 404).
- Harden: RFC 7230-combine split `Accept` header lines; strip `root_path` on a
  segment boundary; strip `Range`/`If-Range` from the rescued shell (full 200,
  never 206/416); clean 404 (not 500) if `index.html` is missing mid-redeploy;
  idempotent mount that raises `RuntimeError` if the middleware stack is
  already built.
- Review-board hardening (2026-07-31): rescued shell is
  `Cache-Control: no-store` (the same URL answers 404 to non-navigation
  clients and the rescue path never revalidates, so keep shared caches out);
  dot-segment `/app` paths (`/app/../x.1`, encoded `%2e%2e`) are refused —
  never rescued; packaging smoke now exercises the dotted deep-link and the
  no-store/404 split against the installed wheel.

## Testing

- `ruff check`, `ruff format --check`, `mypy` — clean.
- `uv run pytest tests/unit` → 2142 passed; `tests/arch` → 262 passed;
  `make test-integration-sqlite` → 541 passed.
- New tests: versioned deep-link boots the shell (`1.0`, `2.1.3`); missing
  asset (incl. root-level `.json`/`.svg`) still 404s; `Range` on the shell → full
  200; rescued shell carries `x-request-id`; split `Accept` still boots;
  rescued shell is `no-store` while the framework-served shell keeps framework
  caching; dot-segment paths 404; missing `index.html` mid-redeploy → 404.
- `make smoke-packaging` (build UI + wheel, install in clean venv): dotted
  deep-link `/app/workspace/stripe-com/stripe-com-api/1.0` serves the packaged
  shell with `no-store`, stays 404 for JSON clients, missing recognized assets
  404 — PASS.
- Red-team probes against the real combined app: 401/403 and API 404s pass
  through untouched (`/api/...` never rescued), `//app`, `%00`, backslash and
  non-`/app` paths never rescued, HEAD returns empty body with correct
  headers, websocket scopes ignored.

## Checklist

- [x] Docs updated where relevant (module docstrings; no separate plan doc)

### DIFF
--- a/scripts/spa_packaging_smoke.py
+++ b/scripts/spa_packaging_smoke.py
@@ -112,6 +112,32 @@ def _check_serving() -> None:
     if deep.status_code != 200 or "<!doctype html" not in deep.text.lower():
         raise AssertionError("client-routed deep link did not fall back to index.html")
 
+    # The #647 case: a versioned deep-link whose final segment looks like a file
+    # extension (``/1.0``) must still boot the shell for a browser navigation
+    # (rescued by the SPA navigation fallback), and must NOT be cacheable —
+    # the same URL answers 404 to non-navigation clients.
+    dotted = client.get(
+        f"{SPA_MOUNT_PATH}/workspace/stripe-com/stripe-com-api/1.0", headers=html_headers
+    )
+    if dotted.status_code != 200 or "<!doctype html" not in dotted.text.lower():
+        raise AssertionError(
+            f"dotted versioned deep-link did not serve the SPA shell (#647) "
+            f"(got {dotted.status_code})"
+        )
+    if dotted.headers.get("cache-control") != "no-store":
+        raise AssertionError("rescued dotted deep-link shell must be Cache-Control: no-store")
+    dotted_api = client.get(
+        f"{SPA_MOUNT_PATH}/workspace/stripe-com/stripe-com-api/1.0",
+        headers={"Accept": "application/json"},
+    )
+    if dotted_api.status_code != 404:
+        raise AssertionError("dotted deep-link must stay a 404 for non-navigation clients")
+    # A missing *recognized* asset type must stay a 404 even for a navigation
+    # (a broken deploy surfaces as an error, never a silent shell boot).
+    missing_asset = client.get(f"{SPA_MOUNT_PATH}/assets/missing-xyz.js", headers=html_headers)
+    if missing_asset.status_code != 404 or "<!doctype" in missing_asset.text.lower():
+        raise AssertionError("missing recognized asset was rescued into the shell")
+
     # Anything OUTSIDE the /app namespace is unambiguously a backend call: an
     # unknown path is a true 404 for every client (no Accept-header guesswork),
     # so the SPA mount can never shadow the API.
@@ -164,8 +190,9 @@ def _check_serving() -> None:
             )
 
     print(
-        "OK  bare root redirects to /app/, SPA served under /app, "
-        "non-/app 404 preserved, root icon probes 307 into /app and resolve, "
+        "OK  bare root redirects to /app/, SPA served under /app "
+        "(incl. dotted versioned deep-links, #647), non-/app 404 preserved, "
+        "root icon probes 307 into /app and resolve, "
         "runtime config exposed, no database required"
     )
 
--- a/src/jentic_one/shared/web/static.py
+++ b/src/jentic_one/shared/web/static.py
@@ -16,7 +16,17 @@ backend call*. An unknown non-``/app`` path is a plain 404 regardless of the
 ``Accept`` header — there is no content-negotiation guesswork and no
 hand-maintained "API-owned prefix" allow-list. Within ``/app/*`` the
 ``fallback="auto"`` behaviour still applies so SPA deep-link refreshes
-(``/app/agents/123``) serve ``index.html``.
+(``/app/agents/123``) serve ``index.html``. A dedicated HTML-navigation
+fallback additionally covers *dotted* deep-links (e.g. an API-version segment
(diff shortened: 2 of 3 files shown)


Manuel-Jentic:
Handover status (taking this over from Ritvik):

- Rebased onto current `main` (`46a919a`). One textual conflict with #857's `app.state.spa_mounted` flag in `mount_spa` — resolved by keeping both (flag, then shim); no semantic overlap (the shim only rewrites 404s, the #813/#857 handler only touches 401s).
- Ran a 4-lens review board (red-team/ASGI security, correctness+architecture, product fit, code quality). No blocking findings; hardening applied on top (`dea3acf`):
  - **`Cache-Control: no-store` on the rescued shell** — the same URL answers 404 (JSON accept) or 200 HTML (browser accept) and the rescue path never honours conditional revalidation, so shared caches are kept out of it entirely. Framework-served shells keep framework caching.
  - **Dot-segment `/app` paths refused** — `/app/../x.1` (decoded `%2e%2e`) is never a legitimate SPA route; it now stays an honest 404 instead of being rescued.
  - **`RuntimeError` instead of `assert`** for the mounted-after-stack-built guard (survives `python -O`).
  - **Packaging smoke extended**: `make smoke-packaging` now exercises the dotted deep-link, the no-store header, the 404-for-JSON split, and the missing-asset 404 against the installed wheel — ran locally end-to-end: PASS.
  - PR body corrected: the CLI `approve_url` is an API-namespace path handled by the #813 login redirect, not this shim; and the dangling "plan doc" reference removed.
- Verified clean by probes: 401/403 and API 404s pass through untouched; `//app`, `%00`, non-`/app` paths never rescued; HEAD rescues with empty body and correct headers; missing `index.html` mid-redeploy degrades to a clean 404; idempotent double-mount no-ops.
- Full local suite green: 2142 unit / 262 arch / 541 integration-sqlite, ruff + mypy clean.

Manuel-Jentic:
Parking this PR for now (merge-ready, CI green) pending a design decision with Manuel.

The open question: this shim is ~206 lines of ASGI logic (plus hardening/tests) to work around FastAPI's `_is_frontend_navigation_request` hard-rejecting dotted final segments — no framework knob avoids it (even `fallback="index.html"` runs the same check), and a catch-all route can't fire because the static mount swallows all `/app/*`. The alternative is a ~30-line UI-only route change (make `:version` not the final segment, e.g. `.../:version/overview`, or encode dots) — far less machinery, but it fixes only this route: anything routed via `encodeURIComponent` (e.g. toolkit ids) preserves dots and would silently re-trigger the class of bug, and it gives up the clean URL that mirrors the backend's `/apis/:vendor/:name/:version` identity triple.

Options on the table: merge the shim (class fix), switch to the URL-scheme change (instance fix, less code), and/or push the relaxation upstream to FastAPI so the shim can eventually be deleted.

Issue #883, opened 2026-07-30
[registry][ui][cli] Detect upstream spec updates in jentic-public-apis and notify operators (update available, re-import)


Manuel-Jentic:
## Problem

Once an operator imports an API from the community catalog, the instance never learns that the upstream spec changed. jentic-public-apis moves (fixes land via overlay PRs — e.g. #21889 — and the new contribute-spec-fix / import-new-api skills will accelerate that), but local copies silently go stale: stale security schemes, stale servers, stale operations. There is no signal in the UI, the CLI, or the events feed.

## Current state (code-verified)

- Catalog manifest is lazily refreshed from `https://raw.githubusercontent.com/jentic/jentic-public-apis/main/apis/openapi/apis.json` when older than `catalog.manifest_max_age_seconds` (86400). Manifest entries carry no digest.
- "Registered" detection already matches catalog `spec_url` against non-archived revisions' `source_url` (`CatalogRepository.registered_spec_urls`).
- Active revisions store `spec_digest` (sha256 of spec bytes) and `origin` (`catalog` vs null).
- Notification plumbing exists: admin events feed + SSE + `ActionInboxBell` (shows `requires_action=true` events); `credential_expiry_scanner` is the precedent for a periodic sweep.

## Proposed design

**MVP (no new scanner):** at manifest-refresh time, issue conditional GETs (`If-None-Match`) for `registered_spec_urls` only; raw.githubusercontent ETags are content-derived, so `304` = unchanged, `200` = update. Persist `(api_id, last_seen_etag, last_notified_digest)` in a small `catalog_update_checks` table (KSUID id, unique index on local `Api.id` — never the manifest id, per D-005a). On change, emit a new `EventType.CATALOG_UPDATE_AVAILABLE` (added to the `ALL` frozenset, `requires_action=True`, deduped on digest) — feed, SSE and bell render it with no UI work.

**Fast-follow (full):**
- Periodic scanner in `shared/jobs/` (gated on registry+admin DBs, jittered batching, per-sweep cap), comparing upstream sha256 vs `spec_digest`; config `catalog.update_check_interval_seconds` (0 disables; must degrade silently when GitHub is unreachable, e.g. air-gapped installs).
[...]

Issue #768, opened 2026-07-22
cloud MCP: imported operations expose only path params — no header/query/body inputs, so every write operation is uncallable (notion.com, linear.app)


shaneslo:
@@## Summary

Every operation imported into the cloud workspace exposes **only its path parameters**. Header parameters, query parameters, and request bodies are absent from the operation schema entirely, so `execute` has no way to send them and no client can supply them.

The practical result: **every write operation is uncallable**, and on `notion.com` every operation of any kind is uncallable, because Notion's required `Notion-Version` header was dropped along with everything else.

Both credentialed APIs in my workspace are affected:

| API | Version | Operations | Callable |
|---|---|---|---|
| `notion.com/notion-api` | `2026-03-11` | 35 | **0** |
| `linear.app/main` | `1.0` | 1 | **0** |

## Reproduction (2026-07-22, cloud `https://api.jentic.com/mcp`)

9 `execute` attempts, all failed. Nothing reaches the upstream API successfully.

| Operation | Inputs sent | Result |
|---|---|---|
| `GET /v1/users/me` (`op_ed0136ae094edd02`) | `{}` | 400 `Notion-Version header should be defined, instead was undefined` |
| `GET /v1/users/me` | `{"Notion-Version": "2026-03-11"}` | 400 — identical |
| `GET /v1/users/me` | `{"headers": {"Notion-Version": "..."}}` | 400 — identical |
| `GET /v1/users` (`op_d95c940f61033c59`) | `{"page_size": 1}` | 400 — identical |
| `POST /v1/search` (`op_e11a502f266f3ce7`) | `{"query": "test", "page_size": 1}` | 400 — identical |
| `GET /v1/pages/{page_id}` (`op_86cb41f4d92a7f97`) | valid UUID path param | 400 — identical |
| `POST /v1/pages` (`op_bd43139f2a0bb0dd`) | full `parent` + `properties` body | 400 — identical |
| `POST /graphql` (`op_db6e60823c8970cb`) | `{"query": "{ viewer { id name } }"}` | 400 Apollo CSRF block |
| `POST /graphql` | `{"query": "query { __typename }", "variables": {}}` | 400 Apollo CSRF block |

Two things worth isolating:

- The `Notion-Version` failure is **universal** — GET and POST, with and without path params, with and without a body. It is not endpoint-specific.
[...]

PR #886, opened 2026-07-30
fix(ingest): resolve effective operation security op-level-else-document-level

### LINKED ISSUE #772: bug: import drops document-level security scheme → broker omits apikey, upstream returns InvalidApiKey
## Summary

`catalog import` drops the API's security scheme when the spec declares auth at the **document level** (top-level `security` + `components.securitySchemes`) rather than per-operation. The imported operation ends up with `auth`, `security`, `security_schemes`, and even `raw_spec` all `null`. As a result the broker has no scheme to bind the credential to, forwards the request **without** the required `apikey` query parameter, and the upstream rejects it with `InvalidApiKey`.

This makes any correctly-specced, document-level-secured API uncallable through Jentic even when a valid credential is bound.

## Affected API (repro case)

`bing.com/webmaster-tools` (Bing Webmaster Tools API 1.0)
Spec: https://raw.githubusercontent.com/jentic/jentic-public-apis/refs/heads/main/apis/openapi/bing.com/webmaster-tools/1.0/openapi.json

### 1. Source spec is correct — auth declared at document level

```json
// components.securitySchemes
"apiKey": {
  "type": "apiKey",
  "in": "query",
  "name": "apikey",
  "description": "Bing Webmaster Tools API key, passed as the `apikey` query parameter."
}

// top-level security (applies to every operation)
"security": [ { "apiKey": [] } ]
```

Note: the requirement is declared **globally** (top-level `security`), not on each operation. `paths./GetUserSites.get.security` is `null` — which is valid OpenAPI; operations inherit the document-level `security`.

### 2. Imported operation has no security metadata at all

After `jentic catalog import bing.com/webmaster-tools`, inspecting the imported op:

```
$ jentic inspect op_e6c0c8ffef502b5c4a47451bc4ea3faf0230bb39 --json
{
  "auth": null,
  "security": null,
  "security_schemes": null,
  "raw_spec": null
}
```

`raw_spec` being `null` is the smoking gun — the security block from the source never made it into the imported record.

### 3. Broker forwards without the key → upstream rejects

```
$ jentic execute GET:https://ssl.bing.com/webmaster/api.svc/json/GetUserSites \
    --header Jentic-Toolkit-Id=<toolkit> --broker-scheme http --broker-host 127.0.0.1:8100

status: 400
Jentic-Upstream-Status: 400
Jentic-Error-Origin: upstream
body: { "ErrorCode": 3, "Message": "ERROR!!! InvalidApiKey" }
```

`Jentic-Error-Origin: upstream` confirms the broker forwarded the request and Bing rejected it — because no `?apikey=` was attached. All Jentic-side gates (import, toolkit binding, GET permission rule) were satisfied; the request reached Bing correctly, just unauthenticated.

## Probable root cause

The importer appears to read only **operation-level** `security` and ignore the **document-level** `security` + `components.securitySchemes`. Specs that declare auth globally (a very common pattern) therefore import with no security scheme, and the broker cannot inject the credential.

## Impact

- Any API secured via a top-level `security` block is effectively uncallable through the broker — `InvalidApiKey` / unauthenticated upstream errors — regardless of a valid bound credential.
- Not recoverable from the dashboard: there is no way to hand-add a security scheme to an already-imported operation, so the user hits a dead end.
- Field renames on the credential (e.g. param name) don't help, because the scheme is missing entirely — a confusing dead-end for the user.

## Suggested fix

- On import, resolve the effective security for each operation as **operation-level `security` if present, else document-level `security`**, and carry `components.securitySchemes` into the imported record so the broker can bind the credential.
- Consider surfacing a warning at import time when a spec declares `securitySchemes` but the resulting operations end up with no resolved security (would have caught this).

## Environment

- Local install (control plane `127.0.0.1:8000`, broker `127.0.0.1:8100`), Docker Compose deployment.
- Surface: Registry / import (ingest). Downstream symptom on Broker.

---
_Filed during a dogfooding session; no credentials or personal data included._

Manuel-Jentic:
## Summary

Fixes #772 — specs that declare auth at the **document level** (top-level `security` + `components.securitySchemes`, e.g. Bing Webmaster Tools) imported with no per-operation security at all, so nothing downstream knew the operation was authenticated and the broker forwarded unauthenticated requests (`InvalidApiKey` upstream).

- `OpenAPIOperationParser` now resolves each operation's **effective** security — the operation-level requirement when the key is present (including the explicit `[]` "this operation is public" opt-out), else the document-level requirement — and retains it in `raw_operation`. Same override-else-inherit semantics as the catalog preview projector (`parse_preview_operations`), same retention pattern as the #768 parameters fix.
- Malformed `security` values (non-list values, non-dict requirement entries) are sanitized out rather than polluting `raw_operation`; a present-but-malformed op-level value overrides (to nothing) rather than silently inheriting, matching the projector.
- Import-time warning: `ExtractSecuritySchemesStage` now logs `security_schemes_unused` (warning) when the spec declares `securitySchemes` but **no** operation resolves an effective requirement — the silent state behind #772. Fed by a new `secured_operation_count` context key produced by `ExtractOperationsStage` (stage order already guarantees availability; contracts updated).

No router/response-model changes, so no `make openapi` regen needed. Operator-visible events for this warning are deliberately out of scope — the ingest pipeline has no warnings channel yet; that fits the flow-3 events work (#883).

## Tests

- Parser unit tests: doc-level inheritance, op-level override, explicit `[]` opt-out, no-security case sprouts no key, malformed-value sanitization.
- Ingest integration: doc-level requirement persisted onto every operation row; `security_schemes_unused` fires for a schemes-but-no-requirement spec and stays silent when the doc-level requirement resolves.
- Full registry unit + integration suites (520) and arch tests (262) green; ruff + mypy clean.

## Test plan

- [ ] CI green
- [ ] Re-import `bing.com/webmaster-tools` from the catalog and confirm `jentic inspect <op>` now shows the security requirement and the broker attaches `?apikey=`

Please use **squash + merge**.

### DIFF
--- a/src/jentic_one/registry/ingest/parsers/openapi.py
+++ b/src/jentic_one/registry/ingest/parsers/openapi.py
@@ -25,6 +25,7 @@ class OpenAPIOperationParser:
             spec_path_count.record(len(paths))
             logger.debug("paths_found", count=len(paths))
             operations: list[dict[str, Any]] = []
+            document_security = self._sanitize_security(spec.get("security"))
 
             for path, path_item in paths.items():
                 if not isinstance(path_item, dict):
@@ -42,7 +43,7 @@ class OpenAPIOperationParser:
                         continue
 
                     op = self._process_operation(
-                        path, method, operation, path_servers, path_parameters
+                        path, method, operation, path_servers, path_parameters, document_security
                     )
                     operations.append(op)
 
@@ -56,6 +57,7 @@ class OpenAPIOperationParser:
         operation: dict[str, Any],
         path_servers: list[dict[str, Any]],
         path_parameters: list[dict[str, Any]],
+        document_security: list[dict[str, Any]] | None,
     ) -> dict[str, Any]:
         """Process a single operation into the canonical output shape.
 
@@ -68,6 +70,15 @@ class OpenAPIOperationParser:
         they are merged in here with operation-level entries taking precedence
         on the same ``(name, in)`` key, mirroring the OpenAPI merge semantics
         already used by the catalogue preview projector.
+
+        Retains the operation's *effective* ``security`` — the operation-level
+        requirement when the key is present (including an explicit ``[]``
+        opt-out), else the document-level requirement (issue #772). Without
+        this, specs that declare auth globally (a very common pattern) import
+        with no per-operation security at all, so nothing downstream knows the
+        operation is authenticated and the broker forwards unauthenticated
+        requests. Same override-else-inherit semantics as the catalogue
+        preview projector.
         """
         operation_servers: list[dict[str, Any]] = operation.get("servers", [])
         servers = operation_servers if operation_servers else path_servers
@@ -89,7 +100,29 @@ class OpenAPIOperationParser:
         if "requestBody" in operation:
             result["requestBody"] = operation["requestBody"]
 
+        operation_security = operation.get("security")
+        effective_security = (
+            self._sanitize_security(operation_security)
+            if operation_security is not None
+            else document_security
+        )
+        if effective_security:
+            result["security"] = effective_security
+
         return result
+
+    @staticmethod
+    def _sanitize_security(security: Any) -> list[dict[str, Any]] | None:
+        """Coerce a ``security`` value into a clean requirement list.
+
+        Returns only the dict entries of a list value (each maps scheme name →
+        scopes); a malformed value (non-list, e.g. a bare string) yields
+        ``None`` so it neither pollutes ``raw_operation`` nor masks the
+        document-level requirement.
+        """
+        if not isinstance(security, list):
+            return None
+        return [req for req in security if isinstance(req, dict)]
 
     @staticmethod
     def _merge_parameters(
--- a/src/jentic_one/registry/ingest/parsers/openapi.py
+++ b/src/jentic_one/registry/ingest/parsers/openapi.py
@@ -75,10 +75,9 @@ class OpenAPIOperationParser:
         requirement when the key is present (including an explicit ``[]``
         opt-out), else the document-level requirement (issue #772). Without
         this, specs that declare auth globally (a very common pattern) import
-        with no per-operation security at all, so nothing downstream knows the
-        operation is authenticated and the broker forwards unauthenticated
-        requests. Same override-else-inherit semantics as the catalogue
-        preview projector.
+        with no per-operation security at all, so nothing downstream can tell
+        the operation requires auth. Same override-else-inherit semantics as
+        the catalogue preview projector (keep the two aligned).
         """
         operation_servers: list[dict[str, Any]] = operation.get("servers", [])
         servers = operation_servers if operation_servers else path_servers
--- a/src/jentic_one/registry/ingest/stages/extract_operations.py
+++ b/src/jentic_one/registry/ingest/stages/extract_operations.py
@@ -3,12 +3,16 @@
 from __future__ import annotations
 
 import uuid
-from typing import ClassVar
+from typing import Any, ClassVar
+
+import structlog
 
 from jentic_one.registry.ingest.parsers.openapi import OpenAPIOperationParser
 from jentic_one.registry.ingest.pipeline.ctx import PipelineContext
 from jentic_one.registry.ingest.stages.base import BasePipelineStage
 from jentic_one.registry.repos import OperationInput, OperationRepository
+
+logger = structlog.get_logger()
 
 
 class ExtractOperationsStage(BasePipelineStage):
@@ -16,7 +20,7 @@ class ExtractOperationsStage(BasePipelineStage):
 
     name: ClassVar[str] = "ExtractOperationsStage"
     _requires: ClassVar[dict[str, type]] = {"revision_id": uuid.UUID}
-    _produces: ClassVar[dict[str, type]] = {"operation_ids": set, "secured_operation_count": int}
+    _produces: ClassVar[dict[str, type]] = {"operation_ids": set}
 
     async def pre_run(self, ctx: PipelineContext) -> None:
         revision_id = ctx.require("revision_id", uuid.UUID)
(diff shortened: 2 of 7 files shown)


Manuel-Jentic:
## Review-board follow-up (3 reviewers: security, architecture, quality+tests)

Pushed fixes for the actionable findings:

- **MEDIUM (security)** — the `security_schemes_unused` warning crashed the import via `sorted(schemes)` when a hostile/malformed spec carried non-string `securitySchemes` keys (YAML allows int/bool keys). Names are now coerced to `str` before sorting, and the helper requires a real dict shape (non-dict `components`/`securitySchemes` are ignored). Added a dedicated unit test covering non-string keys and non-dict shapes.
- **MEDIUM (architecture)** — moved the warning into `ExtractOperationsStage` as a self-contained diagnostic and dropped the `secured_operation_count` context key + the extra `_requires` on `ExtractSecuritySchemesStage`. A log-only signal shouldn't strengthen a functional stage contract (which would hard-fail an otherwise-valid pipeline reorder).
- **nit (logging/docstring)** — dropped the prose `detail=` log field; softened the docstring to "downstream can't tell the operation requires auth" rather than overstating broker mechanics.
- **nit (architecture)** — added a back-reference from the catalog preview projector to the parser so the two op-else-doc implementations stay aligned.
- **tests** — added the missing matrix cells: `security: [{}]` (optional-auth requirement object, retained) and root `security: []` (not inherited).

**Deferred decisions (stated, not changed):**
- *Malformed op-level `security` fails open* (imports as no-security rather than inheriting doc-level). Deliberate and consistent with the preview projector; impact is nil today since nothing enforces per-op auth yet. Happy to flip both to fail-secure if preferred.
- *Doc-level fan-out amplification*: a bounded hostile spec repeats the doc-level requirement onto every operation row. Bounded by the existing spec-size limit; a shared/interned requirement is possible but felt like over-engineering for this fix — flagging for a call.
- *Existing revisions*: this is parse-time only; already-imported revisions keep security-less `raw_operation`s until re-imported (re-ingest repopulates via `delete_for_revision` + recreate). No migration in this PR.

All registry unit + integration + arch suites (790) green; ruff + mypy clean.

ren-jentic:
# PR #886 review — effective operation security (op-else-doc)

**Verdict: request changes — the code is good; the claim is the problem.** The parser work is correct, well-tested groundwork, but nothing in this diff can produce the outcome the PR claims ("broker attaches `?apikey=`"). Verify the #772 repro's actual causal chain before closing the issue against this change.

## What's solid

- **Parser semantics are correct per OpenAPI**: op-level `security` key present (including explicit `[]` opt-out) wins entirely, absence inherits the document level, no merging. Matches the catalog preview projector, and the cross-reference comments keep the two implementations aligned.
- **Test matrix is genuinely complete** after the review-board round: inheritance, override, `[]` opt-out, `[{}]` optional-auth retained, root `security: []` not inherited, malformed-value sanitization, no key sprouting when nothing is declared.
- **`security_schemes_unused` warning is well-placed** post-refactor (self-contained diagnostic in `ExtractOperationsStage`, no cross-stage contract coupling for a log-only signal), and the non-string-key hardening is real — YAML `1:` / `true:` keys would have crashed `sorted()`.

## Findings

### 1. BLOCKER — "Fixes #772" is not demonstrated by this diff

I traced every consumer of `raw_operation` across the codebase (and the enterprise overlay). What this PR persists is **read by nobody**:

- **The broker never sees operation security.** `CredentialService.inject()` takes `(api_vendor, api_name, api_version, identity)` — it doesn't even receive the operation. Resolution is credential-scope-driven (`CredentialRepository.list_by_vendor` → coverage → specificity), and `inject_auth` builds the apikey header/query from the **stored credential's** `field_name`/`location`. Nothing in that chain consults `raw_operation["security"]`.
- **`jentic inspect`'s auth block doesn't come from here either.** It's fed by `_load_security_schemes(operation.revision_id)` → revision-level `SecurityScheme` rows → `translate_security_schemes` (which handles `apiKey`-in-query correctly). Those rows are persisted by `ExtractSecuritySchemesStage` from `components.securitySchemes` — untouched by this PR and already working on main.
- The `raw_operation` inputs projector (`inspect/inputs.py`) reads `parameters`/`requestBody` only.

So the test-plan claim — *"confirm `jentic inspect <op>` now shows the security requirement and the broker attaches `?apikey=`"* — cannot be an effect of this diff. Whether the broker attaches `?apikey=` is a pure function of the bound credential, before and after this PR. The PR's own review-board note concedes it from the other direction: *"impact is nil today since nothing enforces per-op auth yet."* That sentence and "Fixes #772" can't both be true.

A plausible reading of the #772 evidence (`auth`, `security_schemes`, `raw_spec` all null in inspect output): the original bing import **predates `ExtractSecuritySchemesStage`**, so a plain re-import on today's main would already restore the inspect auth block — with or without this PR. If the manual bing test passes after merging, that's likely why, and attributing it to this change would enshrine a wrong causal model of the bug.

**Ask:** either (a) link the consumer-side change that makes per-op effective security load-bearing (broker gating, MCP tool schema, provisioning suggestion — whatever the intended consumer is) and re-scope this PR as "persist the data (part 1 of #772)", or (b) run the bing re-import repro on main *without* this branch first, so the issue closes against the change that actually fixes it.

### 2. LOW — warning false-positive on explicit universal opt-out

A spec that declares `securitySchemes` but where every operation *explicitly* opts out (`security: []` on each op, no doc-level default) triggers `security_schemes_unused`, even though the author said exactly what they meant. Contrived and harmless, but the docstring's "almost certainly relies on something the importer failed to carry over" is then wrong. A one-line addition to the log text ("or every operation explicitly opts out") keeps it honest.

### 3. LOW — unnamed logger

`extract_operations.py` uses `structlog.get_logger()` without `__name__`; the parser module and siblings pass a name. Keep it consistent for log filtering.

### 4. Deferred decisions — agree, with one condition

- **Fail-open on malformed op-level security** (present-but-bad → nothing, not doc-inheritance): keep as-is — consistent with the projector, and since nothing enforces per-op auth yet, fail-secure would be theater. But the eventual consumer PR must revisit this: at that point fail-open becomes a real decision, not a cosmetic one.
- **Doc-level fan-out amplification**: bounded by the spec-size limit; interning is over-engineering for now. Agree.
- **Existing revisions stay security-less until re-import**: fine to defer, but put one line in the issue on close so operators know `bing.com/webmaster-tools` needs a re-import, not just an upgrade.

PR #893, opened 2026-07-31
feat(catalog): notify when a registered API's upstream spec changes (Flow 3 MVP)

### LINKED ISSUE #883: [registry][ui][cli] Detect upstream spec updates in jentic-public-apis and notify operators (update available, re-import)

Manuel-Jentic:
## Summary

Phase-2 MVP of **Flow 3** (upstream API spec changes → we detect it and notify),
from `docs/plans/spec-flywheel-and-catalog-update-notify.md`. Closes #883.

When someone browses the catalog and the cached manifest is stale, we already
refresh it. This PR piggybacks on that refresh: right after it, we conditionally
re-fetch the spec URLs of **registered** APIs using `If-None-Match`, hash the
body, and emit a `catalog.update_available` event when the bytes actually
changed. No new background loop, no UI/CLI surfaces yet — those are later phases.

---

### What you get
- If you imported an API from the catalog and its upstream OpenAPI spec later
  changes, an event shows up in the **events feed** (`catalog.update_available`)
  with a friendly summary: *"Upstream spec updated for {vendor}/{name}
  ({version})"*.
- It fires **once per real change** (deduped on the content hash), and a spec
  that reverts back to what you have registered goes quiet again.
- **Kill switch:** set `catalog.update_check_interval_seconds = 0` (air-gapped
  installs get zero egress, zero events).

### How it behaves (and what it deliberately does *not* do yet)
- Detection is **piggybacked on catalog reads**, not a scheduler — an instance
  nobody browses won't notice changes until someone opens the catalog and the
  24h manifest cache is stale. A dedicated scanner is Phase 4.
- The event is **informational** (`requires_action=False`) for now: there is no
  one-click re-import or dismiss yet (Phase 4), so we don't want unresolvable
  items piling up in the action inbox. It still shows in the feed.
- The sweep runs **off** the user's request (fire-and-forget) and is bounded by a
  wall-clock deadline + limited concurrency, so a slow/hostile upstream can never
  stall a catalog page load.

### The flow, end to end

```mermaid
sequenceDiagram
    actor User
    participant Catalog as CatalogService
    participant Refresh as manifest refresh
    participant Sweep as update-notify sweep (background task)
    participant Up as upstream spec host
    participant RDB as registry DB
    participant ADB as admin DB (events)

    User->>Catalog: GET /catalog (authenticated)
    Catalog->>Refresh: _safe_refresh() (advisory-lock, max-age gated)
    Refresh-->>Catalog: manifest refreshed
    Catalog-)Sweep: spawn (fire-and-forget) — request returns now
    Note over User,Catalog: user response is NOT blocked on the sweep

    loop each registered spec (bounded concurrency, under a deadline)
        Sweep->>RDB: read last check row (etag/digest)
        Sweep->>Up: GET spec  (If-None-Match: last etag)
        alt 304 Not Modified
            Up-->>Sweep: 304
            Sweep->>RDB: touch last_checked_at (+ refreshed etag)
        else 200 changed bytes
            Up-->>Sweep: 200 + body
            Note over Sweep: sha256(body) vs registered digest & last-notified
            Sweep->>RDB: upsert etag/digest/notified_digest
            Sweep->>ADB: emit catalog.update_available (once)
        end
    end
```

---

### Change map
| Area | File | What |
| --- | --- | --- |
| fetch | `registry/services/catalog/fetch.py` | `fetch_bytes_conditional()` beside `fetch_json` — same SSRF/redirect/size guards (per-hop `validate_upstream_url`, now threading `config.egress`), sends `If-None-Match`, returns a 304 signal **or** raw bytes + `sha256` hex + response ETag. Both fetchers send a descriptive `User-Agent`. |
| schema | `registry/core/schema/catalog_update_checks.py` + migration `d3e4f5a6b7c8` | `catalog_update_checks` table (KSUID id, **unique `local_api_id`** = local `apis.id`, never the manifest id per D-005a). |
| repo | `registry/repos/catalog_update_check_repo.py` | `get()` / `upsert()`, flush-only-never-commit; `upsert` won't clear `notified_digest` on a no-change probe. |
| candidates | `registry/repos/revision_repo.py` | `registered_specs_for_notify()` → non-archived revisions with a `source_url`, joined to identity + current digest. |
| orchestration | `registry/services/catalog/service.py` | `_spawn_update_notify_sweep()` (offload) → `_run_update_notify_sweep()` (deadline + `Semaphore` + summary log) → `_probe_one()` (fetch, compare, upsert, emit). |
| events | `shared/models/events.py` | `EventType.CATALOG_UPDATE_AVAILABLE` added to constants **and** `ALL`. |
| config | `shared/config.py` | `update_check_interval_seconds` (default 86400, `0` = kill switch), `update_sweep_deadline_seconds` (300), `update_sweep_max_concurrency` (4, below the DB pool). |

### The change signal is the body hash, not the ETag
`sha256(body)` is the source of truth. The ETag is used **only** as a `304`
optimization (`If-None-Match`), never trusted for equality (weak ETags don't
promise byte-identity). A change notifies iff the upstream hash differs from
**both** the registered revision's digest **and** the last-notified digest —
so a revert-to-registered stops notifying, and each distinct change fires once.

```mermaid
flowchart TD
    A[conditional GET spec] --> B{304?}
    B -- yes --> T[touch last_checked_at + etag]
    B -- no --> C[sha256 body = upstream_digest]
    C --> D{upstream_digest != registered digest<br/>AND != last_notified_digest?}
    D -- no --> U[upsert etag/digest only]
    D -- yes --> E[upsert + set last_notified_digest]
    E --> F[emit catalog.update_available once]
```

### Data model

```mermaid
erDiagram
    apis ||--o| catalog_update_checks : "local_api_id (unique, no FK)"
    apis ||--o{ api_revisions : "api_id"
    catalog_update_checks {
        string id PK "KSUID cuc_*"
        uuid local_api_id UK "= apis.id"
        string spec_url
        text last_seen_etag
        string last_seen_digest
        string last_notified_digest "dedupe key"
        datetime last_checked_at
        datetime created_at
    }
```
No FK on `local_api_id` by design: the row is cheap scratch state; a dropped API
is simply ignored by the next sweep (no matching registered spec).

### Concurrency / correctness notes (honesty)
- The refresh advisory lock is **transaction-scoped**, so the sweep is **not**
  globally single-flight. A rare concurrent double-refresh can emit a duplicate
  event; the next sweep dedupes on the persisted digest. Deliberately tolerated
  (idempotent for operators), documented in code.
- Digest convention is **bare sha256 hex** on both sides (ingest computes
  `hashlib.sha256(resp.content).hexdigest()`; the probe hashes the fetched body
  the same way). An integration test exercises the real hex end-to-end so a
  future framing drift can't silently make every probe look "changed".

---

## Review board (two rounds)

**Round 1 — 6 lenses** (security, ORM/migration, correctness, architecture/rules,
testing, plan-fidelity): unanimous **APPROVE-WITH-NITS**; nits applied
(SSRF-via-redirect test, corrected "single-flight" claim).

**Round 2 — 6 lenses** (feasibility/operability, deep security, code-quality,
product-fit, best-practices research, + regression sweep). Two REQUEST-CHANGES
blockers raised and **both fixed in this PR**:

| Finding | Lens | Resolution |
| --- | --- | --- |
| Sweep awaited inline on the user read path; no aggregate bound | feasibility (MUST), security (SHOULD) | Offloaded to a fire-and-forget task; added wall-clock deadline + bounded concurrency + summary log (`d492e8e`) |
| `requires_action=True` with no resolve path → inbox fatigue | product-fit (High) | Downgraded to `requires_action=False` until Phase-4 re-import lands (`d492e8e`) |
| Digest-prefix mismatch could cause spurious notifies | code-quality (gate) | **Verified non-bug** (both sides bare hex); added a real-hex integration test that had been masked (`d492e8e`) |
| Anonymous GETs to upstream | research (MUST) | Descriptive `User-Agent` on both fetchers (`0179f0f`) |
| Egress-config parity gap (fail-closed) | security (SHOULD) | Threaded `config.egress` into catalog SSRF validation (`d492e8e`) |

## Deferred follow-ups (non-blocking, tracked)
- **DNS-rebind TOCTOU:** wire `DnsPinningTransport` into catalog + ingest fetchers
  (pre-existing across the fetch layer).
- **Backoff / `Retry-After`** on `429/503` (needs a `next_retry_at` column →
  Phase-4 scanner), honor `X-Poll-Interval`/`Cache-Control`.
- **Fairness:** order candidates by oldest `last_checked_at` + `max_probes_per_sweep`.
- **Extract** the shared SSRF/redirect/size guard out of the two fetchers.
- **Semantic diff** (oasdiff) after the hash gate to kill whitespace false-positives;
  **distinct "spec removed"** signal on 404; **coalescing** by vendor.
- **Dedicated scheduler** over lazy-read piggyback (this is the Phase-4
  `catalog_update_scanner`).

## Test plan
- [x] `make lint` (ruff + mypy) clean
- [x] Unit — fetch: 200/304/stale-etag/no-etag/oversize(body+content-length)/
      unsafe-url + SSRF-via-redirect + per-hop `If-None-Match` + too-many-redirects
      + `User-Agent` + 304-refreshed-ETag
- [x] Unit — sweep: interval/boundary/change-once/dedupe/no-change/empty/per-API
      isolation + offload spawn + deadline-swallow + concurrency-bound +
      `_safe_refresh`→spawn wiring + `ALL` membership
- [x] Integration (real Postgres, HTTP boundary stubbed only): sweep emits+records,
      dedupes across runs, no-event-when-matching (opaque + **real bare-hex**),
      repo round-trip + notify join
- [x] ORM arch conventions + registry unit suite

### DIFF
--- a/src/jentic_one/migrations/registry/versions/d3e4f5a6b7c8_add_catalog_update_checks_table.py
+++ b/src/jentic_one/migrations/registry/versions/d3e4f5a6b7c8_add_catalog_update_checks_table.py
@@ -1,0 +1,56 @@
+"""add catalog_update_checks table
+
+Revision ID: d3e4f5a6b7c8
+Revises: c2d3e4f5a6b7
+Create Date: 2026-07-31
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+from jentic_one.shared.db.types import GUID
+
+revision: str = "d3e4f5a6b7c8"
+down_revision: str | None = "c2d3e4f5a6b7"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    pg = op.get_bind().dialect.name == "postgresql"
+    op.create_table(
+        "catalog_update_checks",
+        sa.Column(
+            "id",
+            sa.String(length=30),
+            server_default=sa.text("generate_ksuid('cuc')") if pg else None,
+            nullable=False,
+        ),
+        sa.Column("local_api_id", GUID(), nullable=False),
+        sa.Column("spec_url", sa.String(length=2048), nullable=False),
+        sa.Column("last_seen_etag", sa.Text(), nullable=True),
+        sa.Column("last_seen_digest", sa.String(length=64), nullable=True),
+        sa.Column("last_notified_digest", sa.String(length=64), nullable=True),
+        sa.Column("last_checked_at", sa.DateTime(timezone=True), nullable=True),
+        sa.Column(
+            "created_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.func.now(),
+            nullable=False,
+        ),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_index(
+        "ix_catalog_update_checks_local_api_id",
+        "catalog_update_checks",
+        ["local_api_id"],
+        unique=True,
+    )
+
+
+def downgrade() -> None:
+    op.drop_index("ix_catalog_update_checks_local_api_id", table_name="catalog_update_checks")
+    op.drop_table("catalog_update_checks")
--- a/src/jentic_one/registry/core/schema/__init__.py
+++ b/src/jentic_one/registry/core/schema/__init__.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from jentic_one.registry.core.schema.api_revisions import ApiRevision
 from jentic_one.registry.core.schema.apis import Api
 from jentic_one.registry.core.schema.catalog_snapshots import CatalogSnapshot
+from jentic_one.registry.core.schema.catalog_update_checks import CatalogUpdateCheck
 from jentic_one.registry.core.schema.notes import Note
 from jentic_one.registry.core.schema.operation_url_index import OperationURLIndex
 from jentic_one.registry.core.schema.operations import Operation
@@ -18,6 +19,7 @@ __all__ = [
     "Api",
     "ApiRevision",
     "CatalogSnapshot",
+    "CatalogUpdateCheck",
     "Note",
     "Operation",
     "OperationURLIndex",
--- a/src/jentic_one/registry/core/schema/catalog_update_checks.py
+++ b/src/jentic_one/registry/core/schema/catalog_update_checks.py
@@ -1,0 +1,61 @@
+"""CatalogUpdateCheck ORM model — per-API upstream-change bookkeeping (Flow 3).
+
+One row per **local** registered API (``local_api_id``, unique — never the
+upstream manifest id, per D-005a) recording what the update-notify sweep last
+observed for that API's upstream spec URL:
+
+- ``last_seen_etag`` — the validator sent back as ``If-None-Match`` on the next
+  sweep so an unchanged spec answers ``304`` and transfers no body.
+- ``last_seen_digest`` — ``sha256`` of the last fetched spec bytes. Recorded for
+  observability / future weak-ETag handling; the current MVP dedupes on
+  ``last_notified_digest`` and compares against the registered revision's
+  ``spec_digest``, so this column is written but not yet read.
+- ``last_notified_digest`` — the digest that last produced a
+  ``catalog.update_available`` event, so a persistently-changed spec notifies
+  **once** rather than every sweep.
+- ``last_checked_at`` — when the sweep last probed this API.
+
+The table is deliberately narrow and relationship-free: it is sweep scratch
+state, not part of the API aggregate, so it does not widen ``apis``.
+"""
+
+from __future__ import annotations
+
+import uuid
+from datetime import datetime
+
+from sqlalchemy import Index, String, Text
+from sqlalchemy.orm import Mapped, mapped_column
+from sqlalchemy.sql import func
+
+from jentic_one.shared.db.base import RegistryBase
+from jentic_one.shared.db.ids import generate_ksuid
+from jentic_one.shared.db.types import GUID, UTCDateTime
+from jentic_one.shared.db.utils import utcnow
+
+
+class CatalogUpdateCheck(RegistryBase):
+    """Per-API record of the last upstream-spec observation for update notify."""
+
+    __tablename__ = "catalog_update_checks"
+    __table_args__ = (Index("ix_catalog_update_checks_local_api_id", "local_api_id", unique=True),)
+
+    id: Mapped[str] = mapped_column(
+        String(30),
+        primary_key=True,
+        default=lambda: generate_ksuid("cuc"),
+        server_default=func.generate_ksuid("cuc"),
+    )
+    #: The local registry API this row tracks (``apis.id``). No FK constraint:
+    #: the row is cheap scratch state and a dropped API is reconciled by the
+    #: sweep (a check with no matching registered spec url is simply ignored).
+    local_api_id: Mapped[uuid.UUID] = mapped_column(GUID(), nullable=False)
+    #: The upstream spec URL probed (the API's revision ``source_url``).
+    spec_url: Mapped[str] = mapped_column(String(2048), nullable=False)
+    last_seen_etag: Mapped[str | None] = mapped_column(Text, nullable=True)
+    last_seen_digest: Mapped[str | None] = mapped_column(String(64), nullable=True)
+    last_notified_digest: Mapped[str | None] = mapped_column(String(64), nullable=True)
+    last_checked_at: Mapped[datetime | None] = mapped_column(UTCDateTime(), nullable=True)
+    created_at: Mapped[datetime] = mapped_column(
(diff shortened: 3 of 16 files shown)


ren-jentic:
Finding (medium): candidate selection is nondeterministic when an API has >1 non-archived revision
registered_specs_for_notify has no ORDER BY and dedupes first-row-wins per api_id. Multiple non-archived revisions per API are the documented norm here — registered_spec_urls's own docstring explains that imports create drafts that are never auto-promoted. So which revision's spec_digest (and source_url!) the sweep compares against is DB-row-order dependent.

Concrete false positive: user proactively re-imports an updated spec (no notify ever fired, so last_notified_digest is empty), leaving an older revision alongside the fresh one. If the query happens to return the old revision, the next sweep sees upstream ≠ old digest ≠ last-notified → emits "update available" for bytes the user already has. The dedupe only saves the case where a notify already fired. And if two revisions carry different source_urls, which URL gets probed at all is a coin flip.

Fix is small: prefer the API's current_revision_id when set, else newest created_at, and add a deterministic ORDER BY regardless. The test suite is thorough everywhere else but has no case for an API with two live revisions — add one; it would have caught this.

Nits
The inline-fallback that isn't: _spawn_update_notify_sweep's except RuntimeError branch logs catalog_update_sweep_no_loop_running_inline and the comment says "run inline as a fallback" — then it just returns. Nothing runs inline. It's dead code in practice (callers are async), but the log line is actively misleading; either implement it or rename to _skipped.
Manual POST /catalog:refresh never sweeps — only the lazy _safe_refresh path spawns it. The operator who explicitly hits refresh is the most intentional "check upstream now" signal in the system and gets no update check. Wire it or document the asymmetry.
Kill-switch is checked inside the task, so a disabled install still spawns a no-op task per refresh. Harmless; noted only for completeness.

PR #904, opened 2026-07-31
feat(overlay): materialize confirmed overlays onto the served spec

Manuel-Jentic:
## TL;DR

Confirming an overlay used to be a **no-op** — the overlay was stored but the served spec never changed, and `?overlays=` at read time did nothing. This PR makes `:confirm` **real**: it applies the overlay's JSONPath actions to the API's current spec, validates the result, and re-ingests it as a new revision that becomes the one the platform serves and operates on.

This is **Phase 3** of the spec-flywheel work (contribute → confirm → materialize).

---

### What changes for a user
- An operator (`org:admin`) confirms a pending overlay. The platform re-ingests the *overlaid* spec and promotes it to the current revision.
- Because the full ingest pipeline re-runs, everything downstream of the spec is rebuilt from the fix: `operations`, `operation_url_index`, `servers`, `security_schemes`, and `spec_files`. So broker behavior that keys off those tables (operation identity, PBAC, credential injection, region hints) now reflects the corrected spec.
- The overlay records which revision it produced (`confirmed_revision_id`), now visible on the overlay API responses.
- Contributors still `submit`/`update` overlays with `apis:write`; **confirm** is an operator action (`org:admin`) because it rewrites served state.

> Note (accurate framing): the broker is a transparent proxy that forms the upstream request from the **client-supplied URL**, so a spec fix changes broker behavior *through the extracted tables* (auth scheme, server-variable detection, operation identity) — it does not by itself retarget traffic to a new `servers[].url`.

### Permission change: why `:confirm` now needs `org:admin`

`POST …/overlays/{id}:confirm`: `apis:write` → **`org:admin`**. OpenAPI + endpoint reference regenerated.

**Why we tightened it.** Before this PR, confirming an overlay was a no-op — it just flipped a status field and changed nothing anyone served, so `apis:write` was fine. This PR makes confirm **rewrite the API's served spec for everyone**: it applies the overlay, re-ingests the result, and promotes it to the current revision. That's a much bigger blast radius, so the permission needs to match it.

Think of it like a pull request:
- **Contributors (`apis:write`) propose a fix** — they `submit`/`update` an overlay. That's cheap and reversible and affects nothing that's being served.
- **Operators (`org:admin`) review and apply it** — `:confirm` is the "merge" step that changes what the whole org serves and operates on.

Letting `apis:write` confirm would mean any contributor could unilaterally rewrite the live spec for every consumer of that API — the exact privilege escalation this gate prevents.

**This is deliberately coarse for the MVP and should not stay `org:admin`.** `org:admin` conflates "operator who applies spec fixes" with "org administrator". A named next-steps follow-up (see below) will move `:confirm` to a purpose-scoped permission — a dedicated `overlays:confirm` / operator role or an `apis:promote`-style scope — so applying an overlay doesn't require full org-admin.

### Deferred (named follow-ups, not in this PR)

These are **intentionally out of scope** for the Phase-3 MVP and tracked for later phases. Reviewers: please review this PR against the "materialize a confirmed overlay onto the served spec" goal — the lifecycle gaps below are known and owned, not oversights.

**Phase 3.x — overlay lifecycle (fast-follow):**
- **Confirm authorization scope (`org:admin` is temporary).** `:confirm` is gated on `org:admin` for now as a conservative default (see the permission section above). This should move to a purpose-scoped permission — an `overlays:confirm` / operator role or an `apis:promote`-style scope — so applying an overlay doesn't require full org-admin. Owned as an explicit next step.
- **Un-confirm / rollback.** `deprecate` only flips overlay status; it does **not** roll the materialized revision back, so a deprecated-after-confirm overlay keeps serving with nothing backing it. The prior revision is recoverable (archived, not deleted) — the intended model is "deprecate promotes the prior archived revision back to current", not yet implemented.
- **Update-after-confirm.** *Partly addressed in this PR:* a CONFIRMED-but-unmaterialized overlay (its materialize job failed deterministically) is now editable and resets to PENDING, giving the operator an escape hatch instead of forever re-enqueueing the same failing job; and the confirm/update race (an `update` landing during apply+validate) is closed by re-reading the document inside the claim transaction and bailing if it changed. **Still deferred:** `update` of an already-*materialized* CONFIRMED overlay is rejected (no in-place re-materialize) — a proper "re-materialize on edit" path is future work.
- **`target_revision_id` is advisory.** Materialization always applies to the **current** revision, not the revision the overlay was authored against. Structural drift fails closed (409); semantic base-drift now emits an `overlay_confirm_base_drift` warning. Per-target-revision materialization is deferred.
- **Overlay stacking.** Multiple overlays compose by confirm order (last-confirmed-wins, cumulative, since each applies to current). No per-overlay unwind and no explicit stacking contract yet.

**Phase 4 — Flow-3 (catalog update) interplay:**
- **Overlay ↔ catalog-update collision.** A materialized overlay revision keeps the base catalog `source_url`, so the Flow-3 sweep compares upstream against the *overlaid* digest and the API can perpetually look "update available"; accepting a catalog re-import silently supersedes the overlay. This confirm↔re-import interaction is currently un-owned and will be resolved in Phase 4 (no-op → auto-deprecate; conflict → warn event), tracked with internal#779.

None of the above block this PR; they are recorded in `spec-flywheel-tracker.md` under Phase 3 / Phase 4.

---

### Flow

```mermaid
sequenceDiagram
    actor Op as Operator
    participant API as confirm endpoint
    participant Svc as OverlayService.confirm
    participant Reg as registry_db
    participant Job as admin_db jobs
    participant W as ImportHandler worker

    Op->>API: POST overlays/{id}:confirm
    API->>Svc: confirm(...)
    Svc->>Reg: read overlay (tx1)
    Note over Svc: apply overlay to CURRENT spec and validate<br/>(size caps, SSRF, openapi key) BEFORE any state change
    Svc->>Reg: CAS flip pending to confirmed (WHERE status=pending)
    alt lost race (rowcount = 0)
        Svc-->>API: idempotent return, no job
    else claimed (rowcount = 1)
        Svc->>Job: enqueue inline import (origin=overlay, overlay_id)
        Svc-->>API: confirmed
        Note over Job,W: later, asynchronously
        W->>Job: dequeue materialize job
        W->>W: re-run full ingest pipeline on overlaid bytes
        W->>Reg: archive all active revisions, then create and promote overlay revision
        W->>Reg: set overlay.confirmed_revision_id = new revision
    end
```

### State model

```mermaid
stateDiagram-v2
    [*] --> pending: submit
    pending --> confirmed: confirm (CAS claim), enqueue materialize
    confirmed --> confirmed: re-confirm w/ null confirmed_revision_id -> re-enqueue (recovery)
    confirmed --> confirmed: re-confirm w/ revision set -> idempotent no-op
    pending --> deprecated: deprecate
```

### Correctness / safety properties
- **Apply-before-flip:** a drifted or unsafe overlay is rejected (409) while still `pending` — no torn confirm, no orphaned job.
- **Idempotent confirm:** compare-and-swap `set_status ... WHERE status='pending'`; a lost race returns the current view without enqueueing a duplicate job.
- **Recovery:** a `CONFIRMED` overlay with a null `confirmed_revision_id` (enqueue failed / crash) re-enqueues on the next confirm; if the revision already exists, the worker re-links it on the duplicate re-ingest instead of failing the job.
- **One-active-revision index:** overlay origin archives *all* active revisions (PUBLISHED **and** IMPORTED) so the overlaid revision supersedes a live catalog- or manually-promoted base.
- **DoS bounds:** `max_spec_bytes` enforced on the inline ingest branch, overlaid-spec size cap at confirm, and submit/update validate overlay document shape + size + action count.
- **SSRF:** `servers[].url` validated at document-, path-, and operation-level, including templated-host skeletons; whole-host variables deferred to exec-time validation.

### Key files
- `services/overlay_apply.py` — fail-closed JSONPath applier (`$`/`.key`/`['key']` + remove/update), ported from the contribute-spec-fix skill.
- `services/overlay_service.py` — `confirm` rewrite, validation, enqueue, recovery.
- `services/import_service.py` — back-link + duplicate-reingest recovery.
- `ingest/stages/extract_api.py`, `repos/revision_repo.py` — `archive_all_active` for overlay origin.
- `ingest/fetch.py` — inline size cap + `InlineSource.source_url` provenance threading.
- migration `e4f5a6b7c8d9` — `Overlay.confirmed_revision_id`.

## Review process
Reviewed by a four-lens board — [Security](4f7f4ee7-963f-4820-a4da-77f641c2e329), [Code quality](6e56b52c-266e-4099-996d-2eb1e742e6bf), [Feasibility & fit](d21229d3-8dc0-4033-ad42-17c04e042805), and [Best-practices research](fc944314-55ec-4008-b893-866e8e1831ea) — all High/Medium findings addressed in this PR.

## Test plan
- [x] Unit + arch: `pytest tests/unit tests/arch` (2457 passed)
- [x] Integration (Postgres): overlay repo + import handler incl. published-base archive, cross-origin archive, duplicate-reingest recovery
- [x] `ruff` + `mypy` clean; `openapi_export` + `endpoint_tree` regenerated
- [x] Scripted end-to-end vs local Postgres: submit→confirm→worker→served-spec rewrite + link, re-confirm no-op, recovery re-link, operation-level SSRF reject, document-validation rejects (16/16)


ren-jentic (commented):
Finding 1 (medium): confirm/update race produces the exact divergence this PR defers
confirm reads overlay.document in tx1, then does apply+validate (deepcopy + json.dumps of a potentially multi-MB spec — a real window), then the CAS in tx2 guards status only. An update landing in that window is legal (the overlay is still PENDING), so the CAS succeeds and the materialize job carries the old document while the row stores the new one: confirmed_revision_id ends up pointing at a revision that does not embody the stored document. That's precisely the "stored document and served revision diverge" hazard the PR explicitly defers for update-after-confirm — but this variant happens inside a single confirm and isn't documented or tested. Cheap fix: include a version guard in the CAS (WHERE status='pending' AND updated_at = <read value>), or re-read the document inside the claim transaction and bail to 409 if it changed.

Finding 2 (medium): deterministic ingest failure after confirm is a lifecycle dead-end
Confirm's pre-flip checks (size, SSRF, openapi-is-a-string) are a subset of what the async pipeline enforces (ValidateOpenAPISpec alone also rejects e.g. any spec containing an arazzo key, and downstream extract stages have their own failure modes). If the worker fails deterministically, the overlay is CONFIRMED with a null confirmed_revision_id — and there is no way out: re-confirm takes the recovery path and re-enqueues the same failing job; update requires PENDING; deprecate abandons it. Served state stays safe (unchanged) and the job failure is observable, so this isn't silent — but the state machine has a terminal loop with no operator lever to fix the document. Smallest fix: permit update (or an explicit reset-to-PENDING) when status=CONFIRMED AND confirmed_revision_id IS NULL. This also slots naturally into your Phase-3.x lifecycle follow-ups.

Manuel-Jentic:
Thanks @ren-jentic — both addressed in `07980b8`.

**Finding 1 (confirm/update race).** Confirmed the window is real: apply+validate runs outside any transaction, and an `update` landing while it runs is legal (still PENDING), so the status-only CAS would have enqueued a job carrying the old document against a row storing the new one. Fix: re-read the document inside the claim transaction and bail if it changed (status ≠ PENDING or `document` differs) → idempotent return of the current view, no claim, no enqueue. So the caller re-confirms against the new document. Test: `test_confirm_document_changed_during_apply_returns_without_enqueue`.

**Finding 2 (deterministic-failure dead-end).** Agreed it was a terminal loop with no operator lever. Fix: `update` is now permitted when `status=CONFIRMED AND confirmed_revision_id IS NULL`, and it resets the overlay to PENDING (clearing stale confirm metadata) so the operator can fix the document and re-confirm instead of re-enqueueing the same failing job. A *materialized* CONFIRMED overlay still 409s. Tests: `test_update_stuck_confirmed_resets_to_pending` + `test_update_confirmed_raises_conflict` (now uses a materialized overlay).

Also documented in the PR body / tracker: the `org:admin` gate on `:confirm` is explicitly temporary — a named next step moves it to a purpose-scoped `overlays:confirm` / operator role rather than full org-admin.

PR #911, opened 2026-07-31
feat(access-requests): surface existing-toolkit reuse on the provision path

### LINKED ISSUE #897: Broker directive should suggest binding to existing toolkit, not only --provision
## Summary

When an agent without a toolkit binding tries to execute an operation, the broker returns a 403 with an `agent_directive` that only suggests `jentic access request --provision <vendor/name>`. This creates a **new** toolkit from scratch, even when the instance already has a toolkit (e.g. "My Travel Agent") that the user wants to add the API to.

## Context / How I got here

1. Had a working Jentic One instance with a toolkit ("My Travel Agent") containing 3 credentials
2. The original agent became invalid (token assertion failure after a database mixup caused by `make dev` spinning up a second Postgres container)
3. Registered a new agent to recover
4. Tried to execute an operation through the broker — got 403 `no_toolkit_binding`
5. The broker's `agent_directive` only suggested `--provision` (create new toolkit)
6. Filed the provisioning request, which tried to create a brand new toolkit
7. Had to deny it because the existing toolkit should be reused

## Expected behavior

The broker or CLI should:
1. Check if a toolkit already exists in the instance that could serve the API
2. If yes: suggest `jentic access request --toolkit <vendor/name>` to bind the agent to it, or offer to add the new API to the existing toolkit
3. If no: then suggest `--provision`

Alternatively, the `--provision` flow should detect existing toolkits and ask "Do you want to add this to an existing toolkit or create a new one?"

## Additional context

- The `--toolkit` flag exists on `jentic access request` but the broker's error message never mentions it
- The user has no way to know from the error message that their existing toolkits are still there and reusable
- This is especially painful when recovering from an invalidated agent — you're forced through the full provisioning flow as if nothing existed before

## Related

- Same instance has "My Travel Agent" toolkit (1 key, 3 credentials, active 4 days ago)
- The new agent needs to be bound to it, not create a parallel toolkit

ren-jentic:
Closes #897

## Summary

The wizard's adopt-an-existing-toolkit mechanics already shipped (#885/#902), but the `serves=False` provision flow still *read* as "recreate everything from scratch", and a reuse approval left no record of which toolkit fulfilled the plan. This PR closes that residual gap:

- **Broker directive** — the `no_toolkit_binding` / `toolkit_serves_api: false` instruction now tells the agent the operator can add the credential to an **existing** toolkit during fulfilment. `parameters` (`toolkit_serves_api`, `suggested_command`) are deliberately untouched so CLI/skill parsing stays stable; the docstring records why the broker never enumerates toolkits here (inventory leak to an unbound agent, no scoped view of a future approver).
- **Skill** (CLI embed + served copy, kept byte-identical per the drift arch test) — the `toolkit_serves_api: false` recovery bullet and the execute-denial pitfalls entry now mention the reuse choice so agents relay it to their operator.
- **Wizard audit honesty** — submit now also stamps the fulfilled toolkit/credential ids onto the inert `toolkit:create` / `credential:provision` placeholders, so a reuse approval records "fulfilled by tk_…" instead of implying a create that never happened. No backend change needed: `:amend` already accepts `resource_id` on any pending item, and approving a fulfilment-only item stays a recorded no-op.
- **Tests** — broker unit test guards the reuse hint *and* the exact parameter shape; a control e2e drives the full reuse amend+decide shape (full approval — the agent's `--wait` exits 0 — binds land on the pre-existing toolkit, no parallel toolkit row); wizard tests assert the placeholder amendments on both create and adopt paths.

CLI `--provision --toolkit-id` (filing a 3-item plan targeting a known toolkit) is deliberately deferred: the reuse choice belongs to the operator in the wizard, and agents don't know ids of toolkits they aren't bound to.

## Test plan

- [x] `make lint` (ruff + format + mypy) clean
- [x] `pytest tests/unit tests/arch` — 2459 passed
- [x] `JENTIC_TEST_BACKEND=sqlite pytest tests/integration/control/test_provisioning_plan_e2e.py` — 6 passed (Postgres backend runs in CI)
- [x] `ui`: `npm run lint:fix`, `npm run build`, `npm run test:run` — 967 passed
- [x] `cli`: `go test ./internal/skillgen/...` — ok
- [ ] Manual: trigger a `serves=False` denial and confirm the directive/skill text; fulfil a plan via "use an existing toolkit" and check the decided request names the reused toolkit on the `toolkit:create` item

Please squash-merge.

### DIFF
--- a/cli/internal/skillgen/content/jentic.md
+++ b/cli/internal/skillgen/content/jentic.md
@@ -147,7 +147,10 @@ jentic access request --provision stripe.com/api \
     describes the whole path (create toolkit, provision + bind a credential with
     your proposed rules, bind you), which a human fulfils and approves in the
     dashboard. The directive's `suggested_command` already points at `--provision`
-    in this case.
+    in this case. The plan does **not** force a new toolkit: during fulfilment
+    the operator can add the credential to a toolkit they already have (the
+    wizard offers both) — worth relaying when your operator mentions an
+    existing toolkit they want to extend.
   - `toolkit_serves_api: true` — a toolkit already serves this API and you just
     aren't bound to it; the directive suggests `jentic access request --toolkit
     <vendor/name> --wait`. File that and wait for approval.
@@ -451,7 +454,8 @@ jentic execute <operation_id> --broker-scheme http --broker-host 127.0.0.1:8100
     bare `--toolkit` bind would be **denied** ("No toolkit serves API …"); the
     directive suggests `--provision` instead — file that plan (propose `--auth`
     and `--rules-json` from the spec, pass `--reason`) and your operator fulfils
-    it in the dashboard.
+    it in the dashboard, into a new toolkit **or one they already have** — an
+    existing toolkit never has to be recreated.
   - **424 `credential_not_provisioned`** → the directive gives a
     `provisioning_url` for your operator to connect an account (an access
     request won't help).
--- a/src/jentic_one/broker/core/exceptions.py
+++ b/src/jentic_one/broker/core/exceptions.py
@@ -281,6 +281,12 @@ def no_toolkit_binding_directive(
       because a toolkit only serves an API once a credential for it is bound
       (issue #683). So point the caller at credential provisioning first — the
       step that actually creates a serving toolkit — before the binding request.
+      The instruction also says the operator may fulfil the plan **into an
+      existing toolkit** (#897): toolkits serving *other* APIs may well exist,
+      and without the hint the flow reads as "recreate everything from scratch".
+      Deliberately text-only — the broker never enumerates other toolkits here
+      (that would leak instance inventory to an unbound agent, and this
+      stateless edge has no scoped view of what a future approver could reuse).
 
     Both variants name the exact next step so an autonomous agent can hand it to
     its operator verbatim instead of reading docs.
@@ -308,7 +314,8 @@ def no_toolkit_binding_directive(
     # first (which creates the serving toolkit), then the agent is bound. A bare
     # `toolkit:bind` request now can never be approved — so steer the agent to
     # file the whole path as one provisioning plan (`--provision`), which a human
-    # fulfils and approves in the dashboard (they enter the secret + grant).
+    # fulfils and approves in the dashboard (they enter the secret + grant, into
+    # a new toolkit or one they already have — the wizard offers both, #897).
     parameters["suggested_command"] = (
         f'jentic access request --provision {api} --reason "<why you need this>" --wait'
     )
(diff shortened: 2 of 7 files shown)


Manuel-Jentic (approved):
## Approve

Reviewed for need, feasibility, and product fit — including a supersession check against `main` and all open PRs, source-level verification of the "no backend change needed" claim, and a local re-run of the test suites on this branch.

### Needed, not superseded

- `main` (a11025a5) has none of this: `no_toolkit_binding_directive` carries no existing-toolkit hint, both `jentic.md` copies lack the reuse bullets, and the wizard sends no placeholder amendments. #885/#902 shipped the adopt mechanics but not this residual — the PR is exactly the gap it claims to be.
- The intake bot's `duplicate-candidate` flags on #897 (#662, #683) are both closed and fixed a different failure mode (bind-first auto-denial → credential-first directive). The reuse angle was untouched. Not a duplicate.
- No open PR overlaps (checked all 18).

### Feasibility — verified in source and by execution

- `AccessRequestService.amend()` accepts `resource_id` on any pending item; `_validate_grantable_scope` early-returns for anything that isn't `scope:grant`, so stamping the `toolkit:create`/`credential:provision` placeholders passes amend. At decide time, fulfilment-only items stay inert (`EffectApplicator.validate` only rejects them if they carry `rules`). "No backend change needed" is true.
- Governance safety: `plan_governance_for_items` keys off intent `(type, action, status, resource_reference)` — never `resource_id` — so the stamping cannot flip a bind's governed/plain contract. Good.
- Locally re-ran on this branch: broker unit + arch (287 passed), `JENTIC_TEST_BACKEND=sqlite` provisioning-plan e2e (6 passed, incl. the new reuse test), wizard dialog UI tests (21 passed), `go test ./internal/skillgen/...` ok, CLI/served skill copies byte-identical.
- #912 merged to `main` after this branch was cut and also touches both `jentic.md` files: simulated the merge (`git merge-tree`) — clean, no conflicts, and the two skill copies remain **byte-identical post-merge**, so the drift arch test holds after squash-merge.

### Product fit

- Fixes the reporter's literal dead-end (plan reads as "recreate everything" → operator denies). Intake scored it fit:high/severity:major.
- Text-only directive is the right boundary call: the broker stays stateless and never enumerates toolkits to an unbound agent — the docstring records the security rationale, and the unit test pins the machine-stable `parameters` shape so CLI/skill parsing can't regress.
- Deferring `--provision --toolkit-id` is consistent with the same posture (agents don't know ids of toolkits they aren't bound to).
- The placeholder stamping is a genuine audit improvement: reuse approvals now name the fulfilling toolkit instead of implying a phantom create.

### Non-blocking notes

- The manual-test checkbox is unticked — the new e2e covers the amend+decide shape through the real service layer, so I'm comfortable, but ticking it (or noting it ran) would be good hygiene before merge.
- #897 also wishes for agent re-validation / binding inheritance on re-register. Those die silently when this closes the issue — suggest spinning them into a follow-up issue so they aren't lost.

Squash-merge away.

PR #912, opened 2026-07-31
feat(flow3): standalone catalog-update scanner + update-available surfaces

Manuel-Jentic:
## TL;DR

Phase 2 shipped the Flow-3 **signal** — a lazy, read-triggered ETag check that could notice an upstream spec had changed. This PR makes Flow-3 **whole**: a standalone background **scanner** proactively sweeps upstream-tracked APIs on an interval, and the "update available" state is now **surfaced and actionable** on every operator surface — the API list/detail, the Discover browse page, and the CLI (`catalog outdated`, a marker in `catalog list`, and a line in `jenticctl status`). Re-importing an API **clears** the prompt automatically.

This is **Phase 4** of the spec-flywheel work (contribute → confirm → materialize → **stay current**).

> **Rebased onto [#852](https://github.com/jentic/jentic-one/pull/852)** (merged). #852 persists the catalog identity as an `apis.catalog_api_id` column; this branch was reconciled to *consume* that stored column and drop its own earlier read-time derivation of the same field — one source of truth. Everything else here (scanner, `update_available`, `outdated_*`, Discover/CLI surfaces, the SQLite settle fix) is independent of #852.

---

### The loop this closes

An API's OpenAPI spec is imported from an upstream catalog manifest. Upstream, that spec drifts — a new operation, a changed server, a security-scheme tweak. Before this PR, nothing told the operator; the platform kept serving the stale revision until someone manually re-imported on a hunch.

Now the platform **notices and tells you**, then lets you fix it in one click:

1. A background **scanner** conditionally re-fetches the spec URLs of upstream-tracked APIs (`If-None-Match`) and, when the bytes changed, emits a `catalog.update_available` **actionable** event.
2. That state shows up wherever you'd look — the Workspace API detail ribbon, the Discover page, and the CLI.
3. You **re-import** (one click in the UI, or `jentic catalog import`), which adopts the upstream spec as a new served revision **and clears the prompt**.

### What changes for a user

- **Workspace** — an API whose upstream spec drifted shows an "update available" state; the API detail ribbon's **Re-import** button now targets the API's real catalog id (see below), so umbrella specs re-import correctly instead of 404ing.
- **Discover** — cards render an **"Update available"** badge, a new **"Updates"** filter narrows to just the outdated entries, and the status row shows an outstanding-update count.
- **CLI** — `jentic catalog outdated` lists registered entries with an upstream update; `catalog list --registered` gains an `UPDATE AVAILABLE` marker; `jenticctl status` prints an "N update(s) available" line (best-effort, never fails `status`).
- **Operators of air-gapped installs** — the whole sweep + scanner is a **kill switch**: set the interval to `0` and there's no egress and no event spam. Standalone-registry deployments (no admin DB) get no scanner at all.

### Two gaps this PR also closes (found in manual E2E)

- **Re-import used the wrong id.** The one-click Re-import posted to `/catalog/{vendor}:import`, which **404s for umbrella specs** whose catalog id is `domain/sub` (e.g. `nytimes.com/article_search`). The fix now rides on [#852](https://github.com/jentic/jentic-one/pull/852) (merged), which persists the catalog identity as `apis.catalog_api_id`: the UI consumes that column (surfaced on `ApiResponse.catalog_api_id`) and prefers it over `vendor` for the re-import call, so umbrella specs resolve correctly. This branch was reconciled onto #852 — our earlier read-time derivation of `catalog_api_id` was dropped in favour of #852's single stored source of truth.
- **The prompt didn't clear on SQLite.** Re-importing an API is supposed to acknowledge its `catalog.update_available` event. The settle opened a *second* admin transaction while the job worker already held one — a self-deadlock on SQLite's single writer (`database is locked`). The settle now **reuses the handler's own session** inside a `SAVEPOINT`, so the ack is atomic with the import and can't deadlock.

### What's NOT in this PR (deferred / still missing)

Reviewers: please review against the "proactively notice upstream drift and make it actionable" goal. The items below are deliberately out of scope — known, owned, and tracked in `spec-flywheel-tracker.md` (Phase 3.x / Phase 4 / Phase 5). None of them block this PR.

**Flow-3 scope items not built here (Phase 4):**

- **Sweep jitter.** The sweep runs on a fixed interval with bounded concurrency (`update_sweep_max_concurrency`) under a wall-clock budget (`update_sweep_deadline_seconds`), but there is **no inter-replica jitter** — multiple replicas can align their sweeps. Tolerated for now (the emit is idempotent and dedupes on the persisted digest); jitter is a follow-up.
- **Notification hygiene / snooze.** No per-`(api_id, digest)` snooze (OQ-8) — once an update is notified it stays surfaced until the API is re-imported; an operator can't temporarily dismiss it.
- **Re-import × overlays interplay** — `blocked(internal#779)`. See the collision item below; the actual resolution (no-op → auto-deprecate; conflict → warn event) is gated on fixing internal#779 first.
- **internal#780** (hygiene) — noted, not a gate for this PR.

**Overlay ↔ catalog-update collision** *(carried from #904, Phase 3.x)*. A materialized overlay revision keeps the base catalog `source_url`, so the scanner compares upstream against the *overlaid* digest → the API can perpetually look "update available", and accepting a catalog re-import **silently supersedes the overlay**. This confirm↔re-import interaction stays un-owned here and is resolved next, **tracked with internal#779**.

**`:confirm` authorization gate is still `org:admin`** *(carried from #904, Phase 3.x)*. Overlay `:confirm` remains gated on `org:admin` as a deliberately conservative default (confirm rewrites the served spec). That's coarser than intended — it conflates "operator who applies spec fixes" with "org administrator". Moving it to a **purpose-scoped permission** (an `overlays:confirm` / operator role, or a dedicated `apis:promote`-style scope) is a separate, owned follow-up and intentionally **not** bundled into this Phase-4 PR (it touches the auth model, not the update loop). *(This is the "new scope instead of `org:admin`" item — confirmed not done here.)*

**Other overlay-lifecycle carry-overs (Phase 3.x, not touched here):** update of an already-*materialized* CONFIRMED overlay still 409s (no in-place re-materialize); `target_revision_id` is advisory (materialization always applies to the *current* revision); overlay **stacking** is last-confirmed-wins/cumulative with no per-overlay unwind and no documented composition contract.

**Open questions still unanswered** (defaults chosen, not ratified): OQ-2 (re-import trigger = operator-only), OQ-3 (scanner scope), OQ-5 (upstream version bumps as a separate manifest-diff signal), OQ-7 (telemetry — none by default), OQ-8 (snooze).

**Phase 5 — skill distribution** (entirely future): relocating canonical skill copies + `test_skill_drift`, multi-skill skillgen, `/skills/<name>.md` + `llms.txt` + manifest serving, and `jentic skill update` (`SourceHosted`, `blocked(internal#407)`).

---

### Flow

```mermaid
sequenceDiagram
    participant Sch as CatalogUpdateScanner
    participant Svc as CatalogService
    participant Up as Upstream
    participant Chk as catalog_update_checks
    participant Ev as admin_db events
    actor Op as Operator
    participant Imp as ImportHandler

    loop every interval, 0 disables
        Sch->>Svc: run_update_sweep, serialized by asyncio.Lock
        Svc->>Chk: pick upstream-tracked APIs due for re-probe
        Svc->>Up: conditional GET with If-None-Match
        alt 200 and digest changed
            Svc->>Ev: emit catalog.update_available, requires_action
        else 304 or unchanged
            Svc->>Chk: bump last-checked, no event
        end
    end

    Op->>Imp: re-import API, adopts upstream spec
    Note over Imp: runs inside the worker admin BEGIN IMMEDIATE
    Imp->>Ev: settle_actionable_events in a SAVEPOINT on the handler session
    Note over Imp,Ev: ack is atomic with the import, no nested-txn self-deadlock on SQLite
```

### Backend

- **Scanner** — `shared/jobs/catalog_update_scanner.py`, wired in `app_factory` only when **both** the registry and admin DBs are present; one sweep per `catalog.update_check_interval_seconds`. Each API is re-probed at most once per interval via a **persistent per-API gate** (`catalog_update_checks`), so the scanner and the manual `POST /catalog:refresh` never double-probe. In-process sweeps are serialized by an `asyncio.Lock`.
- **Sweep scope** — the classifier (`_is_upstream_tracked`) keys on the **catalog manifest** `spec_url`s (via `CatalogRepository.manifest_spec_urls`), not local registered URLs — so "catalog-tracked" is derived from what the manifest actually serves. Outdated is computed against the **served** revision (deterministic `row_number()` selection), keyed on `api_id` (`outdated_api_ids`) rather than the ambiguous spec-url set.
- **Endpoints** — `GET /catalog` gains `update_available` per entry, an `outdated_only` filter (mutually exclusive with `unregistered_only`), and an `outdated_count`. `ApiResponse` gains `origin`, `source_url`, and `update_available` (this PR); `catalog_api_id` is provided by #852's persisted column and consumed here. `make openapi` + UI codegen regenerated (no drift).
- **Settle** — `ImportHandler._settle_update_available` acknowledges the API's `catalog.update_available` on the handler's own write session inside a `begin_nested()` SAVEPOINT (see the flow note).

### Config (kill switch)

`catalog.update_check_interval_seconds` gates the whole thing: `0` disables the sweep **and** the scanner (no egress, no events — for air-gapped installs). Egress to the upstream is still SSRF-guarded; private subnets require the explicit `ingest.egress.allowed_private_subnets` allowlist (used by the E2E fixture).

### CLI

`jentic catalog outdated` (agent-commands envelope), an `UPDATE AVAILABLE` marker in `catalog list --registered` + the TUI browser, and a best-effort "N update(s) available" line in `jenticctl status` (uses the cached token; prints nothing rather than erroring if unavailable).

### E2E

New `ui/e2e/docker/flow3-catalog-update.spec.ts` drives the full loop against a live backend + a mutable upstream fixture (`scripts/flow3_upstream_fixture.py`): refresh manifest → import → assert not outdated → bump upstream → sweep → assert `update_available` → re-import → assert cleared **and** the event acknowledged. Wired into the `ui-e2e-realbackend` CI job (fixture launch/teardown + `JENTIC__` env overrides for the manifest URL, sweep interval, and the SSRF allowlist).

## Testing

- [x] `make check` — ruff, mypy (573 files), secrets, arch (incl. openapi-conformance / no drift)
- [x] Backend unit + arch — **2475 passed**
- [x] Backend web/registry (sqlite) — 89 passed; catalog_repo integration — 9 passed
- [x] New backend tests: `_is_upstream_tracked` classifier; sweep serialization (lock already held); outdated-vs-served-revision regressions (F1 collision, F2 asymmetry); settle acks by `api_id`; **settle reuses session — no SQLite self-deadlock**
- [x] UI: `tsc --noEmit`, eslint, prettier clean; production build succeeds; **vitest 67 passed** (discover + workspace) incl. `catalogApiId` threading, `update_available` mapping, `outdated_only` filter, `outdated_count`
- [x] `make openapi` + `npm run codegen` — no drift beyond the intended additive fields
- [x] Full-stack Flow-3 E2E added to `ui-e2e-realbackend` (loop asserted end to end incl. event ack)

## Checklist

- [ ] Commits follow Conventional Commits with a scope and are signed off (`git commit -s`, DCO)
- [x] Docs updated where relevant (`spec-flywheel-tracker.md`)

<sub>Please **squash + merge** — the PR title is the squash commit message; individual branch commits should not land on `main`.</sub>


src/jentic_one/shared/models/registry.py
```diff
@@ -35,3 +35,10 @@ class OverlayStatus(StrEnum):
 #: revision supersedes whatever is currently served. Keep the service that enqueues
 #: the materialize job and the stage that consumes it in sync via this constant.
 ORIGIN_OVERLAY = "overlay"
+
+#: Revision ``origin`` marker for a spec imported from a public-catalog entry. Set by
+#: ``CatalogService`` when it builds the import source, and load-bearing for Flow-3: the
+#: update-notify scanner treats a ``"catalog"``-origin revision (and an overlay-origin one
+#: whose ``source_url`` still points at a catalog spec) as an upstream-tracked candidate.
+#: Kept as a shared constant so the importer and the scanner classifier agree.
+ORIGIN_CATALOG = "catalog"
```
ren-jentic: i feel like this should be its own dataclass, origin can be a few things (more in future im sure) already overlay and catalog, probably "user-uploaded" or something to.

ren-jentic:
PR #912 review — Flow-3 scanner + update-available surfaces
Verdict: approve after one fix — M1 is a real stuck-badge bug in the headline feature
The architecture is right: the scanner mirrors CredentialExpiryScanner, the per-API DB gate makes scanner + manual refresh coexist cleanly, the settle-in-SAVEPOINT fix is correctly diagnosed and has a regression test that reproduces the worker's transaction frame, and the E2E finally asserts the whole loop over real HTTP. All previously flagged findings (nondeterministic candidate selection, misleading inline-fallback log, F1 spec-url collision) are demonstrably fixed. Two findings.

M1 — an upstream revert leaves a permanently stuck "Update available" badge that re-import cannot clear
_probe_one deliberately never clears last_notified_digest (the upsert docstring: "a probe that observes no change must not clear the digest that last produced an event"). That's correct for event dedupe — but this PR now keys the outdated read surface off the same field:


catalog_update_check_repo.py
Lines 69-74
            .join(served, served.c.api_id == CatalogUpdateCheck.local_api_id)
            .where(served.c.rnk == 1)
            .where(CatalogUpdateCheck.last_notified_digest.is_not(None))
            .where(CatalogUpdateCheck.last_notified_digest != served.c.spec_digest)
Walk the revert scenario:

Served revision digest D1. Upstream publishes D2 → sweep notifies, last_notified_digest = D2. Badge lights. Correct.
Upstream rolls back the bad publish → serves D1 again. Next probe: changed = (D1 != spec_digest) → False → notified_digest=None → upsert keeps D2.
_outdated_base: D2 != D1 → API stays outdated. Badge, outdated_count, catalog outdated, jenticctl status all lit.
The operator clicks Re-import — the one action the badge tells them resolves it. The import adopts D1, which is byte-identical to the served revision, so the served digest stays D1 ≠ D2. The badge never clears. The settle acks the inbox event, but every read surface stays stuck until the upstream changes again.
This is precisely the stuck-badge failure mode commit 6a133ac fixed for stale drafts, reintroduced through the revert path. The comment in _probe_one even acknowledges reverts ("a spec that reverts to the registered content stops notifying") — but only the event stops; the state doesn't. Fix is small: when upstream_digest == spec.spec_digest, overwrite last_notified_digest with the upstream digest (or null it) in the upsert — "the upstream is back in sync, nothing outstanding". That keeps event dedupe intact (a future D3 still differs from both) and makes the outdated set mean what every surface claims it means: the upstream currently differs from what you serve. One integration test: notify → revert upstream → probe → assert the API drops out of outdated_api_ids.

M2 — list and detail disagree in exactly the documented caveat case
The emit comment documents the published-over-catalog caveat: "the outdated flag correctly stays lit" when a manually PUBLISHED revision blocks re-import. But the two surfaces disagree:

List (ApiService.list): update_available=row.id in outdated_api_ids — no gate. Lit. ✓ matches the caveat.
Detail (ApiService.get): gated behind if source_url is not None on the current revision — a manually published revision has source_url=None, so the short-circuit returns update_available=False. Off.
So the API list shows the badge, the operator clicks through, and the detail ribbon shows nothing (and hides the Re-import affordance that would at least explain itself). The gate is a cheap-read optimization with a semantic cost in the one state the code itself documents as important. Either drop the gate in get (one extra query on a single-API read is nothing) or apply the same gate in the list path — but they must agree.

Nits
lock.locked() pre-check is TOCTOU. Two triggers racing past the check both reach async with lock and the second queues rather than skips, contradicting the "skip rather than queue" comment. Benign — the queued sweep no-ops on the per-API last_checked_at gate the first sweep just bumped — but the comment overstates the guarantee. (Also: the upsert docstring says "promote to ON CONFLICT if the sweep is ever hoisted onto a shared lock" — that condition is now true in-process.)
outdated_api_ids runs a row_number() window over all non-archived revisions on every GET /apis (list and single-view) request, uncached. Fine at current scale; worth remembering when a registry hits thousands of APIs × revisions, since it's on the hottest read path in the UI.
CI redirects the catalog for the entire realbackend suite — JENTIC__CATALOG__MANIFEST_URL points at the fixture for all specs in ui-e2e-realbackend, not just flow3. "Harmless to the other specs" is true today; any future spec asserting real-catalog Discover content will silently run against the one-entry fixture manifest. A comment in ci.yml is enough.
What I verified and liked
The settle deadlock fix is correct, not a workaround: reusing the handler's session inside begin_nested() is the only shape that can work under SQLite's single writer while keeping the ack atomic with the import, and the regression test drives the real _settle_update_available inside an open admin transaction — it would deadlock on the old code.
F1 fix is exact: outdated_api_ids for per-API surfaces, outdated_spec_urls retained only where the key genuinely is the manifest URL, with a collision regression test.
The served-revision ranking in _outdated_base matches registered_specs_for_notify byte-for-byte (current-first, created_at desc, id), closing my #893 determinism finding on both the write and read sides.
Scanner lifecycle is clean: kill switch checked per tick, boot-time sweep is cheap under the DB gate, cancel-with-timeout teardown, gated off standalone-registry deployments.
E2E is the real thing — conditional-GET 304 path exercised, event ack asserted (the exact path the manual E2E caught), self-skipping fixture, correct process-group reaping in CI.
M1 deserves a fix in this PR rather than a follow-up — it breaks the advertised contract ("re-importing clears it") on all four new surfaces at once, and the fix is a few lines plus one test. M2 can ride along in the same commit.

ren-jentic (changes requested):

PR #935, opened 2026-08-04
feat(cli): make jentic run launch Codex, Cursor, and Hermes as isolated agents

Manuel-Jentic:
> **Stacked on #853** (`feat/local-agent-credential-boundary`). Review/merge that first; this PR targets its branch, not `main`, so the diff shows only the operator-coverage work on top.

## Summary

`jentic run` today only launches `claude`, even though the skill layer supports Codex, Cursor, Hermes, and Generic. This closes that gap: the probe/provision/confine/seed pipeline is already descriptor-driven, so adding an operator is mostly a `localagent.Registry` row plus the per-operator specifics.

- **Three new runnable operators** — `codex` (single-binary), `cursor` (the headless `cursor-agent` CLI, install-only), `hermes` (install-only; facts from Nous Research's docs). Each gets a `Descriptor` with binary, probe paths, install command, single-binary status, and config paths.
- **Runnable vs. skill-only** — `generic` is declared skill-only (no binary), so `jentic run generic` explains itself instead of erroring as "unknown agent". A **registry-parity test** fails if any skillgen operator is neither runnable nor declared skill-only, preventing silent drift as operators are added.
- **Per-operator secret scrubbing** — new `Descriptor.SecretConfigPaths` + `ScrubSecretsCmd`/`ExpandedSecretPaths`. After config seeding, the operator's discrete credential files (Codex's `~/.codex/auth.json`, Hermes's `~/.hermes/.env`) are deleted from the agent home (home-constrained, non-recursive `rm -f`), so the agent inherits the operator's *settings* but authenticates as itself. Claude is unaffected (its key is embedded in the config the agent needs).
- **Two isolation residuals fixed** (they affect every operator, flagged in the #853 review):
  - The Linux unprivileged-userns probe now **fails closed** — only a genuine `ENOENT` counts as "enabled"; an unreadable/masked `/proc` is treated as unavailable, matching the error-closed contract.
  - The agent's own `~/.local/bin` is added to the **read-only exec routes**, so a compromised agent can't overwrite its own launched binary and re-detonate.
- **Docs/help** — `jentic run` help, `local-agent-isolation.md`, and the committed `cli-reference.json` now list the four runnable operators and note `generic` is skill-only. Bootstrap operator detection was already generic (`firstKnownAgent` → whole Registry), so the new operators became selectable automatically.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go test ./...` — all green
- [x] New tests: registry parity, descriptor well-formedness, secret-path home-constraint, scrub-command shape (no recurse/glob), agent-`~/.local/bin` read-only route (emitted after the home re-allow)
- [x] Built binary: `run generic` → skill-only message; `run nope` → unknown-agent; `run codex|cursor|hermes` pass descriptor lookup and reach account logic; `run --help` lists all four
- [ ] Not yet wired: per-operator **trusted-workspaces** detection at bootstrap (still Claude-only via `~/.claude.json`) — tracked as follow-up

## Known follow-up

Trusted-workspaces auto-detection remains Claude-specific; Codex/Cursor/Hermes each have their own trusted-projects source that isn't read yet. The isolation doc keeps that forward-looking note intact.

Reminder: squash-merge on merge, per repo git conventions.


Manuel-Jentic:
## Added: close the minor #853 review residuals (`0dd1914`)

On top of the operator-coverage work, this branch now also lands the review follow-ups that were worth doing but weren't merge-blockers. `go build`/`go vet`/`go test ./...` all green.

- **Seeded credentials no longer survive a default reset.** The keep-path reset re-owns the home to the operator but previously left behind whatever seeding copied in (`~/.claude.json`'s key, `~/.aws`, `~/.codex/auth.json`, `~/.hermes/.env`). New `SeededConfigDirs`/`ScrubSeededConfigCmd` purge those on the keep path (home-constrained, best-effort), and the banner says so. `--delete-home` still removes the whole tree.
- **Passwordless toggle is now authoritative both ways.** Declining passwordless on a reuse run removes any sudoers drop-in an earlier run installed, instead of leaving a stale `NOPASSWD` rule.
- **`--delete-home` no longer silently no-ops.** A profile reset that won't tear the account down now warns when the flag was passed, instead of ignoring it.
- **Doc/help confinement over-claims corrected.** `filesystem-access-model.md` now describes the shipped profile as a *targeted human-home deny* (permissive base + home carve-out, network/exec/`/tmp` still reachable), not an idealized `(deny default)` jail, and drops the stale "700 home as defense-in-depth" line. `jentic run` help/banner no longer imply the login shell alone isolates the agent or that sessions are "fully locked-down".

### Already addressed at head (verified, no action)
- The `RemoveSudoersCmd` substring/cross-account strip is fixed — the filter anchors on the `(<user>)` runas token, so `alice-local-agent` can't wipe `alice-local-agent-2`.
- `huh.ErrUserAborted` is cleanly handled as "Cancelled." at every bootstrap/skill selection point after the post-#824 refactor.

### Deliberately deferred (out of this stack's scope)
- **`base_url` plaintext/scheme validation** (Medium) — `httpx.New` has no error return and the default is `http://127.0.0.1:8000`; adding `https` enforcement with a loopback/dev allow-list is a cross-client signature change better done as its own focused PR than folded into the local-agent stack.
- **Reset blind to a config-less-but-OS-real account** and **profile-name-collision owner selector** — real teardown/UX gaps, but they belong on the base PR's reset surface rather than this operator-coverage stack.

Manuel-Jentic:
## 5-pass testing report (`b15fab1`)

Ran five passes over this branch — build/test, adversarial static review, live e2e, a security review of the new privileged sinks, and a doc/consistency audit. Fixed everything found; final state is **`make build` (both binaries) + `go test -race ./...` + `golangci-lint` all green.**

### Pass 1 — build/test/lint
Both binaries build; full suite passes **with the race detector**; `golangci-lint run ./...` → 0 issues. No breakage.

### Pass 2 — adversarial static review → 2 fixes
- **Latent `rm -rf $HOME` in the reset scrub.** `IsUnderHome(home, dir)` is true for the home dir itself, so a descriptor `ConfigPaths`/`SecretConfigPaths` entry of `"~"` (or anything cleaning to the home) would have made `SeededConfigDirs`/`ExpandedSecretPaths` return the home root and the reset `rm -rf` the whole home. Now require a **strict descendant** (`abs != cleanHome`). No current descriptor hits this, but the boundary check exists precisely to defend against a bad descriptor.
- **Unnecessary privileged call on fresh-account decline.** The passwordless-decline sudoers removal ran unconditionally; on a freshly created account there's never a rule, so it was a no-op `sudo` prompt. Gated to the **reuse** path only.

### Pass 3 — live e2e (built binary)
All four runnable operators pass descriptor lookup; `generic` → skill-only message; typo → unknown-agent; every `--`/leading-`--`/path/arg-count form parses correctly (incl. skill-only detection through the leading-`--` path); `run --help` renders the corrected wording. `cursor` resolved its real `cursor-agent` binary on PATH and hit the same-user warning — a live confirmation the descriptor probe works.

### Pass 4 — security review of the new privileged sinks → 1 hardening + a proof
- Added `--` before the paths in both `ScrubSecretsCmd` and `ScrubSeededConfigCmd` so a path can never be parsed as an `rm` flag.
- **Proved symlink-safety live:** `rm -rf` on a seeded path that's a symlink removes only the link (target survives), and it does not follow a nested symlink inside a real dir (`/etc/hosts` survived a scrub of a dir containing a link to it). So a compromised agent can't repoint a seeded dir (`~/.aws`, `~/.codex`) to trick the root-run reset into deleting arbitrary operator files. The scrub also runs only *after* `ValidateHomeDir`+`VerifyManagedHome` gate the home, so its paths are always built from a validated managed home.

### Pass 5 — doc/consistency audit → 1 fix
- **Corrected a fabricated Codex install URL.** The descriptor had a made-up `raw.githubusercontent.com/.../install.sh`; the official standalone installer is `curl -fsSL https://chatgpt.com/codex/install.sh | sh`. Verified Cursor (`curl https://cursor.com/install -fsS | bash`, binary `cursor-agent`) and Hermes URLs against their official docs — both correct.
- CLI-reference drift test passes; isolation-doc operator references (incl. headless `cursor-agent -p`) are consistent with the descriptors.

Net: no functional regressions, and the review-testing tightened the reset scrub's blast radius (strict-descendant + `--` + proven symlink-safety) and fixed a real fabricated-URL bug before it could ship.

Manuel-Jentic:
## Looping review — iteration 2 (passes 6–10)

Continued the testing loop with five deeper passes. All green: `go build ./...`, `go vet ./...`, `go test -race ./...`, and `golangci-lint` (0 issues) across the changed packages. Three items found and fixed (commit `43435c1`).

### Pass 6 — full-tree diff review + integration seams
Walked every changed line vs the base (`06d1160`). The `ensureAgentConfig` → `scrubSeededSecrets` wiring is consistent with the existing seed error handling (fatal on the `run` path, best-effort/warn on bootstrap). Two accuracy fixes to comments/banners (folded into the commit below).

### Pass 7 — config-seed flow end-to-end
Added a real seed→scrub contract test (`TestSeedThenScrubLeavesSettingsRemovesSecret`): for every operator that declares `SecretConfigPaths` (Codex, Hermes), materialise the secret file **plus a sibling settings file**, run the exact `ExpandedSecretPaths` scrub, and assert the **secret is gone while the settings file survives**. Guards against a wrong `SecretConfigPaths` entry pointing at a whole dir or the wrong file. `quoteProbePath`/`ProbeBinary` confirmed safe for the dashed `cursor-agent` binary and the new `~/.codex|.cursor|.hermes` probes.

### Pass 8 — reset teardown deep-dive
Scrub step (1d) ordering verified: it runs **before** the home is re-owned (still agent-owned, `sudo rm -rf` works regardless) and is correctly **skipped on the delete-home path** (the full `rm` covers it) and on legacy accounts with no recorded home. Corrected the `(1d)` comment to state truthfully that it clears **all** known operator/provider config dirs on teardown (clean-home intent; seed provenance isn't tracked), and added the clear to the plan preview so the operator consents to it up front.

### Pass 9 — confinement parity ⚠️ real bug found + fixed
`SessionAccess`: if the operator granted the agent write access to its **own** `~/.local/bin`, that path was emitted **twice** — read/write (from the grant) *and* read-only (the non-negotiable launched-binary route). The RO deny silently overrode the grant, and the path showed up in **both** the RW and RO sections of `profile view` and in the SBPL/bwrap args. Fixed by dropping the redundant read/write entry for exactly the bin dir (a grant of a **parent** like `~/.local` is still honoured). Added `TestSessionAccessGrantOfOwnBinDirIsReadOnlyNotDuplicated`.

### Pass 10 — cross-cutting + final green
skillgen↔registry parity test still holds; CLI reference in sync (`TestCommittedCLIReferenceUpToDate` passes); full `-race` suite + vet + lint all clean.

**Net:** 1 real bug (self-bin-dir grant dupe/override), 2 messaging/accuracy fixes, 2 new tests. Diff for this iteration: `cli/internal/localagent/confine.go`, `cli/internal/cmd/reset.go`, + 2 test files.

Manuel-Jentic:
## Looping review — iteration 3 (passes 11–15)

Live e2e + adversarial hardening. All green: `go vet`, `go test -race`, `golangci-lint` (0 issues). No new production defects this round; added defense-in-depth tests and a live confinement boundary proof (commit `d544c79`).

### Pass 11 — live e2e (built from PR head)
Built `jentic` from the branch and exercised the real CLI:
- `run --help` lists exactly `claude, codex, cursor, hermes` and flags `generic` as skill-only. ✅
- `run generic` → skill-only guidance (points at `jentic skill`); `run claudee`/`run gpt` → "unknown agent"; bare `run` → "missing agent identifier". ✅
- Arg-forwarding parser: both `run -- <agent> …` and `run <agent> <path> -- …` resolve the agent id correctly; `run claude /tmp extra` → "accepts at most 2 args before `--`". ✅

### Pass 12 — adversarial descriptor fuzz
New `TestScrubTargetsNeverEscapeHomeAdversarial` throws a matrix of hostile descriptor paths (`~/../../etc/shadow`, bare `/etc/passwd`, `/`, valid-prefix-then-`..`-climb, `~`, `~/`) at **both** `ExpandedSecretPaths` and `SeededConfigDirs` and asserts every survivor is a **strict descendant** of the agent home — nothing resolves to the home root or outside it. The strict-descendant guards hold under attack.

### Pass 13 — dedupe/ordering + argv safety
`SeededConfigDirs` dedupes on exact abs path and iterates `Known()` deterministically then providers — stable ordering, no dupes. Strengthened `TestScrubSeededConfigCmd` to assert the exact injection-proof argv: `sudo rm -rf -- <path>…`, each path its **own** argv element, **never** routed through a shell (so a leading-`-` path can't become a flag and no metachar is interpreted).

### Pass 14 — live Seatbelt boundary re-check (codex)
Emitted the real SBPL profile for a codex agent home under `/Users/Shared` and ran `sandbox-exec` against it:
- `/Users` + `/home` denied; agent home + grant re-opened read/write; `~/.local/bin` write-denied **after** the home re-open (last-match wins). ✅
- Live write to `~/.local/bin` via its **resolved** path → `Operation not permitted`, no file created. ✅
- (An initial "leak" was purely the `/tmp`→`/private/tmp` symlink mismatch in the test scratch dir — real homes live under `/Users`/`/home`, unaffected. Home symlink-resolution is pre-existing base-PR behavior, out of scope for this stack.)

### Pass 15 — final regression + green
Full `-race` suite + vet + lint clean on the changed packages; CLI reference still in sync.

**Net:** 0 new bugs; 2 new/strengthened hardening tests. This stack has now held stable across two consecutive fix-free security-focused iterations.

Manuel-Jentic:
## Looping review — iteration 4: real sudo-backed lifecycle (codex operator)

Ran the full privileged account lifecycle for the **codex** operator on macOS with real `sudo`, executing the exact command sequences the PR's code builds (`CreateAccountCmds`, `CopyConfigCmd`, `ScrubSecretsCmd`, `LeafGrantCmd`, `SandboxProfile`, `buildResetSteps`/`ScrubSeededConfigCmd`/`ReownHomeCmd`/`DeleteAccountCmd`). Captured pristine state first and confirmed byte-for-byte restoration at the end.

### 1. Account creation ✅
`sysadminctl -addUser codex-local-agent … -home /Users/Shared/codex-local-agent -password -` → uid 502; `createhomedir -c -u` → home `drwxr-x--- codex-local-agent:staff`; operator ACL grant inserted at index 0 with `file_inherit,directory_inherit`.

### 2. Config seed + secret scrub ✅ (the core of this PR)
Staged a realistic operator `~/.codex` (`config.toml` + `auth.json` with an API key), seeded it in with `cp -RP` + `chown -Rh`, then ran the real `ScrubSecretsCmd` (`sudo rm -f -- …/.codex/auth.json`):
- **`auth.json` (operator API key) — removed.**
- **`config.toml` (settings) — survived.**

The agent inherits the operator's Codex *settings* but not their raw key. Surgical, exactly as designed.

### 3. Grant + confinement boundary ✅
Granted the agent a leaf ACL on a workspace; the `codex-local-agent` user could read/write it. Then the **credential-boundary proof under the real SBPL profile**:
- Confined agent reading operator `~/.jentic/config.yaml` → **`Operation not permitted`**
- Confined agent `ls /Users/manuel` → **`Operation not permitted`**
- Confined agent reading its **granted** workspace → **succeeds**

Worth flagging: at the raw DAC level the agent user is in group `staff` and could `ls` the operator's `drwxr-x---` home top-level (individual secrets like `~/.ssh` at `700` stay denied). **Layer 0 confinement closes this** — under the sandbox the whole `/Users` tree is `Operation not permitted`. This is the layered model working as intended; the sandbox is load-bearing, not just belt-and-braces. (Pre-existing base-PR behaviour, not introduced by this stack.)

### 4. Reset teardown — keep-home scrub path ✅ (new behaviour in this PR)
Enriched the home with `~/.aws` (provider secret), `~/.codex` (seeded), `~/.jentic` (identity), then ran the keep-home teardown in `buildResetSteps` order:
- leaf-ACL revoke → identity removed → **`ScrubSeededConfigCmd`** (`rm -rf -- ~/.claude ~/.codex ~/.cursor ~/.hermes ~/.aws ~/.config/gcloud`) → re-own home → remove sudoers → delete account.

End state verified: **every seeded config/provider secret cleared** (`.aws`, `.codex`, `.jentic`, …), home **kept and re-owned to the operator**, account **deleted**, workspace ACL **revoked**, no sudoers residue. The `chown` non-zero on `Library/Mail`/`ContainerManager` is the expected SIP-protected-template case the `-f`/`BestEffort` handles.

### 5. Full restoration ✅
Final state matches the pristine capture exactly: no `*-local-agent` accounts, no home under `/Users/Shared`, no sudoers drop-in, operator home carries only its default `everyone deny delete` ACL, no `agent_account` in config. All test helpers/binaries removed.

**Net:** the operator-support PR's create → seed → scrub → grant → confine → reset flow works end-to-end under real root for a non-Claude operator, with the credential boundary holding under confinement and teardown leaving no secret residue. No new defects.

PR #942, opened 2026-08-04
feat(cli): guard docker runtime commands against a stopped daemon

Manuel-Jentic:
## What's this about?

If you're running Jentic One on Docker and Docker Desktop happens to be closed (super common after a reboot), the CLI used to be pretty unhelpful about it. `jenticctl start` would just dump a raw `docker compose ... Cannot connect to the Docker daemon` error at you, and `doctor` would only hint at the problem sideways through a confusing `docker compose ps` failure. You'd have to already know that the real fix was "oh, Docker isn't running."

The interesting part: our **installer** already handles this really nicely — it probes the daemon, waits patiently for a cold-starting Docker Desktop, and tells you exactly what to do. The runtime commands just never got the same love. This PR closes that gap so **every** Docker-backed command behaves consistently.

## What changed

Every `jenticctl` command that shells out to `docker compose` now checks the daemon is actually up first, and — if it isn't — fails fast with a clear, actionable message instead of a cryptic transport error:

- **`start` / `stop`** — probe before `docker compose up` / `down`. For `stop --volumes`, the check runs *before* the destructive confirmation prompt, so you fail fast rather than confirm-then-fail.
- **`update`** — probe before the (potentially multi-minute) image build, so you're not left staring at a build that was doomed from the start.
- **`setup` / `reset-password`** — probe *before* prompting for credentials on a Docker install, so you're not asked to type an email + password only to be told the daemon is down afterwards.
- **`doctor`** — reports a proper `docker daemon` warning with a fix-it hint, using a fast single-shot check so the diagnostic stays snappy and CI-safe (a down daemon is a warning, not a hard fail).

Under the hood we reuse the installer's existing probe, which politely waits (~30s) for a cold-starting Docker Desktop rather than giving up on the first ping. All five action commands print a single shared "Checking the Docker daemon ..." line first, so the wait never looks like a hang.

**The message points you at the fix**: it leads with a runtime-agnostic "start your Docker daemon", then offers the Docker Desktop 4.37+ `docker desktop start` one-liner (with the version caveat) and `colima start`, so Linux/Colima/Rancher users aren't misled.

### Deliberately left alone

- **`install`** already gates on this via the wizard's preflight (`Preflight` → `UnhealthyDaemon` → `DaemonError`), so it needs nothing here.
- **`uninstall`** is intentionally lenient: it treats compose failures as non-fatal and prints manual `docker volume rm` hints, so it can still tear down your filesystem state with the daemon down. A hard guard there would *break* graceful teardown.
- **`status`** does no Docker I/O (it's a `FileExists` check), so there's nothing to guard.

## Why fail fast instead of auto-starting Docker?

I did a bit of a survey of how the ecosystem handles this. The consistent pattern for client CLIs (Testcontainers, `act`, Dagger) is to **fail fast with a clear message** — none of them silently launch Docker for you, because on Mac/Windows it's a user app, not a managed service. Auto-starting behind your back is considered too invasive. The one thing everyone *does* do is tolerate a slow-to-wake daemon with a bounded wait, which is exactly what our reused probe does. So this PR deliberately stays in the "tell you clearly, hand you the one-liner" camp rather than starting Docker itself.

## Context

These closed issues describe the same rough edge, so linking them for the trail (not auto-closing):
- jentic-one#783 — instance stopped / Docker not running fails silently
- jentic-api-scorecard#224 — "fails if Docker is installed but not running"
- jentic-one-internal#653 — the installer hardening this mirrors on the runtime side

## Review

Reviewed by a multi-agent pass (Go correctness, tests, UX, architecture, completeness audit, Bugbot, security) — **zero blockers**. Non-blocking findings were folded in (shared announce helper to prevent wording drift, seams moved to a neutral `docker_guard.go`, daemon-probe-before-credential-prompt, consistent doctor fallthrough hint). Two deeper pre-existing probe improvements were split out as follow-ups: #953 (Ctrl-C cancellation of the ~30s wait) and #954 (distinguish "docker not installed" from "daemon down").

## Test plan

- [x] `cd cli && go build ./...`
- [x] `cd cli && go vet ./internal/cmd/... ./internal/install/...`
- [x] `cd cli && go test -race ./...` (whole module green)
- [x] New tests cover start (down + up), stop (incl. `--volumes` ordering), update, setup, reset-password, doctor, and the `install` helper
- [x] Manual sanity check: quit Docker Desktop, ran `jenticctl start` / `stop` / `stop --volumes` / `doctor` on a Docker install — all surface the friendly message (no raw compose error); `start`/`stop` exit 1, `doctor` warns and exits 0

_Heads up: please **squash + merge** to keep `main` tidy._

### DIFF
--- a/cli/internal/cmd/doctor.go
+++ b/cli/internal/cmd/doctor.go
@@ -215,7 +215,7 @@ func (d *doctor) checkDeploy(section string) {
 		// than inferred from a cryptic `docker compose ps` error (#783).
 		if detail, healthy := dockerDaemonProbe(); !healthy {
 			d.add(section, "docker daemon", statusFail, detail,
-				"start Docker Desktop (or your docker daemon), then `jenticctl start`")
+				"start it with `docker desktop start` (or open Docker Desktop), then `jenticctl start`")
 			return
 		}
 		out, err := install.ComposePs(composePath)
--- a/cli/internal/cmd/doctor.go
+++ b/cli/internal/cmd/doctor.go
@@ -198,9 +198,26 @@ func (d *doctor) checkServer() {
 	d.checkDeploy(section)
 }
 
+// dockerDaemonProbe is the seam doctor's deploy check runs through so tests can
+// simulate a stopped daemon without a real Docker. It returns a short reason
+// (empty when healthy) and whether the daemon answered. It is a fast,
+// single-round-trip probe: doctor is read-only and must not block for the full
+// cold-start polling window when the daemon is simply down.
+var dockerDaemonProbe = func() (string, bool) {
+	return install.DockerDaemonResponsiveQuick(2 * time.Second)
+}
+
 func (d *doctor) checkDeploy(section string) {
 	composePath := d.app.Paths.ComposePath()
 	if proc.FileExists(composePath) {
+		// A compose install needs a live daemon. Probe it directly so a stopped
+		// Docker Desktop is reported as an explicit, actionable failure rather
+		// than inferred from a cryptic `docker compose ps` error (#783).
+		if detail, healthy := dockerDaemonProbe(); !healthy {
+			d.add(section, "docker daemon", statusFail, detail,
+				"start Docker Desktop (or your docker daemon), then `jenticctl start`")
+			return
+		}
 		out, err := install.ComposePs(composePath)
 		if err != nil {
 			d.add(section, "deploy", statusWarn, "docker compose ps failed: "+err.Error(), "is the Docker daemon running?")
--- a/cli/internal/cmd/doctor.go
+++ b/cli/internal/cmd/doctor.go
@@ -198,12 +198,13 @@ func (d *doctor) checkServer() {
 	d.checkDeploy(section)
 }
 
-// dockerDaemonProbe is the seam doctor's deploy check runs through so tests can
+// doctorDockerProbe is the seam doctor's deploy check runs through so tests can
 // simulate a stopped daemon without a real Docker. It returns a short reason
 // (empty when healthy) and whether the daemon answered. It is a fast,
-// single-round-trip probe: doctor is read-only and must not block for the full
-// cold-start polling window when the daemon is simply down.
-var dockerDaemonProbe = func() (string, bool) {
+// single-round-trip probe (distinct from install's ~30s polling probe): doctor
+// is read-only and must not block for the full cold-start window when the daemon
+// is simply down.
+var doctorDockerProbe = func() (string, bool) {
 	return install.DockerDaemonResponsiveQuick(2 * time.Second)
 }
 
@@ -211,11 +212,14 @@ func (d *doctor) checkDeploy(section string) {
 	composePath := d.app.Paths.ComposePath()
 	if proc.FileExists(composePath) {
 		// A compose install needs a live daemon. Probe it directly so a stopped
-		// Docker Desktop is reported as an explicit, actionable failure rather
-		// than inferred from a cryptic `docker compose ps` error (#783).
-		if detail, healthy := dockerDaemonProbe(); !healthy {
-			d.add(section, "docker daemon", statusFail, detail,
-				"start it with `docker desktop start` (or open Docker Desktop), then `jenticctl start`")
+		// daemon is reported explicitly rather than inferred from a cryptic
+		// `docker compose ps` error (#783). Keep it a warning, not a fail: doctor
+		// is documented as safe to wire into CI ("warnings keep a zero exit"),
+		// and the sibling `control`/`compose ps` checks also warn on a
+		// not-running dependency — a down daemon shouldn't flip the exit code.
+		if detail, healthy := doctorDockerProbe(); !healthy {
+			d.add(section, "docker daemon", statusWarn, detail,
+				install.DockerDaemonRecoveryHint()+", then `jenticctl start`")
 			return
 		}
 		out, err := install.ComposePs(composePath)
--- a/cli/internal/cmd/doctor.go
+++ b/cli/internal/cmd/doctor.go
@@ -224,7 +224,8 @@ func (d *doctor) checkDeploy(section string) {
 		}
 		out, err := install.ComposePs(composePath)
 		if err != nil {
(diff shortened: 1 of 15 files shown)


cli/internal/cmd/update.go
```diff
@@ -417,6 +417,15 @@ func (a *App) updateStackDocker() error {
 		return fmt.Errorf("no compose stack at %s — run `jenticctl install` first", composePath)
 	}
 
+	// Fail fast with an actionable message when the daemon is down, before the
+	// long image build — otherwise the build/migrations/up sequence surfaces a
+	// raw compose transport error deep into the run. The probe may poll (~30s)
+	// for a cold-starting daemon, so announce it first (see start.go/stop.go).
+	announceDaemonCheck(a.Out)
+	if err := requireDockerDaemon("jenticctl update"); err != nil {
```
ren-jentic: **Non-blocking (two notes):**

1. *Ordering:* in a combined run, `updateCLI` swaps the binaries *before* this guard runs, so a down daemon leaves new binaries + an old stack. It self-heals — re-running `jenticctl update` after starting Docker sees the stack still behind and completes it — but probing up front in `updateE` (when the manifest is docker-mode and `doStack` is set) would be marginally more atomic.

2. *Seam consistency:* the final step of this function (line 445) still calls `install.ComposeUp` directly, while `start.go` goes through the new `composeUp` seam. Functionally fine — the daemon-down test never reaches it — but the next person writing a "daemon up" pass-through test for `update` will have to convert it anyway. Trivial follow-up.

cli/internal/cmd/stop.go
```diff
@@ -115,9 +115,18 @@ func (a *App) stopProcess(pidPath, label string, timeout time.Duration) error {
 // stack's data volumes (`down -v`), which destroys the database; it confirms
 // first unless --yes is set.
 func (a *App) stopDocker(opts *stopOptions, composePath string) error {
+	// `docker compose down` needs a live daemon; with the daemon down it fails
+	// with a raw transport error. Probe first (before the destructive --volumes
+	// confirmation prompt) so a stopped daemon yields actionable recovery
+	// guidance instead (#783). The probe can wait out a cold-starting daemon
+	// (~30s), so announce it — otherwise the command looks hung.
+	announceDaemonCheck(a.Out)
+	if err := requireDockerDaemon("jenticctl stop"); err != nil {
```
ren-jentic, line 124: **Non-blocking (UX observation):** `stop` on a down daemon now tells the user to start Docker so they can stop it — mildly amusing, since a down daemon means the stack is de facto stopped. A no-op success is arguable for plain `stop`, but explicit-fail is the safer, more honest behaviour (and clearly right for `--volumes`, where guard-before-confirm is exactly correct). Fine as is — just noting the trade-off was considered.

cli/internal/cmd/doctor.go
```diff
@@ -198,12 +198,34 @@ func (d *doctor) checkServer() {
...
+		if detail, healthy := doctorDockerProbe(); !healthy {
+			d.add(section, "docker daemon", statusWarn, detail,
+				install.DockerDaemonRecoveryHint()+", then `jenticctl start`")
+			return
+		}
 		out, err := install.ComposePs(composePath)
 		if err != nil {
-			d.add(section, "deploy", statusWarn, "docker compose ps failed: "+err.Error(), "is the Docker daemon running?")
+			d.add(section, "deploy", statusWarn, "docker compose ps failed: "+err.Error(),
+				install.DockerDaemonRecoveryHint())
```
ren-jentic, line 228: **Non-blocking (future refinement):** when the quick probe passes but `docker compose ps` still fails for another reason (e.g. malformed compose file), the hint now points at daemon recovery even though the daemon is demonstrably up. The pre-existing hint ("is the Docker daemon running?") was equally daemon-focused, so no regression — just a candidate for a more discriminating hint later.

ren-jentic (approved):
Reviewed against a local worktree at head `cb50d65` — **approve, no blockers**.

**Independent verification (all pass):**
- `go build` / `go vet` / `go test -race ./...` — 22 packages green locally, matching CI.
- Guard placement is correct: it lives inside the Docker-only branches (`startDocker` / `stopDocker` / `updateStackDocker`, and behind the `FileExists(composePath)` gate in `setup` / `reset-password`), so local venv installs never probe Docker.
- Every "deliberately left alone" claim checks out in code: `install`'s wizard preflight already probes the daemon; `uninstall` is genuinely lenient by design (warn-and-continue with manual `docker volume rm` hints — a hard guard there would break daemon-down teardown); `status` and `logs` do no Docker I/O.
- Probe arithmetic is now honest (4 × 6s + 3 × 2s ≈ 30s, matching the announced wait line).
- No cobra help text changed, so no `cli-reference.json` regen needed.
- Test quality is notably good: `TestStopDockerVolumesFailsFastBeforeConfirm` deliberately omits `--yes`, so a regression that moved the guard below the destructive confirm would *hang* the test instead of passing — it proves the ordering. The daemon-down tests also stub the compose seams with `t.Fatal`, so a bypassed guard fails loudly.

Four **non-blocking** notes left as inline comments — none worth blocking on. Squash + merge as requested.

Manuel-Jentic:
Thanks for the careful review @ren-jentic! Addressed the two update-path notes in `512645cf`:

1. **Ordering** — `updateE` now probes the daemon up front (after the confirm, before `updateCLI`) when the run is combined **and** docker-mode, so a down daemon fails fast instead of leaving new binaries against an old stack. A `daemonChecked` flag threads down to `updateStackDocker` so we don't double-probe/announce; the stack-only path still probes inside `updateStackDocker` as before.
2. **Seam consistency** — `updateStackDocker`'s final bring-up now goes through the `composeUp` seam (matching `start`/`stop`), and the daemon-down test asserts it via a `t.Fatal` stub.

The other two are intentional, as you noted:
- **`stop` on a down daemon** — keeping explicit-fail over a silent no-op (honest, and clearly correct for `--volumes` guard-before-confirm).
- **`doctor` compose-ps-fails-but-daemon-up hint** — no regression over the prior hint; left as a candidate for a more discriminating message later.

`go build` / `vet` / `go test -race` green locally. Thanks again!

PR #964, opened 2026-08-05
feat(app): in-app update banner for new releases

Manuel-Jentic:
## What this does

Tells self-hosted operators when a newer jentic-one release is available, in the two places they'd notice it:

- **Web console** — a dismissible **"update available" banner** in the app shell (shown to admins), plus the **running version in the user menu** (shown to everyone).
- **Terminal** — a one-line **"update available" nudge** on `jenticctl` / `jentic` commands, for people who live in the CLI and rarely open the console.

The running version comes from the build itself. The **latest** version is looked up from GitHub — by the **backend** for the console, and by the **CLI** for the terminal nudge. Both compare with the same semver rules, so they always agree.

### Web console

`GET /system/version` returns the running version, the latest release, and whether an update is available:

```json
{ "current": "0.26.0", "latest": "0.27.0", "update_available": true }
```

```mermaid
sequenceDiagram
    participant UI as Web console (signed in)
    participant API as GET /system/version
    participant RC as ReleaseChecker (in-process cache)
    participant GH as api.github.com

    UI->>API: GET /system/version  (session required)
    API->>RC: latest_version()
    alt cached & fresh (within TTL)
        RC-->>API: cached latest
    else stale / first call
        RC->>GH: GET /repos/{repo}/releases/latest
        GH-->>RC: tag_name (e.g. v0.27.0)
        RC-->>API: latest (cached for next TTL)
    end
    API-->>UI: { current, latest, update_available }
    Note over UI: admin → banner · everyone → version line
```

- The latest release is fetched **server-side** through the existing SSRF-guarded HTTP helper and **cached in-process** (fetch-on-read, single-flight): a burst of pollers makes at most **one** GitHub request per TTL. No background job, no database.
- The endpoint **requires a session** (any signed-in user). Both consumers already render inside the signed-in shell, so this costs nothing and keeps the exact build number away from unauthenticated scanners.
- The **banner is admin-only** (only an admin can act on it by running an update); the version line shows for everyone.
- Any failure (offline, disabled, air-gapped, remote backend) → `latest: null`, no banner. It can never break the page.

Config (`release_check` in the app config):

| Key | Default | Meaning |
|---|---|---|
| `enabled` | `true` | Turn the check off entirely |
| `repo` | `jentic/jentic-one` | Which repo's releases to read |
| `cache_ttl_seconds` | `21600` (6h) | Cache lifetime; **`0` disables** (air-gapped, no egress) |

The check only runs on a **local** backend (`server.backend == "local"`) — a hosted backend isn't something the operator can self-update.

### Terminal (CLI nudge)

```mermaid
flowchart TD
    A[jenticctl / jentic command starts] --> B{Interactive stderr?<br/>not CI / NO_UPDATE_NOTIFIER?<br/>not update/execute/help/...}
    B -- no --> Z[stay silent]
    B -- yes --> C{Cache < 24h old?}
    C -- yes --> D[use cached latest]
    C -- no --> E[git ls-remote, bounded 2s]
    E --> F[write ~/.jentic/update-check.json]
    D --> G{latest newer than running?}
    F --> G
    G -- no --> Z
    G -- yes --> H[print to stderr:<br/>Update available: X → Y — run jenticctl update]
```

- Checks GitHub **at most once per 24h** (cached to `~/.jentic/update-check.json`); the common path is a cheap file read with no network. A stale cache triggers one lookup, **bounded to 2s** so a slow network never delays the command.
- Prints to **stderr** so piped output stays clean; **suppressed** for non-interactive terminals, `CI`, `JENTIC_NO_UPDATE_NOTIFIER` / `JENTIC_NO_BANNER`, and commands that own their output (`update`, `execute`, `help`, `completion`, `install`).
- Best-effort: any error is swallowed and never blocks the command.

## Test plan

- [x] Backend unit — release checker (gating, TTL cache, single-flight, degrade), `/system/version` (auth required, surfaces update, caches, degrades, skipped on remote backend, not on broker), semver parity.
- [x] CLI — nudge (probe-then-cache, refresh when stale, keep prior tag on failure, silent first-run offline, env opt-outs, command skip set).
- [x] UI — banner (show / dismiss / re-show, hidden for non-admins), version hook.
- [x] Arch + reference-drift, `ruff`, `go vet`, `go test ./...`, `tsc`, `eslint`, Bugbot + Security Review — all clean.

## References

Design choices follow established practice for self-hosted apps and CLIs.

**Exposing version info / authenticating the check**
- OWASP ASVS v4.0.3 §14.3.3 — HTTP responses should not expose detailed component versions (CWE-200): https://asvs.dev/v4.0.3/V14-Config/ · https://owasp-aasvs4.readthedocs.io/en/latest/14.3.3.html
- GitLab version check — served from the **admin-only** `/admin/version_check.json`, populated async and short-cached: https://handbook.gitlab.com/handbook/sales/process/version-check/ · https://gitlab.com/gitlab-org/gitlab/-/merge_requests/73504

**Server-side lookup, cached (no background job)**
- GitHub REST — Get the latest release (`GET /repos/{owner}/{repo}/releases/latest`, newest non-prerelease/non-draft): https://docs.github.com/en/rest/releases/releases#get-the-latest-release

**CLI update-notifier pattern**
- GitHub CLI (`gh`) — 24h throttle via a state file, interactive-only, `GH_NO_UPDATE_NOTIFIER`, stderr notice: https://github.com/cli/cli/blob/trunk/internal/update/update.go · once-per-day rationale: https://github.com/cli/cli/pull/2488 · bounded-timeout / update-state-first discussion: https://github.com/cli/cli/issues/12599
- npm `update-notifier` — the canonical "check in the background, nudge after the command" library: https://github.com/yeoman/update-notifier


Manuel-Jentic:
### Update: `GET /system/version` now requires a session (was public)

I revisited whether this endpoint should be public and researched the trade-off. Changing it to **require any authenticated session** (no special permission).

**Why**
- Both consumers — the update banner and the user-menu version line — render **only inside the authenticated shell** (behind `AuthGuard`), so requiring a session costs us nothing.
- Public access leaked the **exact running build** to unauthenticated scanners, which is what OWASP ASVS 14.3.3 / CWE-200 fingerprinting guidance warns against (helps CVE-targeting; "raises the bar," not a hard vuln).
- Comparable self-hosted products gate their update-check data the same way or stricter: **GitLab** serves it from `/admin/version_check.json` (admin-scoped); **Nextcloud** notifies the admin group only.
- The "version discovery should be public" guidance (OpenStack / NIST SP 800-228) is about **API-protocol version negotiation** (`/v2` vs `/v3`), not publishing the exact software build — a different thing.
- Our own public `/instance` isn't a real precedent: it exposes only coarse locality + a one-way digest, never a precise, CVE-matchable build number.

Went with an **any-session** gate rather than admin-only (GitLab-style), because our banner is shown to every logged-in user, not just admins.

The spec now advertises `BearerAuth` for `getVersion`; OpenAPI + endpoint reference + UI client regenerated; tests updated (401 without a token). Graceful `latest=null` DB-degradation is unchanged.

cli/internal/cmd/updatenudge.go
```diff
@@ -0,0 +1,182 @@
...
+// once (bounded by updateNudgeTimeout) and caching the result. ok is false when
+// no tag could be determined (offline, no repo, etc.) — the caller then stays
+// silent. The cache timestamp is advanced on every probe attempt, success or
+// not, so a persistent failure re-probes at most once per interval rather than
+// on every command.
+func (a *App) latestReleaseForNudge(ctx context.Context, repo string) (string, bool) {
+	cache, _ := loadUpdateCheck(a.Paths)
+	if cache.LatestTag != "" && time.Since(cache.CheckedAt) < updateNudgeInterval {
+		return cache.LatestTag, true
+	}
```
ren-jentic: **Bug: the 24h throttle never engages on a machine where the probe has never succeeded** (air-gapped / firewalled). Every failure writes `{CheckedAt: now, LatestTag: ""}`, and an empty `LatestTag` never satisfies this guard — so **every interactive command re-probes and pays the full 2s timeout**. That contradicts this function's own docstring ("a persistent failure re-probes at most once per interval rather than on every command") and the PR body ("at most once per 24h"), and air-gapped operators are exactly the audience most sensitive to mystery per-command latency.

The test matrix confirms the gap: `TestLatestReleaseForNudgeFailureKeepsPriorTag` seeds a *stale* cache and `TestLatestReleaseForNudgeNoCacheNoNetwork` has *no* cache — neither covers "recent `CheckedAt` + empty `LatestTag`".

```suggestion
	if time.Since(cache.CheckedAt) < updateNudgeInterval {
		return cache.LatestTag, cache.LatestTag != ""
	}
```

Please also add the missing regression test: seed `{CheckedAt: time.Now(), LatestTag: ""}`, stub the prober to fail `t.Fatalf` if called, and assert `latestReleaseForNudge` returns `("", false)` without probing.

ren-jentic (changes requested):
Independent review (verified locally: read the full diff, cross-checked the Go/Python semver parity, broker mounting, and the test matrix). The design is solid and well-tested across all three surfaces — one real bug to fix before merge, plus a small hygiene note.

**Rebase note:** the branch is behind `main` (0.27.0, #966, #888 have landed) and the overlap includes generated artifacts (`openapi/control/control.openapi.yaml`, `ui/openapi.json`, generated client index). GitHub reports it mergeable, but please rebase + regenerate rather than trusting the textual auto-merge, and let the "Generated UI client up to date" check re-validate.

Nice detail: with 0.27.0 released, any 0.26 install becomes a live test of the banner the moment this merges.

cli/internal/cmd/updatenudge.go
Manuel-Jentic: Good catch — fixed in acb06f3d. The guard now throttles on `CheckedAt` age alone and returns `(cache.LatestTag, cache.LatestTag != "")`, so a recorded failure (empty tag) is trusted within the interval instead of re-probing every command. Added the regression test you described (`TestLatestReleaseForNudgeThrottlesRecentFailure`): seeds `{CheckedAt: now, LatestTag: ""}`, `t.Fatalf`s if the prober is called, and asserts `("", false)`.

Also rebased onto `main` (now includes 0.27.0 / #966 / #888) and regenerated the OpenAPI + endpoint reference + UI client — `git status` is clean after regen, so the generated-artifacts overlap is resolved. And moved the plan doc out of the public repo into `jentic-one-plans` (`prs/pr-964-app-release-update-banner.md`).

PR #980, opened 2026-08-06
cli(v2): Phase 1: The Public SDK (client/…)

ren-jentic:
## Stacked on #978 (Phase 0)

> **Stacking:** base is `cli-v2-phase-0` (#978), not `main`. Review/merge #978 first; GitHub will retarget this to `main` once the base merges. Part of the CLI-V2 roadmap (`themes/cli-v2/plan.md`).

Phase 1 builds **one typed, importable Go surface for both planes** (`client/…`), with the CLI as its first consumer. This PR lands incrementally — each commit is build/test/lint-green on its own. **Behavior-preserving: no shipped runtime path consumes the new SDK yet.**

### Commit 1 — Codegen pipeline + spec companions (`impl/2.1`)
A hermetic, drift-gated pipeline: the two committed OpenAPI specs → typed Go clients + every spec-derived companion, so adding a route/field to a spec updates the SDK, required-field validation, redaction, and discovery in **one `make generate-api`**.

- **`oapi-codegen v2.8.0`** → `client/generated/{control,broker}`. Specs are vendored into the module (`go:embed` can't cross module roots). `response-type-suffix: HTTPResp` resolves the `HealthResponse`/`LoginResponse` wrapper collision. Verified: both planes compile — **control ≈44.6k lines / 151 ops, broker ≈5.6k / 9 ops**.
- **`tools/specgen`** — one `libopenapi` walker emitting per plane: `required.gen.go` (`RequiredFields()`), `sensitive.gen.go` (`x-sensitive` table), `ops.gen.go` (op index for `jentic api ops`). `SkipExternalRefResolution` keeps CI hermetic (broker error responses `$ref` a remote problem-details schema we never consume). Companions are cross-checked against the generated client's declared types, so module-qualified Pydantic schema names oapi-codegen renames are skipped rather than emitting non-compiling methods.
- **`make generate-api`** + **`make check-cli-gen`** drift gate, wired into the `cli` CI job.
- **Arch tripwires (`impl/0.0`):** `1H` upgraded to a real `x-sensitive` sweep over the vendored specs (`impl/3.1 §1` heuristic) — fails on any **new** secret-shaped property lacking `x-sensitive`/allowlist; 18 pre-existing control-plane fields seeded into the reviewed allowlist pending the backend annotation pass. `1G` re-gated to stay dormant until the Phase-2 curated-command registry exists (it was tripping a phase early).

## Still to come on this branch (pushed as follow-up commits)
[...]

PR #981, opened 2026-08-06
cli(v2): Phase 2: The UX Strategy & Machine Contract

ren-jentic:
## What this is

**Phase 2** of the CLI V2 landing plan (`themes/cli-v2/plan.md`), stacked on
[#980](https://github.com/jentic/jentic-one/pull/980) (Phase 1). It builds the
audience-aware I/O layer and the agent machine contract, and turns on command
fencing. Everything here is **net-new foundation**: the shipped commands still
render through the legacy output path — the per-command strangler-fig cutover to
`aud.Render` is Phase 3. V1 behavior is unchanged (golden tests green).

## Plan items delivered

- **Item 1 — `Audience` interface** (`internal/cli/ux`): `Ask`/`AskConfirm`/
  `Render`/`ReportError`/`Theme`/`IsFenced`/`ForcesNoColor` with `HumanUX`
  (interactive, themed) and `AgentUX` (never prompts, no-color, JSON). Includes the
  **minimal theme registry** (`internal/theme`: `Palette`, `Themes{dark,light,no-color}`,
  `ResolveTheme` ladder) added alongside the existing brand palette, and the
  **pre-activation legacy-read adapter** in `clictx` so wrapping commands in the
  Audience is a no-op for un-migrated users.
- **Item 2 — machine contract** (`ux/contract.go`, `ux/envelope.go`): `Result`/
  `Page`/`AgentError` envelopes with `schema_version` stamping, the closed
  `error_code` enum, the exit-code taxonomy (0/1/2/3/4), and `CodedError.ExitCode()`
  so `core.Run` maps codes to exits with no per-command wiring.
- **Item 3 — fail-closed redaction** (`ux/redact.go`): three layers — typed
  `redact:"true"` + generated `SensitiveFields` registry, scoped key heuristics
  (acronym-aware `camelToSnake`, exact+suffix with a false-positive allowlist for
  `next_token`/`has_api_key`/…), and a byte backstop — applied on **both** stdout
  (`safeMarshal`) and stderr (`redactString`). `render_failed` fail-safe guarantees
  stdout always carries one valid JSON document.
- **Item 4 — command fencing** (`internal/cmd/fencing.go`, `pkg/clitree`): the
  `fenced()` annotation + root `PersistentPreRunE` interceptor that resolves mode/
  theme, constructs the Audience, and **blocks fenced management commands in agent
  mode** with a `FENCED_COMMAND` envelope. `cobra.EnableTraverseRunHooks = true`
  (the impl/3.2 §2 hazard fix). Exported `clitree.MustBeFenced` (run, reset today).
- **Item 5 — pagination UX**: `ux.NewPage` bridges `client/paginate`'s typed walk
  result into the `{schema_version, items, next_token}` envelope (cursor form only).

## Notable decisions

- **Byte-backstop shares the key definition.** Instead of impl/3.1's fixed suffix
  regex, the byte pass re-checks each JSON key through `isSensitiveKey`, so the
  structured and byte passes can never disagree (both preserve `next_token`).
- **HumanUX prompts route through `install.NewForm`/`Input`/`RunConfirm`** — the
  shared, `TERM=dumb`-safe constructors (jentic-one#841), enforced by the existing
  form-guard test — rather than raw `huh`.
- **No `--mode`/`--theme`/`--context` root flags yet.** Mode/theme resolve from
  `JENTIC_MODE`/`JENTIC_THEME`/`NO_COLOR`/config in Phase 2; the flags land with the
  context model in Phase 3, so there's no `cli-reference.json` drift.
- **Arch guard 1C is now live** (was dormant): it walks `clitree.API()` and asserts
  every `MustBeFenced` path is annotated + `EnableTraverseRunHooks` is set. It grows
  automatically as Phase 3 extends the fence set.

## Verification

- Full CLI module: `go build`, `go vet`, `golangci-lint` (0 issues), `go test` all green.
- Arch guards 1A/1B/1C(new)/1E/1F/1H pass; 1G still correctly dormant (waits for
  Phase 3's `internal/cli/api`). V1 golden contract unchanged.
- New unit tests cover all three redaction layers + false-positive protection +
  depth cap + non-mutation + fail-safe, the exit-code table, agent/human behavior,
  the theme ladder, the pagination bridge, the mode ladder, and the legacy adapter.

## Test plan

- [ ] CI green on the stack.
- [ ] Sanity: `JENTIC_MODE=agent jentic reset x` emits the `FENCED_COMMAND` JSON
      envelope on stderr and exits non-zero; human mode is unaffected.

### DIFF
--- a/cli/internal/cli/clictx/clictx.go
+++ b/cli/internal/cli/clictx/clictx.go
@@ -64,16 +64,21 @@ func ResolveActiveState(contextOverride, modeOverride string) (*ActiveState, err
 
 	return &ActiveState{
 		ResolvedState: rs,
-		Mode:          resolveMode(modeOverride, rs.PersistedMode),
+		Mode:          ResolveMode(modeOverride, rs.PersistedMode),
 		ThemeName:     rs.PersistedTheme,
 	}, nil
 }
 
-// resolveMode implements the mode ladder (impl/1.2 §resolveMode, impl/3.2 §2):
+// ResolveMode implements the mode ladder (impl/1.2 §resolveMode, impl/3.2 §2):
 // --mode > $JENTIC_MODE > persisted-context-mode > human. There is deliberately
 // NO TTY rung — mode is an explicit choice, and an unknown value is validated
 // (fail-closed to agent) at UX construction in the root interceptor, not here.
-func resolveMode(flagOverride, persisted string) string {
+//
+// Exported so the root interceptor can re-apply the same ladder on its
+// state-resolution FALLBACK path: when there is no config at all, mode must still
+// honor --mode/$JENTIC_MODE. Fencing is a safety control — it must never silently
+// degrade to unfenced human just because config resolution failed.
+func ResolveMode(flagOverride, persisted string) string {
 	if flagOverride != "" {
 		return flagOverride
 	}
--- a/cli/internal/cli/clictx/clictx.go
+++ b/cli/internal/cli/clictx/clictx.go
@@ -1,0 +1,129 @@
+// Package clictx is the CLI-only leaf that bridges the UX-free SDK config
+// (client/config) to the Cobra command layer. It exists as its own package to
+// break an import cycle: the root interceptor (impl/3.2) and the client helpers
+// (impl/4.2) both need the resolved ActiveState, but neither may live in client/
+// (SDK-clean) nor pull in the whole command tree.
+//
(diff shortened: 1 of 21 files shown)


ren-jentic:
Collapsed into #1013 (cli(v2) A — Foundation & Public SDK, Phases 0–2).

PR #986, opened 2026-08-06
cli(v2): Phase 3: Contexts, XDG & File-less Agents

ren-jentic:
Stacked on #981 (Phase 2). Part of the CLI-V2 rewrite. Plan: `jentic-one-plans` cli-v2 Phase 3 (`impl/1.2`, `impl/1.3`, `14_breaking_changes.md`, `16_landing_strategy.md`).

## What this adds

The Environment × Identity × Context management surface that supersedes flat V1 profiles, plus the one-shot migrator — all writing the XDG layout (`~/.config/jentic/config.yaml`, keys under `~/.config/jentic/keys`, tokens under `~/.local/state/jentic`).

**Coexistence, not activation.** V1 `profile`/`--profile` stays the live, default path. Per `14` rollout item 0 / `16` §1, no breaking-change switch flips on merge — the whole breaking half activates only at the post-Phase-4 activation release. These commands are additive and usable now; `migrate` is opt-in.

### Commands
- `jentic context create|use|list|view|delete` — kubectl-style context binding (env + identity + mode). `view` is active-only + read-only (agent self-introspection); `list` read-only. `create/use/delete` fenced.
- `jentic env add|list|delete` — Control/Broker deployment targets. `list` read-only; `add/delete` fenced. `delete` refuses to strand referencing contexts.
- `jentic identity add|list|delete` — actors + optional `jak_*` API-key credential (stored under XDG **state**, never in `config.yaml`). `list` read-only; `add/delete` fenced.
- `jentic migrate` — copies (never moves) legacy `~/.jentic` profiles → contexts/identities/host-derived environments, copies key/token material into the XDG layout, **drops refresh tokens** (BC-6), sanitizes names to the V2 charset, drops a `MIGRATED` marker, `--purge-legacy` to delete the old tree. Idempotent; **not fenced** (BC-1 directs agents to it); bootstrap-safe.
- `jentic theme <dark|light|no-color>` — persists the global theme (the persistence half of Phase 3 item 4; the mode gate shipped in Phase 2).

### Wiring
- Root `--context`/`--mode`/`--theme` persistent flags (deferred from Phase 2) + `$JENTIC_CONTEXT` fallback in the interceptor. Theme ladder (`--theme` > `JENTIC_THEME` > `NO_COLOR` > config > dark) already implemented in Phase 2.
- New commands render through the Audience (`ux.Result`/`ux.Page`) and fail through `ux.CodedError` so the exit-code taxonomy + structured error envelope (agent mode) are honored, with no double `error:` line.
- `client/auth`: `SaveAPIKey`/`ReadAPIKey` (XDG state, 0600) and `KeyPathForImport` (verbatim key migration, preserving the registered `client_id`).

### Guardrails
- `clitree.MustBeFenced` extended with `context create/use/delete`, `env add/delete`, `identity add/delete`. Arch **1C** walks the tree and asserts each is annotated; **1E** allowlist updated for the two new legitimate secret/state writers (`migrate.go`, `client/auth/apikey.go`).
- Regenerated `ui/public/cli-reference.json` (drift test green).

### V1 alias layer (dormant)
`profile_aliases.go` is a single removable block with **pure, tested** `profile`→`context` mapping helpers + an activation-release registration entry point. It is intentionally **not wired live** — the V1 `profile` command keeps that name until activation (`16` §1). A test asserts the shim stays dormant.

## Notes / deferred
- **API-key credential storage** is introduced here (Phase 3 needs it for `identity add` and the migrator) but its first-class use in the auth flow is Phase 4 item 4.
- **Telemetry-consent carry-over** (BC-2) is deferred until the V2 config schema models telemetry; migration preserves the legacy tree so consent is not lost meanwhile.

## Test plan
- [x] `go test ./...` (unit: env/context/identity/theme lifecycle, migrate copy/idempotent/purge/refresh-drop/api-key, name sanitization, alias mapping, apikey round-trip + traversal guard)
- [x] `make test-arch` (1C fencing, 1E mutator lock)
- [x] `make test-golden` (V1 contract unchanged)
- [x] `make check-cli-gen` (no codegen drift)
- [x] `golangci-lint run ./...` (0 issues), `go vet ./...`
- [ ] CI green on the stacked PR

### DIFF
--- a/cli/client/auth/apikey.go
+++ b/cli/client/auth/apikey.go
@@ -1,0 +1,67 @@
+package auth
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/jentic/jentic-one/cli/client/config"
+)
+
+// APIKeyPrefix is the required prefix for a Jentic API key credential. It mirrors
+// the shipped V1 constant so migrated keys and freshly-added keys validate the
+// same way.
+const APIKeyPrefix = "jak_"
+
+// apiKeyPath returns the on-disk path of the API-key credential for ref, under
+// the XDG STATE dir (a secret, not config — kept out of config.yaml exactly like
+// tokens). The dir is created 0700; the stem is the same validated
+// <identity>_<env> as keys/tokens, so a single ref addresses all three.
+func apiKeyPath(ref IdentityRef) (string, error) {
+	stateDir, err := config.StateDir()
+	if err != nil {
+		return "", err
+	}
+	if err := os.MkdirAll(stateDir, 0o700); err != nil {
+		return "", fmt.Errorf("creating state dir: %w", err)
+	}
+	stem, err := ref.Stem() // validates names — path-traversal guard
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(stateDir, stem+".apikey"), nil
+}
+
+// SaveAPIKey persists a jak_* API key credential for ref (0600). It is the V2
+// successor to V1's per-profile `apikey` file: the credential is a first-class
+// identity credential (Phase 4 item 4) but stored as a secret under XDG state,
+// never in config.yaml (which round-trips through node merges and is not
+// secret-safe). Returns an error if the key lacks the required prefix.
+func SaveAPIKey(ref IdentityRef, key string) error {
+	key = strings.TrimSpace(key)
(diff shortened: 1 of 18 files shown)


ren-jentic:
Collapsed into #1014 (cli(v2) B — Agent Runtime, Phases 3–5).

PR #990, opened 2026-08-06
cli(v2): Phase 4: Auth Hardening

ren-jentic:
Stacked on #986 (Phase 3). Part of the CLI-V2 rewrite. Plan: `jentic-one-plans` cli-v2 Phase 4 (`03_identity_roaming.md`, `impl/4.1`, `impl/4.2`, `13_agent_machine_contract.md` §4-5).

## Scope

Phase 4 is **Auth Hardening applied to the stack lifted in Phase 1**. Items 1-3 already shipped with that lift — env-scoped `<identity>_<env>.key`, the F1/F3 `requireSecureHost` transport guard (token endpoint + request editor), and the F4 renewal-skew clamp (`min(30s, lifetime/2)`) + 270s assertion TTL. This PR adds the remaining deltas.

### `jentic identity register` (item 1)
- RFC 7591 Dynamic Client Registration for the **active** identity × environment (resolved from the context, not flags).
- Generates the env-scoped Ed25519 key lazily if absent (local-only, non-destructive), POSTs the public JWKS to `/register`, and persists `client_id` + `status` per `(identity, environment)` in `config.yaml` via `MutateConfig`.
- Hand-rolled POST (not a generated call): `/register` and `/oauth/token` are auth-server routes **not** in the control-plane OpenAPI surface — this matches the shipped V1 `internal/authclient` contract. Held to the same TLS/loopback guard as the token endpoint.
- **Fencing carve-out**: bootstrap-safe, not fenced — an agent legitimately registers its OWN provisioned identity during autonomous bootstrap. Never switches contexts or mints another identity.

### API-key credential first-class in the flow (item 4)
- `AttachAuth` now presents a stored `jak_*` key as `Authorization: Bearer jak_...` **verbatim** (no exchange/minting), ahead of the OAuth path. Confirmed against the backend (`shared/web/auth.py`, `shared/auth/api_key_resolver.py`): the backend prefix-dispatches on the key, and V1 (`agentauth.Session.ValidToken` → `httpx`) already sends it on `Authorization: Bearer`.
- Presence of the on-disk credential is the signal; a read miss falls through to the exchange. Injected token still wins over an API key.

### SDK response-side policy (item 5, `13` §5)
A `retryTransport` wrapping both planes' HTTP client:
- **401** → exactly one token re-exchange + one retry, and **only** for re-exchangeable creds (the JWT-bearer disk path). Injected tokens and API keys are fixed — a 401 is a hard denial, not looped. On 401 it invalidates the cached token (server rejected what looked valid to us) and re-stamps auth via the shared `AttachAuth`.
- **429** → honor `Retry-After` within a bounded budget (`maxRetryAfter`), else surface the 429.
- **5xx / transport error** → bounded backoff for **idempotent** calls only: GET/HEAD, or a POST carrying an `Idempotency-Key` (the broker dedupes on it). A plain POST is never blind-retried. Bodies are buffered so retries can resend.

### Plumbing
- `clictx.GetControlClient`/`GetBrokerClient` map the resolved `ActiveState` → `client.Config` (broker URL never derived from control URL).
- `auth.InvalidateTokens`, `auth.AttachAuth`, `auth.CanReExchange`, `auth.Register`/`RegistrationResult` added; `RequestEditor` now delegates to `AttachAuth` so the editor and the 401-retry path can never disagree on which credential to present.

## Notes / deferred
- The `X-Jentic-Session-Id` telemetry editor and `Idempotency-Key` *emission* by `jentic execute` are Phase 5; this PR only makes the transport retry-aware of an Idempotency-Key when present.
- Test harness: `withXDG` now isolates `$JENTIC_HOME` so the legacy-read adapter can't pull a developer's real `~/.jentic` into command tests (latent flake, fixed here).

## Test plan
- [x] `go test -race ./...` (register mock DCR, PublicKeyToJWKS shape, API-key-in-flow precedence + transport guard, CanReExchange, InvalidateTokens, retry policy: 5xx idempotent/non-idempotent/idempotency-key, 429 within/over budget, 401 exchangeable/fixed, clictx adapter)
- [x] `make test-arch` (1C fencing, 1E mutator lock), `make test-golden`
- [x] `make check-cli-gen` (no codegen drift)
- [x] `golangci-lint run ./...` (0 issues), `go vet ./...`
- [x] `ui/public/cli-reference.json` regenerated (register entry; drift test green)
- [ ] CI green on the stacked PR

### DIFF
--- a/cli/client/auth/middleware.go
+++ b/cli/client/auth/middleware.go
@@ -27,39 +27,76 @@ func (c Credentials) IdentityRef() IdentityRef {
 // the bearer token, fetching/refreshing it as needed.
 func RequestEditor(creds Credentials) func(ctx context.Context, req *http.Request) error {
 	return func(_ context.Context, req *http.Request) error {
-		// 0. TRANSPORT GUARD (ref F3). Before attaching ANY bearer — injected or
-		// minted — refuse a non-HTTPS, non-loopback target. req.URL is the actual
-		// outbound destination derived from a possibly attacker-influenced BaseURL;
-		// without this, a poisoned base URL turns the middleware into a
-		// token-exfiltration primitive. Same invariant tokenEndpoint enforces (F1).
-		if err := requireSecureHost(req.URL); err != nil {
-			return fmt.Errorf("refusing to attach credentials: %w", err)
-		}
+		return AttachAuth(creds, req)
+	}
+}
 
-		// 1. FILE-LESS OVERRIDE. If the orchestrator injected a token, use it.
-		if creds.InjectedBearerToken != "" {
-			req.Header.Set("Authorization", "Bearer "+creds.InjectedBearerToken)
-			return nil
-		}
+// AttachAuth stamps the appropriate Authorization header on req for creds. It is
+// the shared attach path used by both the request-editor middleware and the
+// response-side 401 retry (transport.go), so the two can never disagree on which
+// credential to present. Order: transport guard -> injected token -> API-key
+// credential -> disk token (exchanged if missing/expired).
+func AttachAuth(creds Credentials, req *http.Request) error {
+	// 0. TRANSPORT GUARD (ref F3). Before attaching ANY bearer — injected or
+	// minted — refuse a non-HTTPS, non-loopback target. req.URL is the actual
+	// outbound destination derived from a possibly attacker-influenced BaseURL;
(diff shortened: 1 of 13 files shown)


ren-jentic:
Collapsed into #1014 (cli(v2) B — Agent Runtime, Phases 3–5).

PR #998, opened 2026-08-07
fix(install): harden the Docker install path (#992)

### LINKED ISSUE #992: Docker install: wizard bind-host ignored (ports published on 0.0.0.0 incl. Postgres), 0600 files unreadable by container uids, initdb failure poisons the volume
## Origin

External install feedback (2026-08-06): fresh `curl | sh` install on a Hetzner CX23, Ubuntu 24.04, Docker path, interactive wizard with all defaults accepted and loopback-only binds chosen (127.0.0.1:8000 app, 127.0.0.1:8100 broker), telemetry accepted. Four issues were hit in sequence getting from install to a running stack. All were verified against the code (file:line references below); three reproduce, one does not.

### 1. Wizard writes files the app containers can't read

> `init-schemas.sql`, `jentic-one.yaml`, and `config.yaml` were all written to `~/.jentic/` as `0600`, owned by the installing user (uid 1000). But the Postgres and app/broker containers run as their own uids (postgres=999, jentic=999) — neither is uid 1000, so every one of those files was unreadable by the process that needed it, and it wasn't the same failure each time (permission denied reading the SQL init script, then the app config, then the log directory).
>
> Fix applied locally: `chmod 644` on the yaml/sql files, `chown 999:999` on the logs directory (644 wasn't enough there since the container user isn't in the host's file-owner group).

### 2. Compose filename mismatch in wizard output

> The wizard's on-screen instructions say `docker compose -p jentic -f ~/.jentic/docker-compose.yml run ...` but the file it actually writes is `docker-compose.yaml` (not `.yml`). Copy-pasting the command as shown fails with "no such file or directory."

### 3. Postgres init script doesn't rerun after a mid-init failure

> Because of #1, the very first `db` container boot failed partway through `docker-entrypoint-initdb.d` (permission denied on the SQL file), *after* Postgres had already created its data directory. Postgres only runs init scripts once, on an empty data dir — so fixing the file permissions afterward didn't help; the schema-creation step was permanently skipped for that volume. Only realised this by noticing later migration errors (`schema "registry" does not exist`) and working out the container had a half-initialized volume. Had to `docker compose down -v` (destroy the Postgres volume) and let it reinit from scratch.

### 4. Wizard's loopback choice isn't wired into the actual Docker ports

> The wizard asks for `bind: 127.0.0.1` on both the app and broker (screen literally says "Host and port the app binds to"), and loopback was picked deliberately. But the generated `docker-compose.yaml` publishes ports as bare `"8000:8000"`, `"8100:8100"`, `"5432:5432"` — no host-IP prefix. Docker always publishes unqualified port mappings on all interfaces, so this **overrides the wizard's choice and exposes Postgres and both app/broker ports to the public internet**, regardless of what was selected on the "Server" screen.
>
> We caught this only by running `docker compose ps` afterward and noticing `0.0.0.0:8000->8000/tcp` in the PORTS column. UFW was active with default-deny incoming, but that did not protect against this — Docker manages its own iptables rules and bypasses UFW's filtering for published ports, so the exposure should be treated as real regardless of the host firewall state.
>
> Fixed locally by hand-editing the compose file to `"127.0.0.1:5432:5432"` / `"127.0.0.1:8000:8000"` / `"127.0.0.1:8100:8100"`, confirmed via `ss -tlnp` that all three now bind loopback-only.
>
> **This is the one I'd want fixed before recommending the wizard to anyone else.**

> Otherwise: once past the above, install was smooth — migrations ran clean across all three schemas, app and broker both came up healthy, no other surprises.

## Verification against the code

**Report item 4 — confirmed (worst).** The Server section prompts for "Bind host" (`cli/internal/install/sections.go:251-258`, default `127.0.0.1` at `cli/internal/install/draft.go:168`), but `ServerHost` is not a field of `composeParams` (`cli/internal/install/compose.go:113-134`) and never reaches the template. All three publishes are bare: db `"{{.PGPublishedPort}}:5432"` (`compose.go:149`), app (`compose.go:171`), broker (`compose.go:197`). `render.go:230-236` deliberately rewrites the in-container bind to `0.0.0.0` — correct for the container bind, but the host-publish address was conflated with it and the wizard answer is silently dropped. One correction to the report: Postgres is not `trust` auth (`POSTGRES_PASSWORD` is set, `compose.go:147`) — but the wizard default password is the literal string `postgres` (`draft.go:165`), so an internet-exposed 5432 is equivalent in practice.

**Report item 1 — confirmed.** `jentic-one.yaml` written `0600` (`cli/internal/cmd/install.go:147-150`); `docker-compose.yaml` and `init-schemas.sql` `0600`, logs dir `0700` (`cli/internal/install/compose.go:259-272`); `~/.jentic` dirs `0700` (`cli/internal/config/paths.go:191`). The app/broker images run as the `jentic` system user, uid 999 (`deploy/docker/python-base.Dockerfile:35`, `app.Dockerfile:14`, `broker.Dockerfile:7`); postgres runs as uid 999. There is no chmod/chown for container uids anywhere in the install path. Docker Desktop's VM uid remapping masks this on macOS; native Linux fails.

**Report item 3 — confirmed.** `init-schemas.sql` is mounted into `/docker-entrypoint-initdb.d` (`compose.go:152`), which runs once on an empty data dir. The healthcheck is liveness-only (`pg_isready`, `compose.go:153-157`); the schema probe (`ComposeSchemaState`, `compose.go:543-563`) degrades to "continuing" on an unknown verdict by design. A mid-init failure therefore poisons the volume silently, surfacing later as a raw `schema "registry" does not exist` with no recovery hint.

**Report item 2 — not reproducible.** The compose file has always been `docker-compose.yaml` (`cli/internal/config/paths.go:85-88`); the printed commands interpolate the real path (`cli/internal/install/commands.go:195`, `cli/internal/cmd/start.go:149`); `git log -S 'docker-compose.yml'` finds nothing. Likely a paraphrase in the (AI-compiled) report. Covered anyway by the "printed commands are executable" test below.

## Fix

Patch release first (the exposure fix), remainder next minor.

### Patch release

1. **Wire the bind host into compose port publishes.** Add a bind-host field to `composeParams` and prefix all published ports (`"{{.BindHost}}:{{.ServerPort}}:{{.ServerPort}}"`, same for broker and db), sourced from the wizard's `ServerHost` answer. Keep the in-container `0.0.0.0` rewrite in `render.go` (that part is correct). Validate the Docker-path answer is an IP; map `localhost` → `127.0.0.1`.
2. **Make Docker-path artifacts readable by container uids.** Write container-read files (`jentic-one.yaml`, `init-schemas.sql`) as `0644` — `~/.jentic` stays `0700`, which still protects them from other host users, and bind mounts don't require parent-dir traversal. Keep `config.yaml` (CLI-only, never mounted) and local-path files at `0600`. Logs dir needs a uid-999 *write*, which mode bits alone can't grant: prefer switching to stdout/stderr logging (`docker compose logs`) and dropping the host bind; if too large for the patch, interim workaround + follow-up issue.
3. **`jenticctl doctor` check**: compare the configured bind host against actual published ports (`docker compose ps` / `docker inspect`) and flag `0.0.0.0` publishes when the config says loopback. This is the only mitigation that reaches existing installs.
4. **Release note for existing Docker installs**: regenerate via `jenticctl install` or hand-prefix the three port mappings; rotate the Postgres password if it is `postgres`.

### Next minor

5. **Stop publishing 5432 by default.** The app/broker reach the db over the compose network; make host exposure an explicit wizard opt-in, omitting the db `ports:` entry otherwise.
6. **Randomize the Postgres password** for new installs — it is machine-managed and never typed, so a generated secret has no UX cost.
7. **Idempotent schema creation.** Move `CREATE SCHEMA IF NOT EXISTS registry/control/admin` into the migration entrypoint and delete `init-schemas.sql` + its initdb mount entirely — no run-once wedge, and a half-initialized volume self-heals on the next migrate.
8. **Fresh-install failure recovery.** On a fresh install (the installer just created the volume; no data to protect), tear down with `down -v` automatically on migration failure — or at minimum print the literal recovery command.

### Regression guards (land with or before the patch)

- **Native-Linux e2e install job in CI**: fresh Ubuntu runner, non-root user, real install path; assert stack health, `ss -tlnp` bindings match the wizard's bind-host answer, generated artifacts are readable by container uids, and every command the CLI prints is executable as printed.
- **"Wizard answers are contracts" unit test**: render compose + config with non-default answers for every wizard field and assert each appears in the output — makes silently dropped answers (the root cause of item 4) structurally impossible.
- **Template lint assertions**: no port publish without a host-IP prefix; no credential field with a well-known default.

## Root cause (process)

All three real issues are native-Linux behaviours invisible on Docker Desktop/macOS: bind-mount uid mapping, Docker's iptables bypass of UFW, initdb run-once semantics. The e2e Linux CI job above is the structural fix for the class.

ren-jentic:
## Summary

Fixes #992 — the full train of hardening fixes from the Hetzner/Ubuntu 24.04 install report, one commit per part:

1. **Bind-host honoured in port publishes + container-readable artifacts** — the wizard's bind-host answer now prefixes every compose port mapping (Docker publishes unqualified mappings on 0.0.0.0 and bypasses UFW), and generated artifacts get modes the container uid (999) can actually read/write. A new contracts test makes silently-dropped wizard answers structurally impossible.
2. **`doctor` exposure check** — warns older installs whose compose file (or live containers) publish ports on all interfaces, with the regenerate/re-up fix spelled out.
3. **Headless install** — `jenticctl install --defaults` / `--answers <file>` for CI and scripted deployments, validated with the exact wizard rules.
4. **Native-Linux e2e install job** — installs headlessly on `ubuntu-latest`, then asserts loopback binds on the host, container-readable artifacts, idempotent schema migration, a clean `doctor`, and answers-file propagation against the live stack.
5. **Idempotent schema bootstrap** — schema creation moved from the run-once `docker-entrypoint-initdb.d` script into the Alembic migration runner (`CREATE SCHEMA IF NOT EXISTS` on every run), eliminating the poisoned-volume class; stale `init-schemas.sql` files are cleaned up.
6. **Postgres publish opt-in + random password** — the managed db is no longer published on the host by default (the app/broker reach it over the compose network); opting in is an explicit wizard confirm / answers key. The Docker path's password is a generated random credential instead of `postgres`, carried over on reinstall because `POSTGRES_PASSWORD` only applies at initdb.
7. **Fresh-install migration-failure recovery** — if the first migration fails, a volume created by that very run is discarded automatically so a re-run starts clean; a pre-existing volume is never touched (manual `down -v` command + backup warning instead).

## Test plan

- [x] `go test ./...`, `gofmt`, `go vet`, `golangci-lint run` clean in `cli/`
- [x] `pytest tests/integration/test_migrations_bootstrap.py` (idempotent schema creation against real Postgres)
- [ ] `e2e-install-linux` CI job green on this PR (exercises the full headless install on native Linux Docker)
- [ ] Manual: wizard flow on a Docker install shows the new database exposure confirm and the summary reflects it


ren-jentic:
Closing: this was the old-CLI install-hardening train. Its full fix set (idempotent Alembic schema bootstrap, loopback-bound port publishes, opt-in Postgres exposure, random PG password, container-readable artifacts, migration-failure recovery, exposure doctor check) has been ported into the CLI-V2 rewrite and lands via #1016 (cli(v2) D — Hardening & Cross-Platform, Phases 8–9), reconciled to the decomposed internal/cli/ctlcmd tree.

PR #1001, opened 2026-08-07
cli(v2): Phase 8: Conformance Reconciliation & Binary Decomposition

ren-jentic:
## Summary

Phase 8 closes the gap between the CLI-V2 plans/impl guides and the code that
actually landed. The headline is the **binary decomposition the guides mandated
from the start** (`impl/1.1 §1a`, `06 §1`, `07 §3`) but which Phase 1 shipped as
a single flat `internal/cmd` package.

Because Go links a package as a unit, the flat package forced the **`jentic`
(agent/data-plane) binary to link the operator/installer packages**
(`internal/install`, `internal/proc`, `internal/update`) — breaking the airgap
that keeps the agent surface small and unprivileged.

### What changed

- **`internal/cmd` → `internal/cli/{cmdcore,api,ctlcmd}`.** `App` is a thin
  dependency container, so the split is mechanical (behaviour-preserving
  relocation + package/receiver renaming via a tree-local
  `type app struct{ *cmdcore.App }` embed):
  - `cmdcore` — shared base: the `App` container, root wiring, help/header/
    output/fencing/updatenudge, cross-tree helpers (agent-account setup,
    register/bootstrap, skill, identity target). Version ldflags vars live here.
  - `api` — the `jentic` tree. **Imports zero installer packages**; interactive
    prompts route through the new `internal/cli/prompt` package.
  - `ctlcmd` — the `jenticctl` tree (install/wizard/setup/doctor/status/start/
    stop/logs/update/uninstall/…); legitimately links the operator packages.
    Named `ctlcmd` to avoid clashing with the existing `internal/cli/ctl`
    (Phase-6 config machinery).
- **Shared form helpers extracted** from `internal/install` into
  `internal/cli/prompt` — the last soft coupling that forced the agent tree to
  link the installer. `install` re-exports them during the transition.
- **`pkg/clitree`** now composes `api.TreeBuilder()` / `ctl.TreeBuilder()`; the
  CLI-reference builder and skill-freshness gate move here (they need both
  trees). Entrypoints, `.goreleaser.yaml`, `Makefile`, `tools/install.sh`
  ldflags rewired to the new package paths.
- **New arch gate `Test8_AgentBinaryExcludesOperatorPackages`** locks the
  boundary by import-graph analysis: the agent surface + `cmd/jentic` must have
  an empty transitive intersection with the operator-only packages, with an
  anti-vacuity check that `cmd/jenticctl` *does* link them.
- **`Test1G_SpecFlagCoverageParity`** probe re-pointed at the real registry
  artifact (`internal/cli/api/curated.go`) so the Phase-8 binary split doesn't
  trip it spuriously — it stays correctly dormant (no curated registry exists)
  with its loud tripwire intact.

### Guarantee

`go list -deps ./cmd/jentic | grep internal/(install|proc|update)` → **empty**;
the same for `./cmd/jenticctl` still lists all three.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test -race ./...` green (behaviour-preserving; every command/flag/
      output string unchanged)
- [x] `golangci-lint run ./tests/arch/...` clean
- [x] arch guardrails (SDK boundary, stdout/stderr, config-mutator lock, print,
      new binary boundary) green
- [x] new binary-boundary gate proven non-vacuous (injecting an installer
      import into the api tree fails it)
- [x] golden/V1-characterization + external SDK-consumer contract green


ren-jentic:
Collapsed into #1016 (cli(v2) D — Hardening & Cross-Platform, Phases 8–9).

PR #1021, opened 2026-08-11
feat(auth): add SSO login seams — Google provider, provisioning hook, superset verifier

ren-jentic:
## Summary

Adds the minimal **OSS-side seams** for external-IdP (SSO) human login so a deployment can offer "Continue with Google" and control who gets provisioned — without forking the auth flow. Everything defaults to **today's behaviour**: with no overlay and no IdP configured, login is byte-identical to now. The admission-policy engine, provider registry, and multi-user console are deliberately **not** in OSS (they belong to the overlay tier). Follows the `jentic-one-plans` OSS impl guide (`research/enterprise-sso-google-oauth-oss-impl.md`).

The whole OIDC flow already exists in OSS (`/authorize` PKCE → `/oauth/callback` → `/oauth/token`, `OidcAdapter`, `external_identities`, the `at_`-capable verifier). This PR only adds four small, additive seams on top.

## Changes

- **`auth` — provisioning-decision seam** (`auth/core/idp/provisioning.py`, new): an injectable `AdmissionPolicy` (module-level registry `set_/get_admission_policy`, mirroring OSS's `register_config` pattern). Default `open_admission_policy` admits **every** brand-new email — exactly today's behaviour. `_resolve_or_create_user` consults it only at the brand-new-user step; already-linked / existing-account / unverified-collision fail-closed logic is untouched. Reject raises `UserNotAdmittedError` (new), audited in its own committed transaction; the callback redirects to `/error?error=access_denied`.
- **`auth` — Google provider** (`idp/oidc.py`, `idp/adapter.py`, `shared/config.py`): `provider: google` defaults Google's authorize/token/userinfo endpoints (still overridable) and `map_claims` surfaces the `hd` claim → new `IdpClaims.hosted_domain` + `IdpConfig.hosted_domain`. Generic `oidc` is unchanged. No provider registry.
- **`shared` — deterministic verifier** (`auth/web/app.py`, `shared/web/app_factory.py`): exposed `make_superset_verifier(ctx)`; `create_combined_app` installs it whenever the `auth` surface is enabled, so `at_` tokens resolve on admin routes **regardless of surface ordering** (captured dynamically — no static `shared→auth` import, no inline import).
- **`ui` — login entry point**: public `GET /auth/idp` capability hint (tagged Discovery, classified public), a "Continue with Google" button shown only when the backend advertises an IdP, `pkce.ts` + `sso.ts` + `api/idp.ts`, an `/auth/callback` route (`SsoCallbackPage`) that exchanges the code and feeds the existing `setSession` (token store unchanged). `AuthContext.loginWithSession` reuses the password-login adoption path.
- **generated artifacts**: regenerated `openapi/control/control.openapi.yaml`, `docs/reference/endpoints.{md,json}`, and the UI client (`ui/openapi.json` + `DiscoveryService`).

**Out of scope (overlay tier):** the admission-policy *engine* (rules/modes/domains, invite-vs-domain composition), the multi-user console/invite UI, and a provider registry for Okta/Azure/SAML. OSS ships only the hook + open default.

## Risk & rollback

- **Auth / provisioning surface** — the seam sits on the external-IdP new-user path. Default policy admits every email (== today); verified regression test (`test_unverified_email_without_collision_creates_user`) confirms no behaviour change with no overlay.
- **New public route** `GET /auth/idp` — unauthenticated, secret-free (returns only `enabled` + provider name); classified Public in the endpoint reference.
- **New config field** `auth.idp.hosted_domain` (optional, default `None`) — OSS surfaces the `hd` claim but does not enforce it.
- **Combined-app verifier**: now deterministically the auth superset when `auth` is enabled. This is a hardening (previously order-dependent); admin-only deployments are unaffected.
- Rollback: revert the squash commit — all changes are additive and behind config/default-open.

## Test plan

- Backend: `uv run ruff check` + `uv run mypy` clean on all changed modules.
- `uv run pytest tests/arch tests/unit/auth --no-cov` → **485 passed** (incl. new `test_provisioning_policy.py`, Google-profile cases in `test_idp_adapter.py`, `test_combined_verifier.py`).
- `uv run pytest tests/web/auth/test_oidc_login_flow.py -m integration --no-cov` (Postgres fixtures) → **7 passed**, incl. new `/auth/idp` descriptor + reject-policy end-to-end tests.
- UI: `npx tsc --noEmit` + `eslint` clean; `vitest run` auth suites → **34 passed** (5 new SSO specs in `sso-login.test.tsx`: button hidden/shown, full callback → session → dashboard, error paths).
- Verified generated artifacts are idempotent (`make openapi && make endpoints` produce no further diff).

### DIFF
--- a/src/jentic_one/auth/core/idp/oidc.py
+++ b/src/jentic_one/auth/core/idp/oidc.py
@@ -1,7 +1,14 @@
-"""Generic OIDC identity provider adapter."""
+"""OIDC identity provider adapters.
+
+`OidcAdapter` is the generic, standards-compliant base. Provider-specific
+adapters subclass it to supply well-known endpoint defaults and any
+provider-specific claim handling (e.g. Google's `hd`). Construct the right one
+for a config via the factory in `jentic_one.auth.core.idp.factory`.
+"""
 
 from __future__ import annotations
 
+from dataclasses import replace
 from urllib.parse import urlencode
 
 import httpx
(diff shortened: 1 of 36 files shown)


src/jentic_one/auth/core/idp/oidc.py
```diff
@@ -9,33 +9,54 @@
...
+# doesn't override the endpoints explicitly, these are used so a tenant only has
+# to supply client_id/client_secret. Google is a standards-compliant OIDC
+# provider, so the generic adapter handles it — this is just endpoint defaulting.
+_GOOGLE_ISSUER = "https://accounts.google.com"
+_GOOGLE_AUTHORIZATION_ENDPOINT = "https://accounts.google.com/o/oauth2/v2/auth"
+_GOOGLE_TOKEN_ENDPOINT = "https://oauth2.googleapis.com/token"
+_GOOGLE_USERINFO_ENDPOINT = "https://openidconnect.googleapis.com/v1/userinfo"
+
 
 class OidcAdapter:
```
ren-jentic, line 20: instead i want GoogleOIDCAdapter and a factory to get the right adaptor basd on cfg, this is hacky and not ideal for when we want to add other oidc providers

ren-jentic: Done in da2591b. Refactored so the generic `OidcAdapter` no longer knows anything about Google:

- `OidcAdapter` is purely generic — endpoint defaults derive from the issuer via overridable `_default_*_endpoint` hooks, and it never sets `hosted_domain`.
- `GoogleOidcAdapter(OidcAdapter)` overrides those hooks with Google's well-known endpoints and surfaces the `hd` claim in `map_claims`.
- New `build_idp_adapter(config)` factory (`idp/factory.py`) maps `provider` → adapter class via a small registry (defaults to the generic `OidcAdapter`). A future `OktaOidcAdapter` slots in with a single registry entry — no changes to the authorize service.
- `authorize_service._get_idp_adapter` now just calls the factory.

Tests updated to target `GoogleOidcAdapter` + the factory, plus a case asserting the generic adapter never sets `hosted_domain` even if `hd` is present.

PR #1042, opened 2026-08-13
feat(auth): add agent ownership-claim primitive

ren-jentic:
## Summary

Self-registered (DCR) agents are created anonymously (`POST /register` is unauthenticated per RFC 7591), so they land with `owner_id = NULL` and nothing binds the pending agent to the human who registered it. This adds an **opt-in** ownership-claim flow: `/register` can mint a single-use claim token, and any authenticated caller presents it to `POST /agents/{id}:claim` to set themselves as the owner. Once owned, the existing scoping and approve paths just work — an operator approving later no longer silently becomes the owner.

The whole flow is **dormant by default**: the claim-token minter seam mints nothing unless a deployment installs its own minter, so single-user behaviour and the `/register` response are byte-for-byte unchanged.

## Changes

- **`auth/core/claim.py` (auth)** — new `ClaimTokenMinter` seam (`set_claim_token_minter` / `get_claim_token_minter`); default `no_claim_token` returns `None`. Mirrors the existing IdP admission-policy seam.
- **`admin/core/schema/agents.py` + migration `e1f2a3b4c5d6` (admin)** — nullable `claim_token_hash` / `claim_expires_at` columns (+ index on the hash). Kept deliberately **separate** from the RAT columns, which `set_approval` nulls — reusing them would drop the claim token at approval time.
- **`RegistrationService.register` + `RegisterResponse` (auth)** — asks the seam for a token; if one is returned, stores only its sha256 hash + expiry and returns the plaintext **once** on `RegisterResponse.claim_token`. Minted lazily (needs the agent id) but cached so a transaction retry reuses the same plaintext.
- **`POST /agents/{id}:claim` + `AgentService.claim` (auth)** — authenticated by the platform bearer token but requires **no** agent permission; the single-use token is the proof, so a plain member can claim the agent they registered. Locks the row, constant-time-compares the hash (treats "no token issued" like a mismatch so claimability isn't leaked), checks expiry + unowned, sets `owner_id`, consumes the token, and audits with a new `AuditAction.CLAIM`.
- New service errors `ClaimTokenInvalidError` (400) / `AgentAlreadyOwnedError` (409); new `auth.claim_ttl_seconds` config (inert without a minter).
- Regenerated `openapi/`, `ui/openapi.json`, `docs/reference/endpoints.*` — `:claim` is classified as *any authenticated* / no scope.

**Out of scope:** the CLI printing the claim pointer (deferred to the CLI v2 rewrite). `/register`, `:approve`, and the visibility filter are untouched.

## Risk & rollback

- **Migration:** two nullable columns + one index on `agents`; no backfill, no rewrite. Existing agents simply have no claim token.
- **Auth/permissions surface:** new endpoint is authenticated but scope-free by design — the token is the credential. It only mutates a still-unowned agent and consumes a single-use token, so it cannot re-assign an already-owned agent.
- **Default behaviour unchanged:** with no minter installed, `/register` mints nothing and the response is identical to before.
- Rollback: revert the squash commit; the migration `downgrade()` drops the columns/index.

## Test plan

- `uv run pytest tests/unit/auth tests/arch --no-cov` — 507 passed (seam default/set/last-wins, register mints-none/mints-when-installed, claim happy/owned/wrong/no-token/expired/archived/unknown, all arch invariants).
- `uv run pytest tests/web/auth/test_agents.py --no-cov` — 25 passed, incl. real-app claim: sets owner, single-use replay → 409, wrong token → 400, unauthenticated → 401.
- `uv run pytest tests/unit/test_migrations.py tests/unit/test_migration_status.py --no-cov` — migration/model drift check passes; single admin head.
- `uv run mypy` on changed modules, `uv run ruff check` / `ruff format` — clean.
- `make openapi` / `make endpoints` — regenerated, no manual edits.


ren-jentic (commented):
## Review (agent-assisted; would be "request changes" for #1 but GitHub disallows that on own PRs)

## Verdict

Well-shaped PR. It faithfully implements the owner-attribution plan: dormant-by-default seam (mirrors the IdP admission-policy pattern), dedicated claim columns kept separate from the RAT pair (the plan's key trap, correctly avoided and well-commented), hash-only storage, single-use consumption under a row lock, constant-time compare, clean error taxonomy, regenerated OpenAPI/docs, and solid unit + real-app web tests. One real bug and a few smaller items — **#1** should land before merge; the rest can be follow-ups.

## Findings

### 1. Non-user actors can call `:claim` → FK violation → 500 (bug)

`Agent.owner_id` is a FK to `users.id` (`admin/core/schema/agents.py`):

```python
owner_id: Mapped[str | None] = mapped_column(
    String(30),
    ForeignKey("users.id", ondelete="RESTRICT"),
    nullable=True,
)
```

But the endpoint is "any authenticated actor" (the generated `endpoints.json` lists `user, agent, service_account, toolkit`), and `AgentService.claim` sets `owner_id = identity.sub` with no `actor_type` check. An authenticated **agent/service-account/toolkit** presenting the token would write an `agnt_…`/`svc_…` id into a users-FK column → `IntegrityError` → unhandled 500.

It's also a design hole, not just a crash: the claim token is returned on the `/register` response, i.e. to the agent process itself — the whole point is that the *human* claims. Realistically hard to exploit (a pending agent can't mint a bearer token, and approval sets `owner_id` first), but it's a one-line guard versus a 500.

**Fix:** reject `identity.actor_type != USER` early (403/400), and add a test.

### 2. Approve/claim race can still steal ownership (minor, follow-up OK)

`claim` correctly takes `with_for_update`, but `set_approval` reads via plain `session.get` (`admin/repos/agent_repo.py`):

```python
agent = await session.get(Agent, agent_id)
...
if agent.owner_id is None:
    agent.owner_id = approved_by
```

Under READ COMMITTED, an approve that reads `owner_id IS NULL` concurrently with a committing claim can overwrite the freshly-claimed owner — exactly the "operator silently becomes owner" outcome this PR exists to prevent, in its narrowest window. Cheap fix: `with_for_update()` in `set_approval` too (or re-check under lock).

### 3. Check ordering leaks owned/unowned state (accept, but document)

`claim` checks existence (404) and ownership (409) *before* validating the token, so any authenticated caller can probe a known agent id's ownership state without a token. The docstring's "never leak which agents are claimable" only holds for the no-token-issued case. The ordering is what makes the replay test's 409 semantics work, and agent ids are unguessable KSUIDs — likely acceptable, but say so explicitly in the docstring rather than implying the stronger property.

### 4. Retry cache can hand a token minted for a stale agent id (low)

The `minted` dict caches the plaintext across transaction retries, but `create_dcr` generates a **new** `agnt_` KSUID on each retry — the cached token was minted with the first attempt's id. Hash consistency still holds (the hash is recomputed onto the new row), so OSS is unaffected; but the seam's contract passes `agent_id` to the minter, so an overlay minter that *encodes* the id (e.g. a signed token) would embed a stale one. Either key the cache per-`agent.id` (re-mint on retry — plaintext/hash stay consistent since both re-derive), or document on the `ClaimTokenMinter` protocol that the id is advisory and must not be embedded.

### 5. Nits

- `ix_agents_claim_token_hash` is never queried — lookup is by PK, then hash-compare. Harmless (and symmetric with the RAT column), but it's a write-amplifying index with no reader; consider dropping from the migration.
- `allow_expired_password=True` on a mutation endpoint — matches the `GET /agents/{id}` precedent and there's a plausible onboarding rationale, but every other agent mutation requires full standing. Worth one sentence in the handler docstring saying it's intentional.
- The audit record sets `after={"owner_id": …}` but no `before` — `before={"owner_id": None}` would make the "unowned → owned" transition self-evident in the log.

## What's notably good

- The RAT/claim column separation with the rationale comment in the model and migration — future readers will need exactly that.
- Lazy-mint-with-cache reasoning documented at the call site; hash-then-`hmac.compare_digest` (length-equal inputs, timing-safe); expiry checked only *after* a matching token so expiry isn't probeable.
- Web tests use a real member user (valid FK target) with zero permissions — proving "token is the proof" rather than accidentally testing an admin path.
- Docs/OpenAPI/generated-client regeneration all included, and the endpoint classification ("any authenticated / no scope") is accurate.

ren-jentic (commented):
## Deep review (round 2) — post-fix

The round-1 findings are all addressed well (non-user 403 guard + tests, `set_approval` row lock, per-id retry cache, dropped index, docstrings, audit `before`). Re-reviewing the whole diff critically, the design is sound: high-entropy random token, hash-only storage, single-use consumption under a row lock, constant-time compare, expiry checked only after a matching hash, `claim_token` returned exactly once (not on the poll response), and the owner-scoped read filter makes the claim→GET→approve flow work for a plain member. Below are the remaining items — one worth fixing before merge, the rest hardening/nits.

### 1. `endpoints.json` now advertises actor types the endpoint rejects (fix before merge)

The regenerated classification for `:claim` still says:

```json
"actor_types": ["user", "agent", "service_account", "toolkit"],
"typical_caller": "any", "group": "Any authenticated actor"
```

But the round-1 fix makes the service **403 every non-USER actor**. So the machine-readable contract (and the rendered endpoint reference) is now misleading — it tells an agent/SA/toolkit it may call this, when it always gets `claim_actor_not_allowed`.

**Cleanest fix:** declare the constraint at the boundary instead of only in the service —
`get_current_identity(allow_expired_password=True, require_actor_type=ActorType.USER)`.
That (a) regenerates `actor_types: ["user"]` + the right typical_caller, (b) rejects non-users at the web layer (earlier, one place), and (c) lets the service guard stay as defense-in-depth. Then re-run `make openapi && cd ui && npm run codegen`.

### 2. No integrity-error backstop on the claim write (hardening)

The non-USER 403 depends entirely on `identity.actor_type` being set correctly upstream. `Identity.actor_type` **defaults to `USER`** (identity.py:57) — the comment notes `verify_token` fails closed on a token lacking the claim (#862), so this holds *today*. But `AgentService.claim` wraps the `owner_id` write in no `try/except`, so if that invariant is ever violated by a future resolver, we're back to a raw `IntegrityError` → 500 (the exact failure #1 set out to remove). `update_agent` already models the safe pattern (catches `DatabaseIntegrityError` → `InvalidOwnerError` 422). Cheap belt-and-suspenders: wrap `set_owner_from_claim` the same way.

### 3. Token-entropy contract is implicit (doc nit)

Storage is unsalted SHA-256 — fine for the shipped minter (`secrets.token_urlsafe(32)`, 256-bit), where brute-forcing a leaked hash is infeasible. But the `ClaimTokenMinter` protocol only says "opaque", not "high-entropy". A deployment whose minter returns a low-entropy/guessable token would be brute-forceable from a DB leak (unsalted). One sentence on the protocol mandating a cryptographically-random, ≥128-bit token would close the contract.

### 4. Threat-model note (not a change request)

The claim token is returned to the **agent process** in the `/register` body, yet confers ownership to whichever *authenticated user* presents it. If it leaks (CLI history, CI logs), any org user could claim the agent as themselves. Blast radius is bounded — single-use, 900s TTL, sets only `owner_id` (no scopes), admin-reassignable — so acceptable, but worth an explicit line in the design doc.

### Nits
- Expired token → `ClaimTokenInvalidError` → 400. `410 Gone` is arguably more precise for expiry, but 400 is fine and non-leaky; not worth churning.
- Web tests cover happy/replay/wrong-token/non-user/unauthenticated; the expired-token and archived-agent paths are unit-only. Fine given the unit coverage, but a web-layer expired-token case would exercise the full 400 mapping.

Nothing here blocks except #1 (which is a doc/contract-accuracy fix, low effort). #2 is a 4-line hardening I'd take too.

Issue #994, opened 2026-08-07
[install][ax] A sandboxed agent cannot complete the install: three independent security controls each block a step, and our own T1 tier describes that environment


ash-jentic:
## Context

Field report from a developer evaluating Jentic One at launch, attempting the install **from an agent** (the flow we explicitly encourage: "point your coding agent at the repo and ask it to install and register itself").

The agent could read and explain the entire system without executing anything. It could not install it. **Three independent security controls each blocked a different step, and all three were behaving correctly.**

## What happened

1. **The agent's sandbox refused the `curl … | sh` bootstrap as an injection pattern.** "A message suggests installing software, here is a pipe-to-shell command" is structurally indistinguishable from an attack, and the agent has no way to tell a genuine instruction from a spoofed one. Refusing is the correct behaviour.
2. **With the sandbox explicitly overridden by the user, the installer ran and then died on restricted egress.** `go.dev` and `proxy.golang.org` are not on the enterprise allowlist, so the download-Go-and-build-from-source step cannot complete. Filed separately: see the egress issue.
3. **The documented fallback also failed.** Standing the app up to use the raw-HTTP agent path was declined by the agent's safety layer, even with direct user authorisation.

## Why this is worth prioritising rather than filing under "unusual environment"

**Our own hardening guide's T1 tier describes almost exactly the environment this agent was in.** We have designed the *runtime* for sandboxed, restricted-egress agents, and designed the *installer* for an unrestricted developer laptop. The product's target environment and its install path disagree.

The reporter's framing, which is hard to argue with: the agents that matter commercially will be the ones living behind exactly these controls.

## Suggested remedies (reporter's, and they look tractable)

- **Signed prebuilt release binaries** with **published checksums**, so an agent can verify what it is running instead of piping an opaque script to a shell. (This is the same remedy #974 asks for on GitHub-outage grounds — two independent drivers for one fix.)
- **Treat the `llms.txt` / raw-HTTP path as a first-class, install-free option**, documented as such, so a restricted agent has a supported route to a first successful call. Related: #624 (no zero-config path to a first successful call).
- Any wording change that lets an agent justify the bootstrap to its own safety layer — a verifiable artifact and checksum is exactly that justification.

## Related

- #755 — no clear path to install/reach self-hosted Jentic One from inside an agent session (same theme, human-UX angle)
- #974 — prebuilt release artifacts (same remedy, outage-resilience driver)
- #624 — no zero-config path to a first successful call
[...]

PR #1051, opened 2026-08-14
docs(readme): rewrite the front door for discovery and first-run success

ash-jentic:
Draft for review. Rewrites the README's top half, keeps all existing content, and adds a truthful install plus a path to a first successful call. Full reasoning is in the commit message.

**To preview it rendered**, browse the branch rather than reading the diff:
https://github.com/ash-jentic/jentic-one/tree/docs/readme-discovery — GitHub renders the root README there exactly as it will look on merge, with the figures and relative links resolving in place.

### The changes that need a maintainer's eye

| Change | Why |
| --- | --- |
| Install rewritten | `tools/install.sh` downloads a Go toolchain, clones, builds, and hands off to the `jenticctl install` wizard. The README described it as "Install `jenticctl`". |
| Signed binary promoted to Option A | v0.31.0 already ships tarballs, SBOMs and a signed `checksums.txt` that the documented path ignored. Per #994 a sandboxed agent cannot use a pipe-to-shell bootstrap. |
| New "First brokered call" | `jentic register` blocks on operator approval, and the bootstrap login is `admin@local` / `1234` — documented only at `cli/README.md:183`. Neither was discoverable. |
| Duplicate Quick start deleted | Two H2s differing only by capitalisation, same curl command, broken heading anchors. |
| `/app/docs` corrected | It 404s on a running deployment. Generated references are at `/docs` and `/redoc`. |
| Two claims corrected | "Secrets never leave your infrastructure" is not airtight — the Broker forwards the credential upstream. Replaced with the forms that hold. |
| Gateway section added | Answers whether this replaces an API gateway, and states that private and internal APIs are a first-class path. |
| Mermaid topology replaced | `docs/assets/architecture.png`, built from the product page's design tokens and typeface. |

### Deliberately preserved

The same-OS-user caveat, unsoftened. The telemetry closed-schema paragraph and its pointer to `tests/arch/test_telemetry_no_pii.py`. Default-deny. The vendored arch-rule-facts mechanism. Every docs link, every `make` target, the full Components table.

This is checked mechanically, not by eye: a preservation gate inventories every link, command, heading, image and code block in the current README and fails unless each is kept, moved, or explicitly declared dropped. Four headings are declared: the two duplicate Quick starts, `What is Jentic One?` (renamed), and `Architecture rule facts` (folded into Development & testing).

**Badges are untouched.** Trimming them is a separate proposal.

### Please confirm

- `admin@local` / `1234` is still the bootstrap credential, and naming it in the README is acceptable.
- The API count is written as "thousands of OpenAPI specs". The verified count on `jentic-public-apis@main` is 6,056 distinct APIs, if you would rather be specific.


ren-jentic:
Reviewed this end-to-end (and verified the load-bearing claims against `main` and live GitHub rather than taking them on faith). This is genuinely strong work — the standout is the **evidentiary discipline**: nearly every claim is checkable, and I confirmed the big ones (bootstrap login at `cli/README.md:183`, the `jentic register` approval-blocking flow, Go 1.26.2 in `install.sh`, all the `make` targets, and the signed `checksums.txt`/`.sig`/`.pem` + SBOMs on `v0.31.0`). The `/app/docs` → `/docs`+`/redoc` correction is a real fix, the preservation-gate approach is exactly right, and the "What it is not" framing does a lot of honest positioning in very little space.

One thing to fix before merge, then a few polish items.

## Must fix: `docs/quickstart.md` is a broken link (used twice)

The README points at `docs/quickstart.md` in two places:

- header nav: `First brokered call → docs/quickstart.md`
- the *First brokered call* section: *"Full walkthrough: [docs/quickstart.md](docs/quickstart.md)."*

…and positions it as **the** canonical walkthrough the six-step summary defers to. But `docs/quickstart.md` isn't on `main`, isn't on this branch, and isn't added by this PR. So on merge, the reader who follows the "first success" promise hits a 404 — in the most important new section of the front door.

Options, in order of preference:
1. Add `docs/quickstart.md` in this PR (the six-step block is already ~90% of its content — promote it into a real page).
2. Point both links at an existing target (`cli/README.md` already documents the register → approve → execute loop) until the page exists.
3. Drop the "full walkthrough" links and let the six inline steps stand alone.

Note: the preservation gate checks that *existing* links are kept — it won't catch a *new* link to a non-existent file. Worth adding a "new relative links must resolve" check to the same gate.

## Polish (non-blocking)

- **Go badge vs. reality.** The badge says **Go 1.25**; `install.sh` downloads **Go 1.26.2**. I know badges are deliberately out of scope, but this PR is specifically rewriting the install story, so the adjacent contradiction stands out. Either bump the one badge or note it as known.
- **`how-it-works.gif` is ~601 KB.** It's the heaviest thing on the landing page and loads above the fold for every visitor, which cuts against the discovery goal. Worth confirming it's optimised (palette/frames) or whether the 229 KB PNG carries enough on its own.
- **API count.** Agree with taking the specific number you offered — "6,056" (or "6,000+") reads as a real, countable catalogue; "thousands" reads as a hedge to exactly the technical audience being courted. (Keep it consistent with `llms.txt`, which also says "thousands".)
- **Prerequisites scoping.** The Quickstart lists `git`/`uv`/Docker up front, but Option A (signed binary → `jenticctl install`) doesn't need the build tools — and a security-conscious reader likely chose Option A *because* it avoids that path. Consider scoping prerequisites per-option.

On the confirmations you asked for: `admin@local` / `1234` is already public at `cli/README.md:183`, so naming it here isn't new exposure and the "change on first login" instruction is present — fine by me.

ren-jentic:
Pushed two fixes to this branch (maintainer edit):

- **Added `docs/quickstart.md`** — the file the header nav and "First brokered call" section linked to but that didn't exist. It expands the six-step summary into a full walkthrough (sign in + rotate the bootstrap password, import via `jentic catalog`, store a credential, `jentic register` + operator approval, `jentic access request`, `jentic execute`), with the same-OS-user caveat up top and links out to credentials/hardening/CLI/endpoints. Kept the wording to verified commands only — I softened the catalog step to the interactive browser rather than inventing a non-interactive `import` flag.
- **Go badge 1.25 → 1.26** to match `tools/install.sh` (downloads Go 1.26.2).

Left the remaining polish (601 KB `how-it-works.gif`, "thousands" vs. a specific count, per-option prerequisites) to you — those are your calls, not defects. All the relative links in the new page resolve against the branch.

Issue #809, opened 2026-07-27
[request] Serve llms.txt (and /.well-known/llms.txt) so agents can self-onboard from the base URL


Manuel-Jentic:
### What's the problem, friction, or need?

An agent pointed at a Jentic One deployment has no self-describing entry point. The building blocks exist — RFC 8414 metadata at `/.well-known/oauth-authorization-server` (advertising the DCR `registration_endpoint`), `/openapi.json`, the public `/reference/endpoints.json`, and the onboarding skill — but nothing stitches them together at the base URL. Today the only "agent guide" is the markdown the CLI writes to the local disk, which never reaches remote agents (#651).

### What would you like instead?

Host an `llms.txt` at `GET /llms.txt` (and `GET /.well-known/llms.txt`) that gives an LLM everything it needs to self-onboard from the base URL alone:

- Link to the hosted onboarding skill (`GET /skills/jentic.md`, #651)
- Links to `/openapi.json`, `/reference/endpoints.json`, and the RFC 8414 discovery document
- The onboarding quickstart: register (`POST /register` with Ed25519 keypair + JWKS payload shape) → token (`jwt-bearer` grant) → discover → access request → brokered call (#569)

This was already designed in the scopes-permissions implementation guide §7 ("Agent Registration & Discovery — llms.txt & Targeted DCR") but never got a tracking issue.

### Related

- #651 — serve the onboarding skill from a URL (llms.txt should reference it)
- #569 — narrative onboarding guide (the quickstart content)
- #628 — per-agent skill distribution (gated on a hosted canonical copy)

PR #1052, opened 2026-08-14
docs: add llms.txt, and an install-and-use section to AGENTS.md

ash-jentic:
Draft for review. Two additive changes, no existing content altered.

Serves the flow this project already encourages and #994 documents: *"point your coding agent at the repo and ask it to install and register itself."*

### AGENTS.md

`AGENTS.md` is the file a coding agent reads when pointed at a repository. Its current 38 lines are entirely about **changing** this codebase — layering, ESLint, git conventions — so an agent asked to **install and use** Jentic One finds instructions for modifying it instead.

This adds a `## Using Jentic One (not changing it)` section above the existing guidance. The existing guidance is untouched (38 → 85 lines, all additions).

Design choices worth a maintainer's view:

- **Signed release binary listed first.** Per #994, a sandboxed agent refuses `curl | sh` as an injection pattern, and all three controls that blocked that reporter were behaving correctly. A verifiable artifact is the only install path some agents have.
- **Names the two facts that strand people:** `jentic register` blocks on operator approval (`cli/README.md:284`), and the bootstrap login is `admin@local` / `1234` (`cli/README.md:183`).
- **Tells the agent to stop rather than route around a control.** If its own sandbox or egress rules block a step, it should report that and stop. An agent that circumvents a security control to finish an install is a worse outcome than a failed install.

Every claim is drawn from existing docs and cited above.

### llms.txt

Follows the [llmstxt.org](https://llmstxt.org) convention: a short document an agent or assistant reads to evaluate and install. It **defers to** the richer document a running instance already serves at `/llms.txt` and `/.well-known/llms.txt` (#809) and covers only evaluation and installation, so the two do not compete.

It also states that private and internal APIs are a first-class path, and that both PostgreSQL and SQLite are supported production backends.

### Please confirm

- `admin@local` / `1234` is still the bootstrap credential, and that naming it in `AGENTS.md` is acceptable rather than leaving it in the CLI reference only.
- Nothing in the `Using` section conflicts with how you would want an agent to behave against a real deployment.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -3,6 +3,53 @@
 > **If `.rules/` exists, read `.rules/AGENTS.md` and use it instead of this file.**
 > Otherwise use the guidance below.
 
+## Using Jentic One (not changing it)
+
+If your task is to install Jentic One and connect an agent to it, follow this section. The
+rest of this file is for changing the codebase.
+
+**Install.** Two supported paths:
+
+1. Signed release binary, from https://github.com/jentic/jentic-one/releases/latest. Verify
+   against `checksums.txt` (signature: `checksums.txt.sig`, certificate: `checksums.txt.pem`),
+   then run `jenticctl install`. Prefer this path: it is verifiable before execution, and a
+   sandboxed agent will usually be refused permission to pipe a script to a shell.
+2. Bootstrap script:
+   `curl -fsSL https://raw.githubusercontent.com/jentic/jentic-one/main/tools/install.sh | sh`
+   This downloads a Go toolchain if none is present, clones this repository, builds `jenticctl`
+   and `jentic`, then runs `jenticctl install`. Set `JENTIC_NO_INSTALL=1` to stop after the
+   binaries are installed.
+
+**Run Jentic One on a different machine from the agent.** An agent running as the same OS user
+can read the credential database and encryption key from disk, whatever the API-level controls
+allow. See `docs/security/hardening.md` before using real credentials.
+
+**Register and reach a first call:**
+
+1. `jentic register` creates the agent identity. It then blocks, waiting for an operator to
+   approve the agent. Report this to the user: on a single-operator install they are the
+   operator, and they approve it in the UI at `/app`.
+2. The bootstrap operator login is `admin@local` / `1234`. Tell the user to change it.
+3. Import an API from https://github.com/jentic/jentic-public-apis, or register a private
+   OpenAPI description of the user's own service.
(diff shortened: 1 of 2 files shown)


ren-jentic:
Reviewed alongside #1051. This one I'd take close to as-is — only nits below, no blockers.

The part I want to call out as genuinely good: the AGENTS.md instruction to **stop rather than route around a control** — *"If a step is blocked by your own sandbox or by network egress rules, say so and stop. Do not work around a security control."* That's the right posture (an agent that circumvents its sandbox to finish an install is a worse outcome than a failed install), and it's the kind of thing that arguably deserves to be a shared rule, not just AGENTS.md prose. Splitting `Using Jentic One` from the existing `changing it` guidance is the correct fix for the real problem that a coding agent pointed at the repo currently finds only "how to modify this codebase." And the `llms.txt` deferring to the richer `/llms.txt` a live instance serves (rather than competing with it) is exactly right.

I verified the claims that carry weight here: bootstrap login at `cli/README.md:183`, the `register` approval-blocking flow, `install.sh` behaviour and `JENTIC_NO_INSTALL=1`, SQLite + PostgreSQL both as backends, and every linked doc resolves.

Nits (batchable):

- **Citation off by one.** AGENTS.md cites `cli/README.md:284` for the approval fact; the actual text is at `:285`/`:286` (`register` polls the token endpoint and continues once approved). PR #1051 makes the same citation — worth fixing in both.
- **Raw-URL form is inconsistent in `llms.txt`.** Most links use `.../refs/heads/main/...`, but the install one-liner uses `.../main/tools/install.sh`. Harmless, but pick one form.
- **"thousands" of APIs.** Same note as on #1051 — if the README moves to the specific count (6,056 / 6,000+), keep `llms.txt` consistent (it says "thousands" in two places).

On your confirmation asks: naming `admin@local` / `1234` here is fine (already public at `cli/README.md:183`), and nothing in the `Using` section conflicts with how I'd want an agent behaving against a real deployment — the "stop, don't route around" rule is exactly the behaviour I'd want.

ren-jentic:
Pushed one small fix to this branch (maintainer edit): the install one-liner in `llms.txt` now uses the `/refs/heads/main/` raw-URL form to match every other link in the file. Commitlint passed.

The off-by-one citation I mentioned (`cli/README.md:284` → `:285`/`:286`) is only in the PR description, not in the committed files, so there's nothing to change in the tree for it. Left the "thousands" vs. specific-count wording to you.

PR #1064, opened 2026-08-17
fix(localagent): fail closed when agent account creation silently fails

ren-jentic:
## Summary

Creating a second agent account on macOS always failed silently and left the CLI in a dead-end loop (`jentic run` says the account doesn't exist → sends you to `jentic bootstrap` → bootstrap "succeeds" again). Root cause chain, reproduced on a machine with an existing `<operator>-local-agent` account:

1. `sysadminctl -addUser` was given a **constant per-operator full name** (`"<operator> Local Agent"`); macOS Open Directory refuses duplicate full names, so any second account name collided with the first's record.
2. `sysadminctl` reports that refusal on stderr but **exits 0**, and the create path trusted the exit code — so bootstrap sailed on: `createhomedir` no-op'd, ACL/chown/seed steps degraded into best-effort "unknown user" noise, a NOPASSWD sudoers rule was installed for a **nonexistent** user, and the phantom account was recorded as `account_created: true`.
3. Separately, the setup form prefills the home from the **default** name; renaming the account but leaving the home meant a brand-new account targeted the **existing** agent account's live home — the create path (unlike reuse, which goes through `VerifyManagedHome`) would stamp operator ACLs over it and chown it wholesale.

Changes:

- `AccountFullName`: macOS full name now embeds the account name (unique by construction), so second accounts can't collide
- `createAgentAccount` verifies `UserExists` after the create steps and fails closed **before** installing the sudoers rule or recording the account, surfacing a clear error instead of the "unknown user" cascade
- `HomeClaimedBy`: create path refuses a home that another existing account records as its home (`dscl` on macOS, `getent passwd` on Linux)
- `followRenamedHome`: renaming the account in the form re-derives the home from the new name when it was left at the stale default prefill (with a printed notice); a customised home is always kept
- docs: manual recipe in `local-agent-isolation.md` updated to the per-account full name + an `id -u` verification step

Out of scope (follow-up): the `agent_account` record still lives in the legacy `~/.jentic/config.yaml`, which `jentic migrate --purge-legacy` deletes wholesale.

## Test plan

- [x] `make lint` (golangci-lint) — 0 issues
- [x] `go test ./internal/...` — all packages pass (new: `TestAccountFullName`, `TestHomeClaimedBy`, `TestFollowRenamedHome`)
- [ ] On a macOS box with an existing `<operator>-local-agent`: `jentic bootstrap` with a new account name now creates the account (unique full name) instead of silently failing
- [ ] Bootstrap with a new name but the stale default home prefill re-derives the home and never touches the existing account's home

### DIFF
--- a/cli/internal/cli/localagentcmd/agentuser.go
+++ b/cli/internal/cli/localagentcmd/agentuser.go
@@ -146,6 +146,17 @@ func (a *Cmd) setupAgentUser(ctx context.Context, operators []string, interactiv
 		return agentSetup{}, err
 	}
 
+	// The home field was prefilled from the DEFAULT account name; if the operator
+	// renamed the account but left the home at that stale default, follow the name
+	// rather than silently keeping a directory derived from a name they rejected —
+	// which, when the default-named account already exists, is that OTHER account's
+	// live home (see HomeClaimedBy for the guard behind this convenience).
+	if home, changed := followRenamedHome(fields.name, defaultName, fields.homeDir); changed {
+		fields.homeDir = home
+		fmt.Fprintln(a.Out, theme.Dim.Render(
+			"Agent home follows the account name: "+fields.homeDir))
+	}
+
 	if err := a.createAgentAccount(ctx, operator, fields); err != nil {
 		return agentSetup{}, err
 	}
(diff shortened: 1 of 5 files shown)


ren-jentic:
Folded into #1065 via cherry-pick (2feab7cf) — its "out of scope" note about the agent_account record living in legacy ~/.jentic/config.yaml is exactly what #1065 fixes, so the two land as one unit there.

PR #1083, opened 2026-08-18
fix(registry): accept canonical vendor/name/version slugs in search api filters

### LINKED ISSUE #1080: [bug] `search --api` rejects every documented filter format: CLI sends vendor/name/version, server parses colons
### What's broken?

`jentic search --api <filter>` returns **422 `Unknown API filter`** for every format the CLI documents. The flag help, the command example, and the ecosystem-wide canonical slug all use **slash-separated** `vendor/name/version`, but the server parses the filter as **colon-separated**.

- CLI: `cli/internal/cmd/search.go:53` — `restrict to these APIs (vendor/name/version; repeatable)`; example at line 38 uses `--api stripe.com/api/v1`. The strings are passed verbatim into the POST `/search` body (`cli/internal/searchclient/client.go:32`).
- Server: `_resolve_api_filters` in `src/jentic_one/registry/services/search_service.py:86-100` does `entry.split(":")`.

So any documented invocation resolves the whole string as a vendor, finds nothing, and 422s. The undocumented colon form (`--api vendor:name:version`) works. Scoped search is the main tool an agent has for navigating a large imported API (GitHub is ~1,100 operations), so this bites exactly when a deployment matures past one API.

Two aggravators:
1. The 422 detail (`Unknown API filter: '…'`) does not say what format is expected.
2. The `apis` field on `SearchRequest` (`src/jentic_one/registry/web/schemas/search.py:18`) has no description, so the OpenAPI spec doesn't document the wire format either. (`revision_pins` keys share the same undocumented colon convention.)

Suggested fix: accept the canonical slash slug server-side (keep colons for back-compat), describe the field in the schema, and name the expected format in the error detail. Unit tests exercise the service with colon strings directly, which is why no test caught the CLI-facing contract.

### Steps to reproduce

1. `jentic catalog import api.github.com`
2. `jentic search "list pull requests" --api github-com/api-github-com/1.1.4`
3. Observe 422. Repeat with `--api github-com/api-github-com` and `--api api.github.com` — same. `--api github-com:api-github-com:1.1.4` (undocumented) works.

### Actual output / error (paste verbatim)

```
error: http 422: Unknown API filter: 'github-com/api-github-com/1.1.4'
```

### Environment

- Jentic One version / commit: server 0.31.0 (Docker image), CLIs v0.28.0
- OS: macOS 15 (darwin 24.6.0), Apple silicon
- Install method: Docker (`jenticctl install`)
- Database backend: Postgres

*Found while dogfooding the GitHub-via-broker loop from a coding agent; filed through the broker itself.*

Manuel-Jentic:
## Summary

`jentic search --api <filter>` failed with 422 `Unknown API filter` for every format the CLI documents: the flag help, its example, and every other surface use the canonical slash slug (`vendor/name/version`), but `_resolve_api_filters` only parsed a colon-separated form. Scoped search is the main tool an agent has for navigating a large imported API (GitHub is ~1,100 operations), so this bit exactly when a deployment grows past one API. The server now accepts the canonical slug and normalizes vendor/name the same way ingest does (so raw spellings like `stripe.com/api` resolve too), keeps the colon form for back-compat, returns a 422 that names the expected format, and documents the wire format in the schema.

## Changes

- `registry`: new `_parse_api_identifier` helper accepts `vendor[/name[/version]]` alongside the legacy colon form, used by both the `apis` filter and `revision_pins` keys. The **first separator present decides the form**, so a slash slug whose verbatim version embeds a colon (versions are never slugified) still parses correctly, and vice versa.
- `registry`: filter vendor/name are normalized with `slugify_api_field` — the same canonicalization ingest applies before storing them — so raw spellings resolve instead of 422ing on exact string comparison. Versions are only trimmed. Empty segments (`vendor/`) degrade to an unfiltered axis rather than exact-matching `''`.
- `registry`: `InvalidApiFilterError` detail names the expected format with an example; partial `revision_pins` keys get a pins-specific hint ("keys need the full 'vendor/name/version'").
- `registry`: `SearchRequest.apis` / `revision_pins` gain `Field(description=…)` so the generated spec documents the identifier format; regenerated `openapi/control/control.openapi.yaml`, `ui/openapi.json`, and the generated UI client (`SearchRequest.ts`, comment-only).
- `cli`: the `--api` flag help and the `search` command example showed an identifier the server rejected (`stripe.com/api/v1` against a registry storing slugs); both now point at real slug shapes and `jentic apis list`.
- `tests`: parser unit tests (both separator forms, first-separator precedence, colon tails, slugification, empty segments, whitespace); end-to-end integration tests through the real `Ingestor`/`SearchService` covering filter shapes, raw spellings, cross-API restriction, `revision_pins` in both forms, partial-pin rejection, and the error hints; new web-layer `tests/web/registry/test_search.py` locking the router's 200 / 422 `invalid_api_filter` contract (the seam the CLI keys its errors on), including auth.
- Out of scope: consolidating identifier parsing into `shared/models/api_identity.py` and migrating the colon-only parsers in `note_service.py` / admin `executions.py` (same bug class), a CLI recovery hint on `invalid_api_filter` 422s, and deprecating the colon form — each a candidate follow-up issue, kept out to keep this a focused bug fix.

## Risk & rollback

- Behavioural deltas are confined to inputs that previously 422'd: colon filters parse identically (vendor/name slugification is idempotent on the already-slug values that were the only ones that ever resolved), and slugs cannot contain either separator, so no previously working filter changes meaning. Two deliberate strictness changes: `a:b:c:d` no longer silently truncates the version at the third colon, and pin keys with colon-bearing versions are now accepted (matching the broker pin grammar).
- No migrations, no auth/credential/permission surface, no wire-shape change (schema edits are description-only).
- Revert the squash commit.

## Test plan

- `uv run pytest tests/unit tests/arch -q` — 2799 passed
- `JENTIC_TEST_BACKEND=sqlite uv run pytest tests/integration/registry -q -m integration` — 172 passed, 1 skipped
- `uv run pytest tests/integration/registry/ingest/test_search_api_filter.py tests/web/registry/test_search.py -m integration` (postgres) — 16 passed
- Replication proven: the new integration tests were run against the unfixed parser and the slash-slug cases failed exactly as reported (colon cases passed)
- `uv run ruff check` / `ruff format --check` / `mypy` clean; `make openapi` drift-free; `go build ./... && go vet ./...` clean in `cli/`

## Review guide

- Start with `_parse_api_identifier` in `src/jentic_one/registry/services/search_service.py` — the whole fix is its separator + normalization rules; the docstring states them.
- Then the two call sites (`_resolve_api_filters`, `_resolve_revision_pins`) and the error hints in `services/errors.py`.
- Tests mirror that order; OpenAPI/UI-client changes are regenerated, description-only.

Closes #1080


Manuel-Jentic:
Pushed review follow-ups (from a four-lens review pass: feasibility/design, code quality, product fit, test/risk):

- **First-separator-wins parsing** — colon-precedence would misparse a canonical slash slug whose verbatim version embeds a colon (versions are never slugified). Either form now parses correctly regardless of what the version contains.
- **Vendor/name slugified like ingest** — the filter used to compare raw strings against slugged DB values, so raw spellings (`stripe.com/api`) still 422'd; they now normalize the same way ingest and the access-request path do. Empty segments (`vendor/`) degrade to an unfiltered axis instead of exact-matching `''`.
- **Pins-specific 422 hint** — a partial `revision_pins` key now says the full triple is required rather than suggesting partial forms are fine.
- **Coverage for the previously untested seams** — `revision_pins` parsing (both separator forms end-to-end + partial-key rejection), colon-tail (`a:b:c:d`) and slugification unit cases, and a new web-layer `POST /search` test (`tests/web/registry/test_search.py`) locking the 200/422 `invalid_api_filter` contract the CLI keys on.
- **CLI help fix** — the shipped `--api stripe.com/api/v1` example was an identifier the server rejected; example and flag help now point at real slug shapes and `jentic apis list`.

Deliberately deferred (candidate follow-up issues): consolidating identifier parsing into `shared/models/api_identity.py` and migrating the colon-only parsers in `note_service.py` / admin `executions.py` (same bug class); a CLI hint on `invalid_api_filter` 422s; partial-ref support in `apis` subcommands.

ren-jentic:
Resolved the merge conflict with `main` (merge commit 1d4d6775).

The conflict was a modify/delete on `cli/internal/cmd/search.go`: the CLI V2 rebuild (#1049) deleted the whole `internal/cmd` tree, while this PR had touched two help-text lines in it. Resolution:

- Accepted main's deletion of `cli/internal/cmd/search.go` and **ported the PR's two-line `--api` help/example fix to the V2 location** `cli/internal/cli/api/search.go` (which still carried the old `stripe.com/api/v1` example this PR exists to fix).
- Regenerated the V2 artifacts the port touches: `make cli-reference` (`ui/public/cli-reference.json`) and `cli make generate-api` — the vendored `client/generated/control/spec.yaml` and generated Go client now carry this PR's `SearchRequest` field descriptions (comment-only).
- Verified the auto-merged `openapi/control/control.openapi.yaml` / `ui/openapi.json` are byte-identical to a fresh `make openapi`.

Post-merge verification: `go build/vet/test ./...` in `cli/` (incl. arch, golden, drift gates) all pass; `uv run pytest tests/unit tests/arch -q` — 2892 passed; sqlite `tests/integration/registry` — 186 passed / 1 pg-only skip; targeted `test_search_api_filter.py` + `tests/web/registry/test_search.py` — 16 passed; ruff / mypy clean.

ren-jentic (approved):

PR #1137, opened 2026-08-25
fix(auth): re-check actor status on every token verdict so disable kills outstanding tokens

### LINKED ISSUE #1136: Token resolvers never re-check actor status — a disabled agent's outstanding access token stays valid until expiry (up to 1h)
## Problem

Disabling an agent (`ActorStatus.DISABLED`) blocks all **future** token mints (`auth/services/assertion_service.py:138-139` requires `ACTIVE`), but **neither per-request token resolver re-checks actor status**, so an already-minted `at_` access token keeps verifying until its `expires_at` — up to `access_ttl_seconds` (default **3600 s**, `shared/config.py:267`):

- **Control plane**: `resolve_access_token` (`auth/services/token_service.py:294-347`) derives `active` only from `revoked_at is None and expires_at > now` (line 339).
- **Broker**: `InProcessTokenResolver.resolve_access_token` (`broker/repos/token_resolver.py:28-80`) — deliberately decoupled from `auth/` — likewise checks only revocation + expiry (line 73) and never joins the actor tables. **This is the execute path**: a disabled agent can keep executing through the broker for up to an hour.

Today the complete kill procedure is *disable + revoke the token family* — two steps, and nothing documents that disable alone is insufficient.

## Proposed fix

Add an actor-status check to **both** resolvers:

1. `token_service.resolve_access_token` — free for agents (the agent row is already loaded for `parent_actor_id`, lines 325-328); service accounts need one extra indexed lookup in the same session.
2. `InProcessTokenResolver` — one extra JOIN against the actor tables in the raw SQL.

Residual propagation bound after the fix: the broker caches token verdicts for `resolve_cache_ttl_seconds` (**3.0 s** default, `shared/config.py:691`; cached entries re-check only expiry, `broker/core/token_validation.py`) — so disable-alone would kill within ≤ ~3 s, which is the right bound.

## Test

Pin the regression test on an **execute through the broker**, not just a control-plane call: disable agent → execute rejected once the verdict cache TTL lapses.

## Context

Found during the local-MCP planning review ([jentic-one-plans#25](https://github.com/jentic/jentic-one-plans/pull/25), §3.10 kill-switch analysis; independently re-verified in review). Transport-independent — affects CLI, REST, and any future MCP surface equally — hence filed standalone rather than gated behind that plan. Verified against `main` @ `65762127`.

ren-jentic:
## Summary

Disabling an agent only blocked *future* token mints: both per-request `at_` token resolvers derived `active` purely from revocation + expiry, so a disabled actor's outstanding access token kept working for up to `access_ttl_seconds` (1 h) — and, worse, the refresh grant never re-checked actor status either, so a disabled actor could keep rotating fresh access tokens for up to `refresh_ttl_seconds` (7 days). Every token verdict now re-checks the actor row, making disable alone a working kill switch: outstanding tokens die within the broker's verdict-cache TTL (≤ ~3 s by default), no separate token-family revocation required.

## Changes

- **auth** `TokenService.resolve_access_token` (control plane): folds actor status into `Identity.active` — agents/service accounts must be `ACTIVE`, users must have `active=true`, a missing actor row fails closed. Agents cost zero extra queries (row was already loaded for `parent_actor_id`).
- **broker** `InProcessTokenResolver` (execute path): the raw SQL now resolves an `actor_status` via per-actor-type subqueries (stays decoupled from `auth/`); `NULL` fails closed; users' boolean is normalised to the same status string. Dialect-independent (Postgres + SQLite).
- **auth** `TokenService.refresh`: gates the refresh grant on actor status (`InvalidGrantError`, no mint, refresh token not consumed) — closes the rotation loophole beyond the issue's 1 h window. Refresh-token reuse detection deliberately still wins, so a replayed token always revokes the family.
- **auth** `TokenService.introspect`: RFC 7662 `active` now agrees with what the resolvers enforce, so an operator inspecting a token after a disable sees the truth.
- Out of scope: `jak_`/`sak_` API keys (already status-gated in `ApiKeyResolver`); JWT-path validation (no DB backing by design); MCP/plan work this was found under.

## Risk & rollback

- Auth surface: token verdicts get stricter — tokens whose actor is non-`ACTIVE` (or whose actor row is missing, e.g. hard-deleted) stop resolving as active. Intended; re-enabling an actor revives its outstanding tokens.
- Refresh grant now returns `invalid_grant` for non-active actors (previously succeeded).
- One extra indexed lookup per verdict for service-account/user tokens on the control plane; one subquery in the broker's resolve (amortised by the 3 s verdict cache). Agents: no extra query on the control plane.
- No migrations, no config changes, no API-shape changes. Rollback: revert the squash commit.

## Test plan

- New regression pins: disable agent → broker validate rejected once the verdict-cache TTL lapses (the issue's pinned test), and re-enable revives the same token; control-plane mint → disable → inactive → enable → active; disabled actor's refresh rejected *without* consuming the refresh token; disabled SA / deactivated user / archived-pending-rejected agent / missing actor row all inactive; reuse detection still revokes the family for disabled actors.
- Ran: `uv run pytest tests/unit/ tests/arch/` (2928 passed), `make test-integration` (961 passed, Postgres), `make test-integration-sqlite` (644 passed), `ruff check`, `ruff format --check`, `mypy` — all green.

## Review guide

- Start with the two resolver seams: `src/jentic_one/auth/services/token_service.py` (`_actor_is_active` + its three call sites) and `src/jentic_one/broker/repos/token_resolver.py` (the `actor_status` CASE). The test churn is mostly mechanical: existing tests now seed real actor rows because orphaned tokens fail closed.

Closes #1136

### DIFF
--- a/src/jentic_one/auth/services/service_account_auth_service.py
+++ b/src/jentic_one/auth/services/service_account_auth_service.py
@@ -6,6 +6,7 @@ import hmac
 
 from jentic_one.admin.repos import (
     ActorScopeGrantRepository,
+    AgentRepository,
     ServiceAccountCredentialRepository,
     ServiceAccountRepository,
 )
@@ -160,6 +161,9 @@ class ServiceAccountAuthService:
         """Mint a short-lived ephemeral token for a task/agent.
 
         Security-critical: requested_scopes MUST be a subset of host_sa_scopes.
+        The target agent must exist and be ACTIVE — resolvers re-check actor
+        status on every verdict (#1136), so minting for a non-active agent
+        would return a token that is dead on arrival; fail fast instead.
         Returns an opaque access token (no refresh token for ephemeral mints).
         """
         host_scope_set = set(host_sa_scopes)
@@ -176,6 +180,14 @@ class ServiceAccountAuthService:
             raise InvalidGrantError(
                 f"invalid_request: ttl_seconds must be between"
                 f" {_MINT_TTL_MIN_SECONDS} and {_MINT_TTL_MAX_SECONDS}"
+            )
+
+        # One message for missing and non-active: no oracle for probing agent ids.
+        async with self._ctx.admin_db.session() as session:
+            agent = await AgentRepository.get_by_id(session, target_agent_id)
+        if agent is None or agent.status != ActorStatus.ACTIVE:
+            raise InvalidGrantError(
+                "invalid_request: target_agent_id does not resolve to an active agent"
             )
 
         return await self._token_svc.issue_access_only(
--- a/src/jentic_one/auth/services/token_service.py
+++ b/src/jentic_one/auth/services/token_service.py
@@ -200,9 +200,12 @@ class TokenService:
                     origin=None,
                 )
             else:
-                # A disabled/archived actor must not rotate its way to fresh
-                # access tokens — refresh is a mint path, gate it like the
-                # jwt-bearer exchange does (#1136).
+                # A non-active actor must not rotate its way to fresh access
+                # tokens — refresh is a mint path, so it gets a status gate
+                # too (#1136). Unlike the jwt-bearer exchange, PENDING gets no
+                # distinct detail here: a refresh token only exists after a
+                # successful exchange, so pending-approval polling never
+                # reaches this path.
                 if not await _actor_is_active(session, rt.actor_id, rt.actor_type):
                     raise InvalidGrantError("actor is not active")
 
--- a/src/jentic_one/auth/services/token_service.py
+++ b/src/jentic_one/auth/services/token_service.py
@@ -13,13 +13,15 @@ from jentic_one.admin.repos import (
     ActorScopeGrantRepository,
     AgentRepository,
     RefreshTokenRepository,
+    ServiceAccountRepository,
+    UserRepository,
 )
 from jentic_one.auth.services.errors import InvalidGrantError
 from jentic_one.shared.audit import AuditAction, AuditTargetType, record_audit
 from jentic_one.shared.auth.identity import Identity
 from jentic_one.shared.context import Context
 from jentic_one.shared.db.ids import generate_ksuid
-from jentic_one.shared.models import ActorType
+from jentic_one.shared.models import ActorStatus, ActorType
 
 ACCESS_TOKEN_PREFIX = "at_"
 REFRESH_TOKEN_PREFIX = "rt_"
@@ -31,6 +33,27 @@ def _hash_token(token: str) -> str:
 
 def _generate_token(prefix: str) -> str:
     return f"{prefix}{secrets.token_urlsafe(32)}"
+
+
+async def _actor_is_active(session: AsyncSession, actor_id: str, actor_type: str) -> bool:
+    """Whether the token's actor is currently allowed to authenticate.
+
+    Disabling an actor must kill its *outstanding* tokens, not just block new
+    mints (#1136) — so every token verdict re-checks the actor row. Fails
+    closed when the actor row is missing (e.g. a hard-deleted user whose
+    tokens were never revoked). Unknown actor types are left to their token's
+    own revocation/expiry checks.
+    """
+    if actor_type == ActorType.AGENT:
+        agent = await AgentRepository.get_by_id(session, actor_id)
+        return agent is not None and agent.status == ActorStatus.ACTIVE
+    if actor_type == ActorType.SERVICE_ACCOUNT:
+        sa = await ServiceAccountRepository.get_by_id(session, actor_id)
+        return sa is not None and sa.status == ActorStatus.ACTIVE
+    if actor_type == ActorType.USER:
+        user = await UserRepository.get_by_id(session, actor_id)
+        return user is not None and user.active
+    return True
 
 
 class TokenService:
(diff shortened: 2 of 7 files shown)


Manuel-Jentic (commented):
Deep review done — six parallel passes: control-plane resolver, broker SQL (dialect-verified by executing it on SQLite), test-coverage audit, end-to-end kill-switch trace across every auth path, workspace-rules compliance, and an independent test run.

**Verdict: this fixes #1136. No blockers.** Both resolvers now re-check the actor row (fail-closed on missing rows), the refresh rotation loophole is closed with reuse-detection correctly still winning, introspection agrees, and the execute-path bound is genuinely ≤ ~3 s (`CachedTokenValidator` re-checks only expiry on cached entries, so the `active` verdict lives at most `resolve_cache_ttl_seconds`; negative verdicts are cached symmetrically, so re-enable lockout is also ≤ ~3 s). Independent run: 34 + 810 unit, 283 arch, 31 SQLite integration (both changed files), ruff + full mypy — all green. `ApiKeyResolver`'s "already status-gated" claim verified (`shared/auth/api_key_resolver.py:55-56, 80-81`).

### Worth fixing before merge (test integrity)

1. **`test_revoke_refresh_token_revokes_family` silently became vacuous** (`tests/integration/auth/test_token_endpoints.py:361-373`, untouched by this PR): `usr_fam1` is never seeded, so post-PR the fail-closed missing-actor path makes both `active is False` assertions pass **even if `revoke()` were a no-op**. It was the only real-DB pin of family revocation. One-line fix: seed the user active. (Same overdetermination, lower stakes: `test_introspect_expired_returns_inactive` / `usr_expired`.)
2. **The broker SQL's `service_account` and `user` CASE branches have zero tests** on either backend — and the `user` branch holds the dialect-riskiest code (the boolean→`'active'/'disabled'` normalisation). `test_auth_e2e.py` seeds agent tokens exclusively; two parametrized cases would close this.

### Worth a decision (silent contract change)

3. **Ephemeral task tokens can now mint dead-on-arrival**: `mint_task_token` (`service_account_auth_service.py:151-183`) never validates `target_agent_id`, but resolve now requires an ACTIVE agent row — so mint returns 200 and every use 401s. The old behavior was pinned by tests this PR flipped (`test_resolve_ephemeral_minted_token_keeps_snapshot`, broker e2e twin) without the change being named in the description. Either gate the mint (fail fast with `invalid_grant`) or state the new contract explicitly.

### Hardening minors (fine as follow-ups)

- Both resolvers **fail open on unknown actor types** (`_actor_is_active`'s final `return True`; SQL `ELSE 'active'`). Unreachable today (only agent/SA/user ever mint `at_` rows — verified all call sites), but a fourth token-bearing actor type would silently bypass this fix. If tightened, change both together.
- `DEFAULT_CACHE_TTL_SECONDS = 30` (`token_validation.py:15`) is 10× the production TTL; production wiring always overrides, but nothing pins that wiring — a `CachedTokenValidator` constructed without the parameter silently stretches the kill bound to 30 s.
- The refresh-gate comment says it's gated "like the jwt-bearer exchange does" — the assertion path distinguishes PENDING; refresh deliberately doesn't. Cosmetic.

### Pre-existing gaps confirmed out of scope — now tracked

The end-to-end trace confirmed disable is still not a *complete* kill switch; none of these are defects of this PR, and they're now filed: **#1138** (queued async executions run post-disable — `_WORKER_IDENTITY_ACTIVE = True`, the one path where a disabled actor still *acts*), **#1139** (disabled users keep login JWTs up to 1 h — the "no DB backing by design" claim is accurate for the broker's trusted-issuer path but not this one), **#1140** (owner-disable doesn't cascade to owned agents — policy decision).

Also noting for the rules repo (not this public PR): `unified-identity.md`'s "one DB lookup" description of the opaque path now drifts — every verdict consults the actor row too.

Great fix — the resolver seams are exactly right, the test churn is honest apart from finding 1, and the description's out-of-scope claims held up under adversarial tracing (with the one JWT nuance above).

ren-jentic:
Thanks for the deep review — all three actionable findings addressed in 0ad8e46:

**1. Vacuous tests (fixed).** `test_revoke_refresh_token_revokes_family` now seeds `usr_fam1` active and asserts `active is True` *before* the revoke, so the two `False` assertions can only come from family revocation, not the missing-actor fail-closed path. Same treatment for `test_introspect_expired_returns_inactive` / `usr_expired` so it pins expiry specifically.

**2. Broker SQL branch coverage (fixed).** Two parametrized tests in `test_auth_e2e.py` now pin the `user` branch (the boolean→status normalisation, active→resolves / deactivated→`token_inactive`) and the `service_account` branch (active/disabled) — verified on both Postgres and SQLite backends.

**3. Mint contract (decided: gate the mint).** `mint_task_token` now resolves `target_agent_id` and rejects with `invalid_grant` unless the agent exists and is ACTIVE — a 200 that mints a dead-on-arrival token is a trap, and failing fast matches the client-credentials/assertion posture. One deliberately indistinct message for missing vs non-active (no oracle for probing agent ids). Unit tests cover missing + all four non-active statuses, asserting no token is issued. The flipped snapshot tests keep their scope-freezing purpose with an active agent seeded.

Also fixed the cosmetic refresh-gate comment (now explains why PENDING gets no distinct detail on this path, rather than claiming parity with the jwt-bearer exchange).

**Hardening minors** — agreed as follow-ups rather than here: the unknown-actor-type fail-open needs both resolvers changed together and a decision on whether a fourth token-bearing actor type should fail closed by default; the `DEFAULT_CACHE_TTL_SECONDS = 30` drift is a one-liner but touches broker wiring guarantees beyond this PR's scope. Happy to file both if you want them tracked like #1138–#1140. The `unified-identity.md` "one DB lookup" drift I'll fix in the rules repo separately.

Full verification re-run: 2933 unit+arch, 965 integration (Postgres), 648 integration (SQLite), ruff + mypy — all green.

PR #1151, opened 2026-08-26
feat(auth): OAuth client registry, agent JWKS PUT, and token provenance

DavidAtJentic:
## Summary

Adds a third-party OAuth client registry, enabling external services to integrate with Jentic One via standard OAuth 2.0 Authorization Code + PKCE. Registered clients are confidential (secret-bearing) and admin-managed; deactivating a client immediately invalidates all tokens it issued.

A JWKS PUT endpoint for agents was also introduced, to allow external services to both create agents and allow them to authenticate without needing to go through the DCR flow.

## Changes

- auth: OAuth client registry- CRUD, secret rotation, redirect URI validation, per-client scope restriction (admin surface)
- auth: Authorization code exchange authenticates third-party clients via `client_secret`; platform clients remain public (PKCE-only)
- auth: `oauth_client_id` column on `access_tokens` / `refresh_tokens` - token resolution and refresh check issuing client active status
- auth: Consent flow hardening - state HMAC, nonce binding, scope display, XSS mitigation on consent page
- auth: `PUT /agents/{id}/jwks` endpoint for post-creation public key updates
- ui: OAuth Clients management section under Profile > Settings > Developer Settings (redirect URI list input, one-time secret display, secret rotation)
- admin: `PlatformClientConfig.redirect_uris` validated at startup; `redirect_uri` must match a registered URI on code exchange

## Risk & rollback

- Migration: adds `oauth_clients table` + `oauth_client_id` column + indexes on token tables - additive, no existing data modified
- Auth surface: new token validity check (client-active lookup) only fires when `oauth_client_id IS NOT NULL` - no impact on existing platform-issued tokens
- Revert: revert the squash commit + run downgrade migration

## Test plan

- `make check` passes (lint + typecheck + unit + arch)
- New unit tests: OAuth client service, consent flow, auth-code grant `oauth_client_id` propagation (platform=None, third-party=set), token service refresh/resolve with deactivated client

## Checklist

- [ ] Commits follow Conventional Commits with a scope and are signed off (`git commit -s`, DCO)
- [ ] `make check` passes (lint, type check, secrets audit, arch tests)
- [ ] No secrets, credentials, or internal-only references included

## Architecture Diagrams

```mermaid
sequenceDiagram
    participant App as External App
    participant UI as Jentic One UI
    participant Auth as /authorize
    participant IdP as Identity Provider
    participant Token as /oauth/token
    participant DB as Admin DB

    App->>Auth: GET /authorize (client_id, redirect_uri, code_challenge)
    Auth->>DB: Lookup OAuth client by client_id
    DB-->>Auth: Client record (redirect_uris, require_consent, allowed_scopes)
    Auth->>Auth: Validate redirect_uri against client's allowed URIs
    Auth->>IdP: Redirect to IdP (upstream code exchange)
    IdP-->>Auth: GET /oauth/callback (code)
    Auth->>Auth: Exchange upstream code, resolve user

    alt Client requires consent
        Auth->>UI: Show consent screen (scopes, client name)
        UI->>Auth: POST /oauth/consent (approved)
    end

    Auth-->>App: Redirect to app with authorization code

    App->>Token: POST /oauth/token (code, code_verifier, client_id, client_secret)
    Token->>Token: Verify client_secret against stored hash
    Token->>Token: Verify PKCE (code_verifier vs code_challenge)
    Token->>DB: Issue tokens with oauth_client_id = client_id
    Token-->>App: access_token + refresh_token + id_token
```

## Platform vs third-party client token issuance

```mermaid
flowchart TD
    A[POST /oauth/token<br/>grant_type=authorization_code] --> B{client_id in<br/>platform_clients config?}

    B -->|Yes: Platform client| C[PKCE-only verification<br/>No client_secret required]
    C --> D[issue_pair<br/>oauth_client_id = NULL]

    B -->|No: Third-party client| E{client_secret<br/>provided?}
    E -->|No| F[400 invalid_client]
    E -->|Yes| G[Verify secret against<br/>OAuthClientService]
    G -->|Invalid| F
    G -->|Valid| H[issue_pair<br/>oauth_client_id = client_id]

    D --> I[Access + Refresh tokens]
    H --> I
```

## Token provenance: invalidation on client deactivation

**Access token resolution:**

```mermaid
flowchart TD
    R1[resolve_access_token] --> R2{oauth_client_id<br/>is NULL?}
    R2 -->|Yes: platform/agent token| R3[Proceed normally]
    R2 -->|No: third-party token| R4[Lookup OAuth client<br/>by oauth_client_id]
    R4 --> R5{Client active?}
    R5 -->|Yes| R3
    R5 -->|No / deleted| R6[Return None<br/>token treated as invalid]
```

**Refresh token rotation:**

```mermaid
flowchart TD
    T1[refresh] --> T2{oauth_client_id<br/>is NULL?}
    T2 -->|Yes| T3[Rotate normally]
    T2 -->|No| T4[Lookup OAuth client]
    T4 --> T5{Client active?}
    T5 -->|Yes| T6[Rotate, propagate<br/>oauth_client_id to new pair]
    T5 -->|No / deleted| T7[InvalidGrantError<br/>issuing client deactivated]
```

## Agent JWKS update access control

```mermaid
flowchart TD
    A[PUT /agents/:id/jwks] --> B[Router: require agents:write permission]
    B --> C[AgentService.update_jwks]
    C --> D[Fetch agent by ID]
    D --> E{Agent exists?}
    E -->|No| F[404 ActorNotFoundError]
    E -->|Yes| G{Caller is owner<br/>OR has org:admin?}
    G -->|No| H[403 AgentWriteAccessDeniedError]
    G -->|Yes| I{Agent status<br/>= active?}
    I -->|No| J[409 InvalidTransitionError]
    I -->|Yes| K[Validate JWKS<br/>Ed25519 required, no private keys]
    K --> L[Update agent.jwks + audit log]
```


ren-jentic:
Posted feedback in the plans repo: https://github.com/jentic/jentic-one-plans/pull/28 (trying something different, but if you hate this can add to this instead)

Manuel-Jentic (commented):
Re-review at head `5baa7329` — delta since the 8-agent review in [jentic-one-plans#28](https://github.com/jentic/jentic-one-plans/pull/28) (grounded at `530234e5`) plus a follow-up pass at `722559ba`. All findings below were re-verified at head; suites were run locally at `6e1c808f` (the three commits since only touch CLI codegen, `x-sensitive` markers, and the trivy allowlist).

**Fixed since the last pass** (observed, not assumed): lint gate green (`ruff check` + `ruff format --check` both clean), `tests/arch` 285 passed, units 390 passed; auth code now minted only on consent approval with nonce anti-replay (`10402f80`); scope ceiling intersected at the token endpoint with tests (`11e1528a`, `928f5074`); deactivated clients rejected at verification; portable array types on `oauth_clients` (`f4a2c368`); token endpoint rate limited (`45f0deaa`); settings UI permission-gated with `RequirePermission`/`usePermission` (`3b99d723`); spec + generated TS/Go clients regenerated and consistent (`06f378e6`, `67580dbd`); `config-schema.json` now carries `platform_clients` with redirect-URI validation (`24172ab0`). Nice turnaround.

**Blocking**: one regression from `f7830d39` — the IdP state token is signed with a derived key but verified with the raw secret, so every `/oauth/callback` fails `invalid_state` (see inline).

**Still open — headline items** (full detail in plans#28):
- **B1**: consented scopes enforced at minting but still not at resolution — `auth/web/app.py:108-126` resolves USER actors to full grants via `resolve_permissions_for_actor(...)`; the token row's `scopes`/`oauth_client_id` are only honored for AGENT actors. A third-party token consented for `toolkits:read` still acts with the user's full grant set on routes using this verifier.
- **allowed_scopes has no UI surface** — the settings dialog and hand-rolled types omit it entirely, so every UI-created client gets an empty allowlist, which after `11e1528a` means deny-all non-OIDC scopes. Safe default, but almost certainly not the intended admin UX (see inline).
- **Rate limiters** are still hardcoded, per-process `MemoryStateBackend`, keyed on `request.client.host` — multiplied by worker count and one shared bucket for everyone behind the documented ALB (see inline).
- **Process**: 0/42 branch commits carry the DCO `Signed-off-by` that `CONTRIBUTING.md` requires (`git rebase --signoff` fixes in one pass), and the branch is 5 commits behind main again.
- Carried from plans#28: S9 (committed local signing key — trivy now allowlisted via `trivy-secret.yaml` but still no production guard), S12 (no operator/integrator docs for the registry), S13 (unreachable agent `client_credentials` fallback), N6 (JSON-only `/oauth/token`). Minor S18 leftovers in the dialog (conditional mount, `key={index}` on URI rows, secret retained in mutation cache).

src/jentic_one/auth/web/routers/authorize.py
```diff
@@ -160,8 +494,14 @@ async def oauth_callback(
 ) -> RedirectResponse:
     """External IdP callback — exchanges upstream code and issues platform auth code."""
     try:
-        params = _verify_state(state, ctx.config.admin.auth.jwt_secret.get_secret_value())
+        params = _verify_payload(
+            state,
+            ctx.config.admin.auth.jwt_secret.get_secret_value(),
```
Manuel-Jentic: **Blocking** — `f7830d39` updated the sign site (`L464`) to `_derive_key(jwt_secret, "state")` and both consent sites in both directions, but this verify still passes the **raw** `jwt_secret`. HMAC comparison can never match, so every IdP callback now redirects to `/error?error=invalid_state` — the entire login flow is broken. Fix: `_derive_key(ctx.config.admin.auth.jwt_secret.get_secret_value(), "state")` here too, plus a regression test that round-trips authorize → callback through `_sign_payload`/`_verify_payload`.

src/jentic_one/admin/services/oauth_client_service.py
```diff
@@ -0,0 +1,313 @@
...
+            client = await OAuthClientRepository.get_by_client_id(session, client_id)
+        if client is None or not client.active:
+            return False
+        return redirect_uri in client.redirect_uris
+
+    async def verify_client_secret(self, client_id: str, client_secret: str) -> bool:
+        """Verify a client's secret. Returns False if not found, inactive, or wrong secret."""
+        async with self._ctx.admin_db.session() as session:
+            client = await OAuthClientRepository.get_by_client_id(session, client_id)
+        if client is None or not client.active:
```
Manuel-Jentic, line 307: The `not client.active` check closes S1 — good. Remaining: this early return answers unknown/inactive `client_id`s in microseconds while known ones cost an argon2 verify (~100ms), a client-id enumeration oracle on the token endpoint. Verify a static dummy hash on this path before returning `False` to equalize timing.

src/jentic_one/admin/services/oauth_client_service.py
```diff
@@ -0,0 +1,313 @@
...
+        """
+        async with self._ctx.admin_db.session() as session:
+            client = await OAuthClientRepository.get_by_client_id(session, client_id)
+        if client is None:
+            return None
+        if client.allowed_scopes is not None:
+            return frozenset(client.allowed_scopes)
+        return None
+
+    async def requires_consent(self, client_id: str) -> bool:
```
Manuel-Jentic: nit: `requires_consent` still has zero callers — the callback reads `.require_consent` off `get_by_client_id` instead (`authorize.py:517`). Delete it or use it; a dead consent helper invites drift.

src/jentic_one/auth/services/agent_service.py
```diff
@@ -553,6 +555,44 @@ async def update_agent(
...
+        The agent must be active (not pending, disabled, or archived). The JWKS
+        is validated to ensure it contains at least one Ed25519 public key and
+        no private key material.
+        """
+        validate_jwks(jwks)
+        async with self._ctx.admin_db.transaction() as session:
+            agent = await AgentRepository.get_by_id_for_update(session, agent_id)
+            if agent is None:
+                raise ActorNotFoundError(agent_id)
+            if "org:admin" not in identity.permissions and agent.owner_id != identity.sub:
```
Manuel-Jentic, line 575: Right layer and predicate — but this raises 403 (`AgentWriteAccessDeniedError`) where the sibling owner-or-admin checks on the same resource (`agent_auth_service.py:121,142,173`) deliberately raise `ActorNotFoundError` (404) to hide existence. Pick one posture; two conventions for the identical denial will drift.

src/jentic_one/auth/web/routers/oauth.py
```diff
@@ -21,11 +23,28 @@
...
 from jentic_one.shared.web import get_current_identity
 from jentic_one.shared.web.deps import get_ctx
 
 router = APIRouter()
 
+_TOKEN_RATE_LIMIT_RPM = 60
+_token_rate_limit_store = MemoryStateBackend()
+_token_ip_rate_limiter = RateLimiter(
+    _token_rate_limit_store, default_rpm=_TOKEN_RATE_LIMIT_RPM, burst=60
+)
```
Manuel-Jentic: Covering the token endpoint closes the biggest gap (B5/S3) — but this limiter inherits the same weaknesses as `authorize.py:52`: hardcoded rpm, per-process `MemoryStateBackend` (limit × workers), keyed on `request.client.host` which is the proxy address behind the documented TLS-terminating ALB — one bucket throttles all clients (login/token DoS lever). Suggest: config fields for rpm/burst, shared state backend, key on trusted-proxy-resolved IP + `client_id`.

src/jentic_one/auth/web/routers/oauth.py
```diff
@@ -43,27 +62,47 @@ def get_sa_auth_service(ctx: Context = Depends(get_ctx)) -> ServiceAccountAuthSe
...
     return AuthorizeService(ctx)
 
 
-@router.post("/oauth/token")
+def get_oauth_client_service(ctx: Context = Depends(get_ctx)) -> OAuthClientService:
+    return OAuthClientService(ctx)
+
+
+@router.post("/oauth/token", dependencies=[Depends(_check_token_rate_limit)])
 async def token_endpoint(
```
Manuel-Jentic, line 185: `/oauth/token` still accepts JSON only — `schemas/oauth.py:10` even documents "not RFC 6749 form-encoded", and it's now frozen into the public spec. No stock OAuth client library (which all send `application/x-www-form-urlencoded` per RFC 6749 §4.1.3) can integrate without custom transport code. Accept both content types before external integrators build against this. (N6 in plans#28.)

src/jentic_one/auth/web/routers/oauth.py
```diff
@@ -112,6 +151,27 @@ async def token_endpoint(
     )
 
 
+async def _authenticate_client_credentials(
+    client_id: str,
+    client_secret: str,
+    sa_auth_svc: ServiceAccountAuthService,
+    agent_auth_svc: AgentAuthService,
+) -> tuple[str, str, int]:
+    """Try service account auth first, then agent auth."""
```
Manuel-Jentic: This agent `client_credentials` fallback is still unreachable: it requires `cred.client_secret_hash`, whose only writer `AgentAuthService.register_client_secret` has no non-test callers — no router or service ever mints an agent client secret. Either wire up secret issuance or drop the branch. (S13 in plans#28.)

src/jentic_one/auth/web/routers/authorize.py
```diff
@@ -62,34 +117,291 @@ def get_authorize_service(ctx: Context = Depends(get_ctx)) -> AuthorizeService:
...
         age = time.time() - float(iat)
-        if age > STATE_MAX_AGE_SECONDS or age < 0:
-            raise InvalidGrantError("state expired")
+        if age > max_age or age < 0:
+            raise InvalidGrantError(f"{purpose} token expired")
     return payload
 
 
-@router.get("/authorize")
+# Consumed consent nonce tracking (in-memory; acceptable for single-instance or
```
Manuel-Jentic: The consent-token anti-replay nonce set is per-process, as the comment acknowledges — but `10402f80`'s replay defense now depends on it: in a multi-worker/multi-instance deploy an approved consent token can be replayed against other workers to mint additional auth codes. Worth moving to the shared state backend alongside the rate limiter.

config/local.yaml
```diff
@@ -80,6 +80,31 @@ credentials:
...
 auth:
   canonical_base_url: "http://127.0.0.1:8000"
+  platform_clients:
+    - client_id: "jentic-one-spa"
+      redirect_uris:
+        - "http://127.0.0.1:8000/app/auth/callback"
+  # ES256 signing key for ID token issuance (local dev only - never use in prod)
+  id_signing:
+    - kid: "local-dev-key"
+      private_key_pem: |
```
Manuel-Jentic, line 90: The trivy allowlist (`trivy-secret.yaml`) quiets the scanner but the committed EC key remains, and `id_signing` has no production guard — `_require_production_secret` (`shared/config.py:35-49`) covers `jwt_secret`/`pepper`/`state_secret` but nothing rejects `local-dev-key` in production, where it would sign real id_tokens with a publicly-known private key. Add it to the guarded set. (S9 in plans#28.)

ui/src/modules/settings/api/client.ts
```diff
@@ -0,0 +1,106 @@
+/**
+ * OAuth clients API client — wraps the admin /oauth-clients endpoints.
+ * TODO: migrate to generated DefaultService (endpoints now in OpenAPI spec).
```
Manuel-Jentic: Codegen is now consistent with the spec (`06f378e6` — `OAuthClientsService` has rotate-secret, `client_secret`, `allowed_scopes`), but this hand-rolled client is still what `hooks.ts` imports; the generated service has zero app callers, and this TODO points at `DefaultService` which no longer exists. Port `hooks.ts` to `OAuthClientsService` and delete this file — it also loses `ApiError` semantics (403s retried, 422 detail arrays mis-render).

ui/src/modules/settings/pages/OAuthClientsSection.tsx
```diff
@@ -0,0 +1,575 @@
...
+			return;
+		}
+
+		try {
+			if (isEdit) {
+				await updateMutation.mutateAsync({
+					id: client.id,
+					input: {
+						name: name.trim(),
+						description: description.trim() || undefined,
```
Manuel-Jentic: Two things here: (1) `description.trim() || undefined` makes it impossible to *clear* a description — emptying the field omits it from the PATCH; the spec accepts `null` for clearing (`OAuthClientUpdateRequest.description?: string | null`). (2) The dialog has no `allowed_scopes` field at all, and after `11e1528a` an empty allowlist means deny-all non-OIDC scopes — so every client created through this UI cannot request any platform scope. Safe default, but if that's intended, the UI should say so; if not, the dialog needs a scopes editor.

src/jentic_one/auth/web/routers/authorize.py
Manuel-Jentic: Confirmed by in-process repro at `83fe9945`:

```python
token = _sign_payload({...}, _derive_key(secret, "state"), purpose="state")   # authorize (L464)
_verify_payload(token, secret, purpose="state", max_age=600)                  # callback (L499)
# InvalidGrantError: state signature invalid
```

`tests/web/auth/test_oidc_login_flow.py` will hit this as `invalid_state` on the `/oauth/callback` step once the fixture fix from `83fe9945` lets it get that far.

Manuel-Jentic: Verified at `6417cec3` — all three `_verify_payload` sites now use derived keys, matching both sign sites; the derived-key round-trip passes. Blocking issue resolved. (The OIDC login-flow integration test covers this path end-to-end, which satisfies the regression-test ask.)

src/jentic_one/auth/web/routers/oauth.py
```diff
@@ -47,23 +66,57 @@ def get_authorize_service(ctx: Context = Depends(get_ctx)) -> AuthorizeService:
     return AuthorizeService(ctx)
 
 
-@router.post("/oauth/token")
+def get_oauth_client_service(ctx: Context = Depends(get_ctx)) -> OAuthClientService:
+    return OAuthClientService(ctx)
+
+
+async def _parse_token_request(request: Request) -> TokenRequest:
```
Manuel-Jentic, line 149: The server-side dual parsing is correct — but moving the body out of the signature made the regenerated spec drop the request body entirely: `/oauth/token` in `openapi/control/control.openapi.yaml` and `ui/openapi.json` now has **no `requestBody`**, so the generated Go client builds a nil-body POST and the generated TS `tokenEndpoint()` takes no arguments. External integrators consuming the published spec can't make a valid token request. Re-declare the body on the operation (FastAPI `openapi_extra` with both `application/x-www-form-urlencoded` and `application/json` schemas, or equivalent) and regenerate.

src/jentic_one/auth/web/app.py
```diff
@@ -119,6 +119,11 @@ async def _verify(token: str, request: Request) -> Identity:
                 # broker's InProcessTokenResolver, which already reads row.scopes.
                 # parent_permissions (owner inheritance) is still resolved above.
                 permissions = list(resolved.permissions)
+            elif resolved.oauth_client_id is not None:
```
Manuel-Jentic, line 142: This closes B1's main bypass — but it *substitutes* the row scopes instead of *intersecting* them with the user's live grants. Since `/authorize` validates requested scopes only against the client allowlist (never the user's own permissions) and `allowed_scopes=None` means unrestricted, a consenting user can mint a token carrying scopes they don't hold, and revoking a user's grants no longer binds their outstanding third-party tokens until expiry. `permissions = [p for p in permissions if p in resolved.permissions]` (plus OIDC passthrough) keeps both properties. Also: the new `Identity` below drops `resolved.oauth_client_id`, so downstream consumers still can't distinguish third-party tokens.

src/jentic_one/auth/web/routers/authorize.py
```diff
@@ -62,34 +117,278 @@ def get_authorize_service(ctx: Context = Depends(get_ctx)) -> AuthorizeService:
...
         age = time.time() - float(iat)
-        if age > STATE_MAX_AGE_SECONDS or age < 0:
-            raise InvalidGrantError("state expired")
+        if age > max_age or age < 0:
+            raise InvalidGrantError(f"{purpose} token expired")
     return payload
 
 
-@router.get("/authorize")
+_consent_nonce_store = MemoryStateBackend()
```
Manuel-Jentic: The commit message says "move consent nonce tracking to state backend", but this is a fresh module-level `MemoryStateBackend()` — still per-process. The atomicity is right (`set_if_absent`), but in a multi-worker/multi-pod deploy an approved consent token can still be replayed against other workers to mint additional auth codes. Wire it to the configured shared backend (the Redis impl's `SET NX` gives the same atomicity) rather than instantiating a private in-memory one.

src/jentic_one/shared/config.py
```diff
@@ -275,6 +308,31 @@ class AuthConfig(BaseModel):
...
+    platform_clients: list[PlatformClientConfig] = Field(default_factory=list)
+
+    def model_post_init(self, __context: object) -> None:
+        """Ensure the SPA platform client is always registered.
+
+        If no platform_clients entry exists for the operator SPA and a
+        canonical_base_url is configured, synthesise one so existing
+        deployments continue to work without a config change on upgrade.
+        """
+        if os.environ.get("JENTIC_ENV", "development") == "production":
```
Manuel-Jentic, line 334: Guard matches by kid string only — copying the committed PEM from `config/local.yaml` under any other kid bypasses it. Match on the key material instead (e.g. compare against the known local-dev PEM's fingerprint, or reject any `id_signing` entry whose private key equals the committed one).

src/jentic_one/auth/web/routers/oauth.py
```diff
@@ -21,11 +23,28 @@
...
 from jentic_one.shared.context import Context
 from jentic_one.shared.models import ActorType
+from jentic_one.shared.resilience import RateLimiter
+from jentic_one.shared.state.backend import MemoryStateBackend
 from jentic_one.shared.web import get_current_identity
 from jentic_one.shared.web.deps import get_ctx
 
 router = APIRouter()
 
+_TOKEN_RATE_LIMIT_RPM = 60
```
Manuel-Jentic: Swapping in the platform `RateLimiter` fixes the leak/429 shape, but the three original asks still stand for both this limiter and `authorize.py:52`: hardcoded rpm/burst (no config fields), module-level `MemoryStateBackend` (limit multiplies per worker), and keyed on raw `request.client.host` — behind the shipped nginx gateway that's the proxy address, so all callers share one bucket. Config-driven limits, the shared backend, and trusted-proxy IP + `client_id` keying.

ui/src/modules/settings/pages/OAuthClientsSection.tsx
```diff
@@ -0,0 +1,723 @@
...
+}
+
+function CreateEditDialog({ open, onClose, onSecretRevealed, client }: CreateEditDialogProps) {
+	const [name, setName] = useState(client?.name ?? '');
+	const [description, setDescription] = useState(client?.description ?? '');
+	const [redirectUris, setRedirectUris] = useState<string[]>(
+		client?.redirect_uris.length ? client.redirect_uris : [''],
+	);
+	const [requireConsent, setRequireConsent] = useState(client?.require_consent ?? true);
+	const [restrictScopes, setRestrictScopes] = useState(
```
Manuel-Jentic: Bug: `restrictScopes` initializes to `!= null && length > 0`, so a deny-all client (`allowed_scopes: []`) opens the edit dialog with the restrict checkbox **off** — saving any change (even a rename) sends `allowed_scopes: null` and silently widens the client to unrestricted. Initialize with `client?.allowed_scopes != null` so the empty-allowlist state round-trips.

Manuel-Jentic (commented):
Follow-up pass at head `238c7688` (post-rebase). Excellent turnaround — verified locally: `ruff check`/`format` clean, `tests/arch` 285, full units 2676, `mypy` clean, and all live CI green.

**Confirmed fixed** (re-verified at head, not just by commit message): state-token derived key; B1's main bypass — third-party USER tokens now resolve to consented scopes (`43e6e4e4`) with first-party and AGENT behavior preserved; constant-time dummy hash on unknown `client_id` (`2ad204f8`); dead `requires_consent` removed; 404 posture on JWKS deny + tests; form-encoded token requests (server side); UI ported to generated `OAuthClientsService` with the hand-rolled client deleted; description clearable; `allowed_scopes` editor with accurate deny-all copy; dialogs conditionally mounted; permission gating intact through the rebase; DCO nearly done (54/56 — `78b4bb2d` and `8fd3d2a3` still lack `Signed-off-by`).

**Remaining / new** (inline): the regenerated spec dropped the `/oauth/token` request body entirely; scope *substitution* (vs intersection) leaves an escalation edge; the consent nonce store is still per-process despite the commit message; the prod key guard matches by kid only; rate limiters still hardcoded/in-memory/raw-IP; and one real UI bug — editing a deny-all client silently un-restricts it.

Minor leftovers, no inline needed: orphaned `AgentWriteAccessDeniedError` + its 403 mapping; orphaned `register_client_secret`/`authenticate_client_credentials` in `agent_auth_service.py` (only test callers); create-flow `client_secret` lingers in the mutation cache (rotate path got `.reset()`, create didn't); `key={`${baseId}-${index}`}` is still index-derived; S12 operator/integrator docs still absent.

src/jentic_one/auth/web/routers/authorize.py
```diff
@@ -1,43 +1,130 @@
...
+    _cached_authorize_limiter = RateLimiter(
+        backend, default_rpm=_AUTHORIZE_RATE_LIMIT_RPM, burst=_AUTHORIZE_RATE_LIMIT_BURST
     )
+    return _cached_authorize_limiter
+
+
+def _client_rate_key(request: Request, client_id: str | None = None) -> str:
+    forwarded = request.headers.get("x-forwarded-for")
+    if forwarded:
+        ip = forwarded.split(",")[0].strip()
```
Manuel-Jentic: Rate-limit bypass: this takes the **first** `X-Forwarded-For` entry unconditionally (same in `oauth.py:59-63`). The first XFF hop is client-controlled — any direct caller, or any client behind a proxy that appends rather than replaces, can rotate a header string and defeat every pre-auth limit on `/authorize`, `/oauth/callback`, `/oauth/consent`, and `/oauth/token`. XFF should only be honoured when `request.client.host` is a configured trusted proxy, and then take the rightmost untrusted hop — otherwise fall back to the socket address.

src/jentic_one/auth/web/routers/oauth.py
```diff
@@ -21,11 +23,50 @@
...
+
+async def _check_token_rate_limit(request: Request) -> None:
+    """Per-client+IP rate limiter for the token endpoint."""
+    forwarded = request.headers.get("x-forwarded-for")
+    if forwarded:
+        ip = forwarded.split(",")[0].strip()
+    else:
+        ip = request.client.host if request.client else "unknown"
+    limiter = _get_token_limiter(request)
+    outcome = await limiter.acquire(ip)
```
Manuel-Jentic: Two gaps vs the docstring and the original ask: (1) "Per-client+IP" — but the key here is IP-only, so all tenants behind one corporate NAT share a single 60 rpm bucket for token exchange while `authorize.py:82-84` did get the `{client_id}:{ip}` composite; include the request's `client_id` once parsed. (2) rpm/burst are still hardcoded module constants (`oauth.py:33-34`, `authorize.py:52-53`) with no config fields or schema entries — operators can't tune them per deployment.

src/jentic_one/shared/config.py
```diff
@@ -260,6 +263,47 @@ class IdpConfig(BaseModel):
...
+_SPA_CALLBACK_PATH = "/app/auth/callback"
+
+_LOCAL_DEV_KEY_FINGERPRINT = "cdf9bf0510b631d8222d7b3be8766165b917e5c0a607151dc11272d7e68e2a11"
+
+
+def _signing_key_fingerprint(pem: str) -> str:
+    """SHA-256 fingerprint of the DER-encoded key material inside a PEM."""
+    lines = [line for line in pem.strip().splitlines() if not line.startswith("-----")]
+    raw = base64.b64decode("".join(lines))
+    return hashlib.sha256(raw).hexdigest()
```
Manuel-Jentic: The fingerprint is over the raw DER bytes inside the PEM, so re-encoding the identical private key as PKCS#8 (`-----BEGIN PRIVATE KEY-----`) produces a different DER and sails past the guard — reproduced locally: SEC1 fingerprint `cdf9bf05…` (blocked) vs PKCS#8 of the same key `f1356388…` (accepted). And since the kid check was removed rather than kept alongside, the bypass works even with `kid: local-dev-key`. Compare parsed key material instead — load the PEM with `cryptography` and fingerprint the public-key SPKI (stable across encodings) — and add a unit test next to the jwt_secret/pepper guard tests.

src/jentic_one/auth/web/routers/authorize.py
```diff
@@ -1,43 +1,130 @@
...
-        return False
-    return (
-        parsed_redirect.scheme == parsed_canonical.scheme
-        and parsed_redirect.netloc == parsed_canonical.netloc
+
+def _resolve_auth_backend(request: Request) -> SharedStateBackend:
+    backend: object = getattr(request.app.state, "auth_state_backend", None)
+    if isinstance(backend, SharedStateBackend):
+        return backend
+    return MemoryStateBackend()
```
Manuel-Jentic, line 62: Fallback + caching interaction: `_resolve_auth_backend` mints a fresh `MemoryStateBackend()` when `app.state.auth_state_backend` is missing, and the module-global caches (`_cached_authorize_limiter` here, `_cached_token_limiter` in `oauth.py:36`, `_consent_nonce_backend` below) pin whatever the **first request** resolved — forever, process-wide. If the first request hits an app that skipped `install_on_app`, the in-memory backend is cached permanently even after a properly wired app is available, silently reverting the cross-worker replay/limit fixes; it also leaks state across app instances in one process (tests, multi-app assemblers). Cache on `app.state` instead of module globals, and consider failing loudly (or logging) when the fallback engages outside tests.

src/jentic_one/auth/web/app.py
```diff
@@ -119,6 +121,9 @@ async def _verify(token: str, request: Request) -> Identity:
...
+                consented = set(resolved.permissions)
+                permissions = [p for p in permissions if p in consented]
```
Manuel-Jentic, line 144: The intersection is right and revocation now binds outstanding tokens — two caveats: (1) OIDC scopes (`openid`/`email`/`profile`) on the token row are silently dropped here since they never appear in `resolve_permissions_for_actor` output; today that's only cosmetic (`scopes=identity.permissions` echoes), but any future consumer gating on `openid` will mysteriously fail for third-party tokens — pass them through explicitly. (2) This branch has no tests: `test_combined_verifier.py` doesn't cover intersection, first-party passthrough, or `oauth_client_id` propagation — worth locking in before merge.

Manuel-Jentic (commented):
Pass-3 follow-up at head `667b71b5` (tree identical to `0794ff35`; the force-push was the DCO rebase — all 63 commits now carry `Signed-off-by`, thanks). Verified locally at this tree: `ruff check`/`format` clean, `tests/arch` 285, full units 2675, `mypy` clean; live CI green/pending, none failing.

**Confirmed fixed** (re-verified at head, not just by commit message): `/oauth/token` requestBody restored via `openapi_extra` with both content types, consistent across `control.openapi.yaml`, `ui/openapi.json`, and the regenerated Go + TS clients (`df793817`); third-party USER tokens now **intersect** live grants with consented scopes and `Identity.oauth_client_id` is carried through (`9bbc3a77`); consent nonce store wired to the shared configured backend with atomic `set_if_absent` (`dd31ed7f`); orphaned `AgentWriteAccessDeniedError` + agent client-secret plumbing fully removed, zero references left (`618a52da`); UI deny-all edit no longer silently un-restricts and URI rows use stable keys (`da6e9937`); DCO complete.

**Remaining** (inline, all reproduced or read at head): rate limits are bypassable via a spoofed `X-Forwarded-For`; the token limiter keys on IP only (docstring says per-client+IP) and rpm/burst are still hardcoded; the id_signing production guard can be bypassed by re-encoding the same key as PKCS#8 (and the kid check was dropped rather than kept alongside); module-global limiter/backend caches can permanently pin the in-memory fallback; and two caveats on the scope-intersection fix (OIDC scope passthrough absent, verifier branch untested).

No inline needed: S12 operator/integrator docs are still absent, and the create-flow `client_secret` still lingers in the mutation cache (rotate got `.reset()`, create didn't).

src/jentic_one/auth/services/token_service.py
```diff
@@ -185,6 +193,13 @@ async def refresh(self, refresh_token: str) -> tuple[str, str]:
             if rt.expires_at <= datetime.now(UTC):
                 raise InvalidGrantError("refresh token expired")
 
+            if rt.oauth_client_id is not None:
+                oauth_client = await OAuthClientRepository.get_by_client_id(
+                    session, rt.oauth_client_id
+                )
+                if oauth_client is None or not oauth_client.active:
+                    raise InvalidGrantError("issuing OAuth client has been deactivated")
```
Manuel-Jentic, line 234: **Blocking — the refresh grant is not client-authenticated, so a leaked refresh token is a standalone credential.** This new block correctly rejects a refresh whose *issuing* client was deactivated, but nothing ever authenticates the *presenter* against `rt.oauth_client_id`. Traced the only call path: `oauth.py:191-197` reads `body.refresh_token` and calls `token_svc.refresh(...)` with no `client_id`/`client_secret` — while the `authorization_code` branch 50 lines earlier (`oauth.py:146-151`) does enforce the secret. So for a confidential third-party client, client auth happens once at code exchange and never again; whoever holds the `rt_` value can rotate it indefinitely from anywhere, and the rotation even copies `oauth_client_id` forward (line 240/251) so the new pair looks legitimately client-bound. RFC 6749 §6 requires client authentication on refresh for confidential clients. Suggest requiring `client_id`+`client_secret` on the refresh branch whenever the token row has a non-null `oauth_client_id`, and comparing the authenticated client to it (mismatch → `invalid_grant`, and arguably revoke the family).

Manuel-Jentic (commented):
Pass-4 fresh-lens review at head `667b71b5` (unchanged since pass 3; CI fully green, and I re-ran locally at this tree: `ruff check`/`format` clean, `mypy` clean on 613 files, `tests/arch` 285, full units 2675). Nothing here repeats passes 1-3 — this lap deliberately looked at three areas the earlier passes hadn't: token/client lifecycle and persistence, protocol conformance, and what operators will actually live with.

The headline is one **blocking** item: the PR introduces confidential clients and enforces `client_secret` on the `authorization_code` grant, but the `refresh_token` grant accepts no client credentials at all, so client authentication is effectively one-shot at code exchange. Second-most important: the client-deactivation kill-switch this PR builds (`oauth_client_id` on token rows, active-checks in the resolver) **does not apply to execution traffic** — the broker's own resolver is untouched by the PR and never consults `oauth_clients`.

Also worth flagging as a regression risk against pass-2/3's own fix: the consent-nonce store and both new rate limiters are keyed off `broker.resilience.backend`, which defaults to in-memory — and the `config/production.yaml.example` this PR ships configures no Redis. So in the documented production config, the cross-worker anti-replay and rate limiting we asked for silently degrade to per-process.

**Smaller items not worth an inline** (all verified at head): `PATCH {active: true}` fully resurrects a deactivated client *and* all its never-revoked tokens, since deactivate touches no token rows — worth documenting or revoking token families on deactivate. `allowed_scopes` items are unvalidated and unbounded at the API boundary but land in `ARRAY(String(128))` on Postgres vs unbounded JSON on SQLite, so a >128-char scope 500s on PG and persists on SQLite. `rotate_secret` audits as `UPDATE`/`{"secret_rotated": true}` rather than the established `AuditReason.CLIENT_SECRET_ROTATED` used by service-account rotation, so reason-based queries miss it. And `issue_id_token` derives `iss` from `auth.canonical_base_url`, which defaults to `""` with no startup validation — an unconfigured deployment mints ID tokens with an empty issuer that every conforming client rejects, rather than failing loudly.

**Checked and clean** (so you know where I looked): refresh-token reuse detection revokes the family atomically under `FOR UPDATE`; auth codes are consumed check-and-set inside one transaction; PKCE is S256-only with a `compare_digest` check, and both `client_id` and exact `redirect_uri` are re-verified at exchange; no open redirect — every error path with an unvalidated target goes to `/error`, never to a client URI; the consent HTML escapes all client-supplied strings and ships `X-Frame-Options: DENY` + `frame-ancestors 'none'`; no secrets in structlog calls or exception messages; `/oauth/mint` can't launder a client-bound token into an unbound ephemeral one; migration upgrade/downgrade is symmetric and dialect-portable; admin CRUD audits carry actor + before/after; the config schema is byte-identical to the CLI copy and CI-gated.

src/jentic_one/migrations/admin/versions/aa6b7c8d9e0f_add_oauth_clients.py
```diff
@@ -0,0 +1,95 @@
...
+        ),
+        sa.Column("created_by", sa.String(255), nullable=True),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_index("ix_oauth_clients_client_id", "oauth_clients", ["client_id"], unique=True)
+    op.create_index("ix_oauth_clients_created_at", "oauth_clients", ["created_at"])
+    op.create_index("ix_oauth_clients_created_by", "oauth_clients", ["created_by"])
+    op.create_index("ix_oauth_clients_active", "oauth_clients", ["active"])
+
+    # Add oauth_client_id to token tables — tracks which third-party OAuth client
```
Manuel-Jentic, line 72: **The kill-switch this comment promises doesn't reach execution traffic, and no query uses these indexes.** `git diff origin/main...HEAD -- src/jentic_one/broker/` is empty, and the broker's resolver — `InProcessTokenResolver`, wired in at `broker/core/setup.py:47` and used by `CachedTokenValidator.validate` (`broker/core/token_validation.py:68`) — is raw SQL that selects neither `t.oauth_client_id` nor joins `oauth_clients` (`broker/repos/token_resolver.py:41-55`). So on the broker path a deactivated client's access tokens keep resolving `active=true` until expiry, and the live `allowed_scopes` narrowing that `TokenService.resolve_access_token` applies (391-399, 420-421) never happens there either — the scope ceiling is enforced on the auth surface but not where tools actually execute. Separately, grepping every `oauth_client_id` reference shows no query filters tokens by it (only per-row reads and inserts) and there is no `revoke_by_oauth_client` anywhere, so both `ix_*_oauth_client_id` indexes are pure write-amplification on the hottest tables. Either add the client join/ceiling to the broker resolver and implement bulk revocation on deactivate, or drop the indexes and soften the comment.

src/jentic_one/auth/services/token_service.py
```diff
@@ -371,6 +388,16 @@ async def resolve_access_token(self, token: str) -> Identity | None:
             if at is None:
                 return None
 
+            client_scope_ceiling: frozenset[str] | None = None
```
Manuel-Jentic, line 447: This new client active-check + scope ceiling is the right shape, but `introspect()` (310-355) never got the same treatment, which makes its own docstring wrong: it promises "`active` reflects the same verdict the resolvers enforce" (313-315), yet both its access branch (325-329) and refresh branch (341-346) check only `revoked_at`/`expires_at`/`_actor_is_active` — no `oauth_client_id` lookup. Net effect: after an admin deactivates a client, RFC 7662 introspection reports its tokens `active: true` while this resolver rejects them, so an operator verifying the kill-switch gets the wrong answer from the endpoint built for exactly that. Also, the scopes reported by introspection are the raw mint-time snapshot, not the ceiling-filtered set computed at 420-421.

src/jentic_one/auth/web/app.py
```diff
@@ -61,8 +62,9 @@ def get_exception_handlers() -> list[tuple[type[Exception], Any]]:
 
 
 def install_on_app(app: FastAPI, ctx: Context) -> None:
-    """Install the auth token verifier on the app for shared identity resolution."""
+    """Install the auth token verifier and shared state on the app."""
     app.state.verify_token = make_superset_verifier(ctx)
+    app.state.auth_state_backend = build_state_backend(ctx.config.broker.resilience.backend)
```
Manuel-Jentic: **This silently undoes the cross-worker guarantee from the last round in the documented production config.** The consent-nonce store and both new rate limiters all resolve through `app.state.auth_state_backend`, which is built from the **broker's** `resilience.backend` — that defaults to `BackendKind.MEMORY` (`shared/state/factory.py:40`), and the `config/production.yaml.example` this PR touches sets no Redis backend (only a commented-out `# backend: local`). So a stock multi-worker production deploy gets per-process consent nonces (the same nonce is replayable on a sibling worker/instance) and per-process rate limits (limits multiplied by worker count) with no warning. Two suggestions: give the auth surface its own explicit state-backend config (or at minimum log loudly at startup when a security-critical store lands on MEMORY with >1 worker), and set the backend in the production example. Minor, same line: this second backend instance is never closed — the broker `aclose()`s only its own (`broker/web/app.py:199`), so with Redis configured you leak a connection pool on shutdown.

src/jentic_one/admin/web/schemas/oauth_clients.py
```diff
@@ -0,0 +1,125 @@
...
+    name: str | None = Field(default=None, max_length=255)
+    description: str | None = None
+    redirect_uris: list[Annotated[str, Field(max_length=2048)]] | None = Field(
+        default=None, min_length=1, max_length=20
+    )
+    active: bool | None = None
+    require_consent: bool | None = None
+    allowed_scopes: list[str] | None = Field(
+        default=None,
+        description="Scope restriction list. Null means no change; "
```
Manuel-Jentic, line 61: **This description is inverted relative to the behaviour, and the failure mode is a silent lockout.** "An empty list clears the restriction (all scopes permitted)" — but `[]` is not `None`, so the repo *stores* it (`oauth_client_repo.py:101-102`), and `[]` is then interpreted as **deny-all**: `get_allowed_scopes` says so explicitly ("An explicit empty list means 'deny all non-OIDC scopes'", `oauth_client_service.py:297-308`) and `resolve_access_token` filters every outstanding token's scopes through `frozenset()` (`token_service.py:420-421`). So an admin who follows this doc to lift a restriction instead strips all non-OIDC scopes from the client *and* from its already-issued tokens. Compounding it, because `None` means "no change", there is no way through the API to reset the column back to `NULL` once set. Either fix the description and add an explicit sentinel for "unrestrict", or make `[]` genuinely clear it — but the two docstrings must stop disagreeing.

src/jentic_one/auth/web/routers/authorize.py
```diff
@@ -172,8 +540,48 @@ async def oauth_callback(
...
+                "client_name": oauth_client.name,
+                "client_description": oauth_client.description,
+                "user_email": user_email,
+                "iat": str(int(time.time())),
+            },
+            _derive_key(ctx.config.admin.auth.jwt_secret.get_secret_value(), "consent"),
+            purpose="consent",
+        )
+        return RedirectResponse(
+            url=f"/oauth/consent?consent_token={consent_token}", status_code=302
```
Manuel-Jentic: The consent token is a bearer credential carrying `user_id`, `user_email`, and the requested scopes, and it's passed in a **GET query string** — so it lands in proxy/LB access logs, referrer-adjacent tooling, and browser history. That would be tolerable if the token were single-use on arrival, but the nonce is only consumed in the POST handler (`_consume_consent_nonce`, ~723); the GET `consent_page` (655-656) merely verifies the signature. So anyone with access-log read (or shared-machine history) can replay the token to `POST /oauth/consent` and grant consent as that user within the 5-minute TTL, no session required. Consider a POST-redirect (auto-submitting form) or handing the browser an opaque single-use handle whose payload is server-side, keeping the signed material out of the URL.

src/jentic_one/auth/web/routers/authorize.py
```diff
@@ -205,6 +613,150 @@ async def error_page(error: str = Query(default="unknown_error")) -> dict[str, s
...
+    consent_nonce = params.get("consent_nonce") or ""
+    if not await _consume_consent_nonce(consent_nonce, request):
+        logger.warning("oauth_consent_nonce_replay", nonce=consent_nonce[:8])
+        return RedirectResponse(url="/error?error=invalid_consent", status_code=302)
+
+    redirect_uri = params.get("redirect_uri") or ""
+    original_state = params.get("original_state")
+    client_id = params.get("client_id") or ""
+
+    if action == "deny":
```
Manuel-Jentic, line 785: Consent approve and deny emit **no audit record** — `authorize.py` doesn't even import `record_audit`. The only trace is `logger.info("oauth_consent_denied", client_id=...)` here (and `..._approved` at ~740), neither carrying the user, so a security team cannot answer "which user authorised which third-party app, when, for which scopes" — the single most audit-worthy event this PR introduces. `AuditAction.APPROVE`/`DENY` already exist (`shared/models/audit.py:21-22`). Two related gaps: the token-issuance audit in `issue_pair` (`token_service.py:160-168`) omits the new `oauth_client_id` from its `after` payload, and the `LOGIN` audit in `exchange_code` (`authorize_service.py:276-285`) carries no client either — so a third-party token grant is indistinguishable in the audit log from a first-party SPA login.

src/jentic_one/auth/web/routers/oauth.py
```diff
@@ -47,23 +88,73 @@ def get_authorize_service(ctx: Context = Depends(get_ctx)) -> AuthorizeService:
...
     """Exchange a refresh token, JWT assertion, authorization code, or client creds for tokens."""
     if body.grant_type == _AUTHORIZATION_CODE_GRANT:
         if not body.code or not body.code_verifier or not body.redirect_uri or not body.client_id:
             raise InvalidGrantError("code, code_verifier, redirect_uri, and client_id are required")
+        is_platform = any(pc.client_id == body.client_id for pc in ctx.config.auth.platform_clients)
+        third_party_client_id: str | None = None
+        if not is_platform:
+            if not body.client_secret:
+                raise InvalidGrantError("invalid_client")
+            if not await oauth_client_svc.verify_client_secret(body.client_id, body.client_secret):
```
Manuel-Jentic, line 222: Three conformance/interop gaps around this new secret check. (1) **The metadata doesn't advertise it**: `discovery.py:57` (untouched by this PR) still returns `"token_endpoint_auth_methods_supported": ["private_key_jwt", "none"]`, so a client that configures itself from RFC 8414 metadata concludes it cannot present a secret at all — the auth method this PR adds is `client_secret_post` and needs to be listed. (2) **No HTTP Basic support**: there is no `Authorization` header parsing anywhere in this router, and RFC 6749 §2.3.1 says the AS MUST support Basic for clients issued a password; several common client libraries try Basic first, so this is a real integration trap. (3) **No `Cache-Control: no-store`** on token responses — RFC 6749 §5.1 requires it, and grepping the whole auth surface finds no `no-store` anywhere; the same applies to the consent HTML, whose headers (`authorize.py:154-158`) cover clickjacking and sniffing but not caching, and that page embeds the user's email.

src/jentic_one/auth/web/errors.py
```diff
@@ -30,6 +31,7 @@
     NoApiKeyError: (409, "no_api_key"),
     ToolkitBindingConflictError: (409, "toolkit_binding_conflict"),
     ToolkitBindingNotFoundError: (404, "toolkit_binding_not_found"),
+    RateLimitExceededError: (429, "rate_limit_exceeded"),
```
Manuel-Jentic, line 48: This maps the new limiter to a bare 429 with no `response_hook`, so the `retry_after` that `authorize.py:94` / `oauth.py:67` carefully attach to the exception — and `RateLimitOutcome.headers()` — are discarded: clients get no `Retry-After`/`RateLimit-*` and can only guess. The broker already has the pattern to copy, attaching those headers and incrementing `broker.rate_limited_total` (`broker/web/deps.py:40-43, 162-168`); the new auth limiters emit neither headers nor a metric, so rate-limit pressure on the login surface is invisible in telemetry. Related failure mode, same error map: `redis.ConnectionError` is mapped nowhere and neither `RedisStateBackend` nor `RateLimiter.acquire` catches it, so a Redis outage turns every `/authorize`, `/oauth/callback`, `/oauth/consent`, and `/oauth/token` request (including platform-SPA refresh) into an unmapped 500 — total login outage. Failing closed is defensible, but it should surface as a mapped 503 with a structured log, the way `DatabaseUnavailableError` does (`auth/web/errors.py:45-56`).

src/jentic_one/admin/web/routers/oauth_clients.py
```diff
@@ -0,0 +1,149 @@
...
+        created_at=view.created_at,
+        updated_at=view.updated_at,
+        created_by=view.created_by,
+    )
+
+
+@router.post("/admin/oauth-clients", status_code=201, summary="Register OAuth client")
+async def create_oauth_client(
+    body: OAuthClientCreateRequest,
+    identity: Identity = get_current_identity(required_permissions=["org:admin"]),
```
Manuel-Jentic: **All six endpoints here require `org:admin`, which breaks this codebase's own permission pattern and leaves the new resource outside the permission system entirely.** Every sibling admin router uses fine-grained resource permissions — `users:read`/`users:write` (`users.py:56,74,…`), `audit:read` (`audit.py:47`), `config:read`/`config:write` (`config.py:35,56`), and `actors.py:18` uses `users:read` — and `permission_catalog.py:33-56` declares a `:read`/`:write` pair for all 13 existing resources. It gained **no oauth-client entries at all** (grepped: zero `oauth` matches in the catalog), so: an auditor can't be granted read-only visibility into registered clients; registering a client can't be delegated without handing over full org admin (`org:admin` implies everything per `permissions.py:9`); and the resource never appears in the permission catalog the roles UI and `/admin/permissions` are driven from. Suggest adding `oauth-clients:read`/`oauth-clients:write` to the catalog and using them here (read on GET/list, write on POST/PATCH/DELETE/rotate), which also gives you the natural place to require something stronger for the consent-skip flag below.

src/jentic_one/auth/web/routers/authorize.py
```diff
@@ -172,8 +540,48 @@ async def oauth_callback(
     original_state = params.get("original_state")
 
     callback_uri = _callback_uri(request, ctx.config.auth.canonical_base_url)
+
+    oauth_client = await OAuthClientService(ctx).get_by_client_id(client_id or "")
+    needs_consent = oauth_client is not None and oauth_client.require_consent
```
Manuel-Jentic: **The consent screen is fully bypassable per-client by configuration, and nothing about consent is ever persisted — so users get neither a record nor a way to revoke.** `needs_consent` is just `oauth_client.require_consent`, and that flag is settable at creation (`schemas/oauth_clients.py:36`, default `True` — good) *and* flippable on an existing client via `PATCH` (`:57`). With it false, a third-party client completes the flow silently: the user authenticates to the IdP and a code is minted for the client with no prompt, no indication that a third party now holds a token acting as them. The description says "Set to false for trusted first-party integrations", but nothing enforces first-partyness — any API-registered third-party client can have it. Worse, I grepped for any consent persistence (`class *Consent*`, `consent_grant`, `user_consent`, `ConsentRecord`): **there is none**, and there's no consent table in the migration. So (a) there's no record of which user approved which scopes for which client — and since consent also emits no audit event (my pass-4 inline), the approval is unrecoverable after the fact; (b) there is no "connected applications" surface where a user can review or revoke third-party access, which is the standard counterpart to an OAuth client registry and the only user-facing way to undo a grant; (c) as a side effect, consent is re-prompted on every single authorization since nothing is remembered. At minimum I'd restrict consent-skip to config-declared platform clients rather than API-registered ones, and treat persisted grants + user-facing revocation as required follow-up before this is advertised to third parties.

src/jentic_one/auth/web/routers/authorize.py
```diff
@@ -1,43 +1,130 @@
...
+    """Validate redirect_uri against platform clients (config) or registered clients (DB).
+
+    Platform clients are validated against their configured redirect_uris.
+    Third-party clients are looked up in the oauth_clients registry.
+    Unknown client_ids are rejected.
+    """
+    if _is_platform_client(client_id, ctx):
+        return _platform_client_allows_redirect(redirect_uri, client_id, ctx)
+
+    return await OAuthClientService(ctx).is_redirect_uri_allowed(client_id, redirect_uri)
```
Manuel-Jentic: Per-request cost on the login path: `_is_allowed_redirect_uri` constructs `OAuthClientService(ctx)` here and `_get_client_allowed_scopes` constructs another one 7 lines below, and each service method opens its **own** `admin_db.session()` and issues its own `get_by_client_id` for the *same* `client_id` (`oauth_client_service.py:280-286, 297-308`). So every `/authorize` for a third-party client does two independent connection checkouts and two identical row reads, before the flow has authenticated anyone — which is also the surface the new rate limiter is protecting, so it's the one place cheap-per-request matters most. `/oauth/callback:544` then re-reads the same row a third time, and `verify_client_secret` a fourth at token exchange. Suggest fetching the client row once per request (a small request-scoped cache, or just resolve it at the top of the endpoint and pass it into both checks). Same shape at `token_service.py:392-395`: `resolve_access_token` now adds a client lookup to *every* authenticated request whose token carries `oauth_client_id` — fine as-is since it shares the session, but worth knowing it's on the hot path for all third-party API traffic.

tests/unit/admin/services/test_oauth_client_service.py
```diff
@@ -0,0 +1,120 @@
...
+    ctx = MagicMock()
+    mock_session = AsyncMock()
+    ctx.admin_db.session.return_value.__aenter__ = AsyncMock(return_value=mock_session)
+    ctx.admin_db.session.return_value.__aexit__ = AsyncMock(return_value=False)
+    return ctx
+
+
+@patch("jentic_one.admin.services.oauth_client_service._verify_password_async")
+@patch("jentic_one.admin.services.oauth_client_service.OAuthClientRepository")
+async def test_verify_client_secret_rejects_inactive_client(
```
Manuel-Jentic, line 91: The test surface is skewed toward the easy half. This file has 11 redirect-URI validation cases plus these two `verify_client_secret` tests, but **nothing** covers `rotate_secret` (old secret rejected + new one accepted after rotation — the security-critical assertion), `deactivate`, `update`, or that any of them emit the audit records the service now writes. More notably, `require_consent` has **zero coverage anywhere in the repo** — grepping `tests/` for `require_consent` returns nothing (all 8 hits for that grep pattern are `allowed_scopes`), so no test asserts that `require_consent=True` actually gates the flow or that `False` skips it. That's the primary user-facing security control this PR introduces, and it's the one branch (`authorize.py:545`) nothing pins down; `test_authorize_router.py` only exercises the state-signing and callback-URI helpers, never the endpoint's redirect-URI rejection, scope-allowlist rejection, or consent branch. Worth adding before merge, since these are exactly the behaviours a future refactor could silently invert.

ui/src/modules/settings/api/hooks.ts
```diff
@@ -0,0 +1,80 @@
...
+			OAuthClientsService.updateOauthClient({ id, requestBody: { active: true } }),
+		onSuccess: () => {
+			void qc.invalidateQueries({ queryKey: QUERY_KEY });
+		},
+	});
+}
+
+export function useRotateOAuthClientSecret() {
+	return useMutation({
+		mutationFn: (id: string) => OAuthClientsService.rotateOauthClientSecret({ id }),
```
Manuel-Jentic, line 83: `useRotateOAuthClientSecret` is the only mutation here without an `onSuccess` → `invalidateQueries`, so after a rotation the list keeps serving the cached row and its `updated_at` (which `rotate_secret` does bump server-side, `oauth_client_service.py:268-277`) goes stale until something else refetches. Worth mirroring the other four hooks for consistency. Separately — not a defect, just flagging the consequence: `useReactivateOAuthClient` (line 58) makes "deactivate then reactivate" a one-click operator action, and because deactivation revokes no tokens, reactivating restores every outstanding access/refresh token that was issued before it. That's the resurrect path I mentioned in the pass-4 body; a UI button for it argues for either revoking token families on deactivate or warning in the reactivate confirmation.

Manuel-Jentic (commented):
Pass-5 review at head `667b71b5` — still unchanged, CI still fully green, no replies yet on the pass-3/pass-4 threads. Since the code hasn't moved, this lap deliberately went after four angles none of the first four passes touched: **the authorization model on the new admin surface**, **end-user consent visibility**, **per-request cost**, and **whether the tests actually assert the security behaviour being claimed**. Nothing here repeats earlier findings.

The two most substantive items are both about who is in control. First, the new admin resource is gated entirely on `org:admin` while every sibling router in this codebase uses fine-grained `<resource>:read`/`:write` — and `permission_catalog.py` gained no oauth-client entries at all, so a whole new resource class is invisible to the permission system and can't be delegated or granted read-only. Second, and more consequential for users: `require_consent=False` is freely settable through the admin API and turns the consent screen into a no-op for that client — and because consent is **never persisted anywhere** (no table, no grant rows), there is no record of what any user approved and no surface where a user can see or revoke which third-party applications hold tokens acting as them. For a PR whose whole purpose is third-party OAuth access, that's the missing half of the feature.

**Credit where it's due**, since I went looking for problems and didn't find them: the `ui/eslint.config.js` change is legitimate, not a loosened rule — the repo has a guard requiring every module dir to declare its query-key roots, and the new `settings` entry satisfies it. The `allowed_scopes` resolve semantics are genuinely well tested (intersection, `[]` deny-all, `None` unrestricted, all at `test_token_service.py`), which is precisely why my pass-4 note on the update schema is a **documentation** bug rather than a behavioural one — the code does what the tests say, the API description just describes the opposite. Redirect-URI validation is thoroughly tested (11 cases including `javascript:` and non-localhost http). The consent XSS and consent-verification suites this PR adds are real tests, not smoke.

**Body-only notes:** `rotate_secret` invalidates the old secret with no overlap window, so rotation is a hard cutover for a live integration — worth documenting, or supporting two concurrent secrets like most providers do. And the UI ships a first-class **Reactivate** button (`useReactivateOAuthClient`), which makes the resurrect-a-deactivated-client-and-all-its-tokens path from my pass-4 body note a one-click operator action rather than a theoretical API call — that raises its priority a little.

src/jentic_one/admin/services/oauth_client_service.py
```diff
@@ -0,0 +1,308 @@
...
+        if client is None or not client.active:
+            return False
+        return redirect_uri in client.redirect_uris
+
+    async def verify_client_secret(self, client_id: str, client_secret: str) -> bool:
+        """Verify a client's secret. Returns False if not found, inactive, or wrong secret."""
+        async with self._ctx.admin_db.session() as session:
+            client = await OAuthClientRepository.get_by_client_id(session, client_id)
+        if client is None or not client.active:
+            await _verify_password_async(client_secret, _DUMMY_ARGON2_HASH)
```
Manuel-Jentic: **This makes `/oauth/token` an unauthenticated memory-and-CPU amplifier.** I measured the configured hasher in this tree: argon2id `time_cost=3`, `memory_cost=65536 KiB` (**64 MiB**), `parallelism=4`, ~**25 ms** per verify. Two properties combine badly here. (1) The dummy-hash branch means an unknown or inactive `client_id` still costs a **full** argon2 verify — that's deliberate for timing, and correct as far as it goes, but it also removes the cheap early-out, so garbage input is as expensive as real input. (2) In `oauth.py:146-151` this runs **before** `exchange_code`, so no valid authorization code, PKCE verifier, or redirect_uri is needed to reach it — `grant_type=authorization_code` with four junk strings is enough. The only thing in front is the IP-keyed token limiter, which pass 3 established is bypassable with a rotating `X-Forwarded-For`. Since `_verify_password_async` dispatches to the **default** executor (`run_in_executor(None, …)`), concurrency is `min(32, cpu+4)` — 16 on this box — so a trivial request stream pins ~1 GiB of transient RAM and, because argon2 releases the GIL, genuinely saturates cores; it also shares that executor with every other `run_in_executor` user in the process, so unrelated work stalls. Suggestions, cheapest first: gate the argon2 call behind the cheap checks (verify the code exists and is unconsumed before spending 64 MiB), give this verification a dedicated bounded executor so it can't starve the default pool, and consider lower-memory parameters for machine-to-machine secrets — a 256-bit `secrets.token_urlsafe(32)` client secret has no dictionary risk, so heavy memory-hard KDF parameters buy far less here than they do for human passwords.

cli/tests/arch/spec_test.go
```diff
@@ -187,6 +187,7 @@ var sensitiveSweepAllowlist = map[string]bool{
 	"has_api_key":          true, // presence flag, not the key
 	"must_change_password": true, // policy boolean
 	"clear_session_token":  true, // "clear the token?" boolean directive
+	"consent_token":        true, // HMAC-signed state token for OAuth consent flow, not a credential
```
Manuel-Jentic: **This allowlist entry is incorrect, and it defeats the redaction this test exists to enforce.** The consent token *is* a credential: it's a signed bearer blob carrying `user_id`, `user_email`, and the requested scopes, and possession alone is sufficient to approve consent as that user — the GET handler doesn't consume the nonce (only the POST does, `authorize.py:723`), so a captured token is replayable for its full 5-minute TTL with no session or cookie required. It's the sole authenticator on `POST /oauth/consent`. The allowlist's own doc comment sets the bar: "A new secret-shaped field must be annotated in the backend model — **never added here unless it truly carries no secret value**", and the reviewed exemptions above it are all booleans (`has_api_key`, `must_change_password`, `clear_session_token`). Consequence: `consent_token` is absent from `sensitive.gen.go` (grepped — no match), so `ux.IsSensitiveKey` won't redact it in CLI output or debug dumps, and it appears in the spec both as a `Body_consentSubmit` property and as a **query parameter** on `GET /oauth/consent` — which is the same value my pass-4 inline flags as landing in access logs. Two independent leak channels now. The fix is the one the comment prescribes and that this PR already applied correctly to `client_secret`: annotate the field at the Pydantic/Form source with `json_schema_extra=SENSITIVE` and drop this line.

src/jentic_one/admin/services/oauth_client_service.py
```diff
@@ -0,0 +1,308 @@
...
+from jentic_one.shared.audit import record_audit
+from jentic_one.shared.auth.identity import Identity
+from jentic_one.shared.context import Context
+from jentic_one.shared.models.audit import AuditAction, AuditTargetType
+
+_MAX_REDIRECT_URIS = 20
+_MAX_REDIRECT_URI_LENGTH = 2048
+
+_DUMMY_ARGON2_HASH = hash_password("dummy-timing-equalizer")
+
```
Manuel-Jentic, line 28: Import-time side effect with a real price tag: this computes an argon2id hash **at module import**, which per the parameters measured above is ~25 ms and a 64 MiB allocation before anything is served. It's paid by every process that imports the admin service layer — the API server at boot, each pytest-xdist worker at collection, and any short-lived CLI/script import — and it can't be avoided even by callers that never touch OAuth clients. Making it a lazily-initialised cached value (`functools.cache` on a small accessor, or computing it on the first `verify_client_secret` miss) keeps the timing-equalizer property while taking it off the import path.

Manuel-Jentic (commented):
Pass-6 review at head `667b71b5` (unchanged; CI green; still no replies on the earlier threads). This lap took two angles nothing before it had looked at: **the cost of the cryptography on the unauthenticated request path**, and **the CLI/spec redaction plumbing** the PR quietly adjusts.

The significant one is a resource-exhaustion primitive. Verifying a client secret is an argon2id operation, and I measured the configured parameters in this tree: `time_cost=3`, `memory_cost=65536 KiB` (**64 MiB**), `parallelism=4`, ~25 ms per verify (argon2-cffi 25.1.0, defaults). Because the check runs **before** the authorization code is validated and the dummy-hash equalizer guarantees a full verify even for a `client_id` that doesn't exist, any unauthenticated caller can convert a few hundred bytes of request into 64 MiB + 25 ms of server work, with only the IP-keyed limiter in front of it — the same limiter I reported as `X-Forwarded-For`-spoofable in pass 3. That's the part worth re-weighing: I originally framed the XFF bypass as "unlimited pre-auth attempts", but combined with this it's a memory-exhaustion lever (the default executor allows 16 concurrent verifies here → ~1 GiB transient, and argon2 releases the GIL so it saturates cores too). Fixing either one defuses it; fixing both is better.

The second is smaller but pointed: the PR adds `consent_token` to the CLI's sensitive-annotation allowlist labelled "not a credential", which is the opposite of what it is.

**Worth crediting, because I went looking:** `client_secret` is annotated correctly with `json_schema_extra=SENSITIVE` on both the create and rotate responses, and that propagates properly into `sensitive.gen.go` — so the mechanism was understood and used correctly, which is exactly why the `consent_token` allowlist entry stands out. I also checked whether the missing CLI commands for this resource were a gap and they're **not** — `cli/internal/cli/ctl/` only carries config and overlay, so no admin resource has CLI CRUD and OAuth clients are consistent with the rest. And the `verify_token` install ordering in `app_factory.py:601-608` is handled properly.

**One nit not worth an inline:** `_validate_redirect_uris` permits `http` only for `localhost` and `127.0.0.1`, so the IPv6 loopback form `http://[::1]:port/cb` is rejected — RFC 8252 §7.3 asks native-app clients to support both loopback forms, and a dev on an IPv6-first stack will hit this.

**One observation that strengthens an earlier finding rather than adding a new one:** `app_factory.py:601-608` shows the codebase already learned that surface-install side effects are too fragile to rely on ordering — it re-installs the auth verifier after the surface loop explicitly, precisely so `at_` resolution can't depend on which surface ran first. The new `auth_state_backend` set in the same `install_on_app` got no equivalent guarantee, which is the concrete mechanism behind the memory-fallback caching I flagged in passes 3 and 4.

Manuel-Jentic (commented):
Pass-7 review at head `667b71b5` (unchanged; CI green; no replies yet). Three seams left unmined by passes 1-6: the **IdP-federation entry point** the consent flow added, the **new settings page read end-to-end** rather than spot-checked, and how the new client type relates to the **existing `/register` client registry**.

The most useful finding is a duplication-and-drift one in the UI. The scope picker hardcodes a 14-entry `COMMON_SCOPES` array, but `permission_catalog.py` defines **25** permissions — so 11 are unreachable through the picker and have to be typed by hand into a free-text field that validates nothing. This codebase already solved that exact problem: `usePermissionCatalogue()` in `ui/src/modules/agents/api/hooks.ts:632-639` fetches the live catalogue from `GET /permissions` and, per its own doc comment, already feeds "the Scopes editor … scope list" with `grantableByCaller` gating. The new page reimplements a worse version of a picker that already exists.

Second, worth knowing before this ships: because `resolve_idp_user` provisions accounts, **any registered third-party client can drive local account creation before the user has consented to anything**, and a subsequent "Deny" leaves the account and identity link in place.

**On `/register` — a follow-on to my pass-4 discovery comment rather than a new issue.** I checked whether the new client type could collide with the existing registry and it can't (agent clients live in a separate table and an agent `client_id` is rejected at `/authorize`), which is good. But the two registries are now disjoint in a way the metadata doesn't express: `/register` is genuine self-service RFC 7591 issuing **agent** clients (JWKS + `private_key_jwt`), while this PR's third-party clients are admin-only and use `client_secret`. So a third-party developer who follows `registration_endpoint` from the discovery document will successfully self-register a client that **cannot** perform the authorization-code flow this PR adds, with nothing explaining why. That's the same root cause as the `token_endpoint_auth_methods_supported` gap I flagged earlier — the metadata describes only the agent client type — and it argues for fixing the discovery document as one job.

**Credit where due:** `_resolve_or_create_user` itself is genuinely careful work — it fails closed when an unverified email matches an existing account rather than linking or duplicating, handles the concurrent-callback race via the `(provider, external_subject)` unique constraint with a retry-as-lookup, consults the admission policy for brand-new emails, and audits the reject path deliberately outside the rolled-back transaction. I went looking for a takeover primitive there and there isn't one. The list/empty/loading/error states on the new page are all handled, the deactivate flow has a confirmation dialog with inline error display, `CopyButton` is used for the client ID, and the "Show inactive" checkbox is correctly associated with its `Label htmlFor`.

ui/src/modules/settings/pages/OAuthClientsSection.tsx
```diff
@@ -0,0 +1,726 @@
...
+			))}
+			<Button type="button" variant="outline" size="sm" onClick={handleAdd}>
+				<Plus className="mr-1 h-3 w-3" />
+				Add URI
+			</Button>
+		</div>
+	);
+}
+
+const COMMON_SCOPES = [
```
Manuel-Jentic: **This hardcodes a copy of the permission catalogue that has already drifted, and the escape hatch validates nothing.** `permission_catalog.py:33-56` defines 25 permissions; this array lists 14. Missing, and therefore not selectable: `catalog:import`, `config:read`, `config:write`, `events:read`, `events:write`, `jobs:write`, `org:admin`, `overlays:confirm`, `service-accounts:read`, `service-accounts:write`, `users:write`. (Nothing here is invented, at least — every entry is a real permission.) Two consequences. First, restricting a client to any of those 11 requires typing it into the custom field, and `addCustom` (line 144) accepts **any** non-empty trimmed string — no catalogue check, no format validation — while the backend doesn't validate scope strings either. Second, that's not a harmless typo: an unrecognised entry in `allowed_scopes` simply never matches at `resolve_access_token` (`token_service.py:420-421`), so `capabilties:read` silently narrows the client's ceiling and drops that permission from every token it holds, with no error anywhere. Combined with `[]` meaning deny-all, the whole field fails silently in both directions. The fix already exists in this codebase: `usePermissionCatalogue()` (`ui/src/modules/agents/api/hooks.ts:632-639`) fetches the live catalogue from `GET /permissions` with a 5-minute `staleTime`, and its docstring says it already backs a Scopes editor picker including `grantableByCaller` gating. Reuse it and validate the custom input against it.

ui/src/modules/settings/pages/OAuthClientsSection.tsx
```diff
@@ -0,0 +1,726 @@
...
+											onClick={(): void => setRotateTarget(client)}
+											title="Rotate secret"
+										>
+											<ShieldAlert className="h-4 w-4" />
+										</Button>
+									)}
+									<Button
+										variant="ghost"
+										size="sm"
+										onClick={(): void => setEditClient(client)}
```
Manuel-Jentic, line 565: Icon-only buttons with no accessible name: this Edit (`Pencil`) button, the Delete (`Trash2`, line 583-589) and Reactivate (`RotateCcw`, line 591-598) buttons all render a bare icon with no `aria-label` and no `title`, so a screen reader announces them as unlabelled buttons — and in a list of clients there's no way to tell which row's button has focus. The Rotate button immediately above (line 566-573) *does* carry `title="Rotate secret"`, so the inconsistency is visible within a single JSX block. Same issue on the scope-removal `X` button (line 183-189). Adding `aria-label={`Edit ${client.name}`}` (and equivalents) fixes it and gives each row's controls distinct names. Worth noting the destructive one is the least discoverable: the Delete button is the one that needs the clearest label.

src/jentic_one/auth/services/authorize_service.py
```diff
@@ -112,14 +139,35 @@ async def handle_idp_callback(
...
+        Returns (user_id, user_email) without minting an authorization code.
+        Used by the consent flow to defer code minting until after approval.
+        """
+        adapter = self._get_idp_adapter()
+        if adapter is None:
+            raise InvalidGrantError("No external IdP configured")
+
+        userinfo = await adapter.exchange_code(code, redirect_uri=redirect_uri)
+        claims = adapter.map_claims(userinfo)
+        user_id = await self._resolve_or_create_user(claims)
```
Manuel-Jentic, line 169: **Provisioning now happens before consent, on a third-party client's behalf.** `_resolve_or_create_user` doesn't just resolve — for a brand-new admissible email it creates the `User`, the `ExternalIdentity` link, and applies `get_default_idp_grants()` baseline permissions (lines 370-402). This new helper is called from the consent branch of `/oauth/callback` (`authorize.py:549`), i.e. **before** the consent screen is rendered, so any registered third-party client can drive local account creation in the org simply by sending a user through its authorize URL — and if the user then clicks **Deny**, the account, the identity link, and the default grants all persist with nothing to indicate the user refused. The audit records `created_by="self"` and `actor_id=new_user.id`, so the log reads as a self-service signup with no trace of which third-party client initiated it (related to the missing consent audit I flagged in pass 4). Deferring creation until after approval would be the clean fix — the consent token already carries everything needed to create the user on the POST path. If provisioning-before-consent is intentional, it should at least be audited with the initiating `client_id` so an admin can tell org growth from third-party probing.

src/jentic_one/admin/services/oauth_client_service.py
```diff
@@ -0,0 +1,308 @@
...
+
+def _validate_redirect_uris(uris: list[str]) -> None:
+    """Validate that all redirect URIs are well-formed HTTPS URLs (http only for localhost)."""
+    if not uris:
+        raise InvalidInputError("at least one redirect_uri is required")
```
Manuel-Jentic, line 33: `_validate_redirect_uris` checks count, length, scheme, host, and fragment, but never **deduplicates** — `["https://a/cb", "https://a/cb"]` is accepted and stored as-is. That's mostly cosmetic server-side (the membership tests in `is_redirect_uri_allowed` still work), but it breaks the UI: the client card renders `client.redirect_uris.map((uri) => <li key={uri}>)` (`OAuthClientsSection.tsx:617-621`), so duplicates produce duplicate React keys and the console warning/reconciliation glitch that comes with them. A `dict.fromkeys()` pass here — before the count check, so dedupe can't be used to smuggle past `_MAX_REDIRECT_URIS` — fixes both ends, and also makes the 20-URI limit mean 20 *distinct* URIs.

PR #1178, opened 2026-08-28
feat(telemetry): add Origin.MCP + MCP session/config events

### LINKED ISSUE #1177: local MCP phase 2 (lane D): Origin.MCP + execution-event origin tag + mcp_session_started/mcp_config_registered events
## Context

Part of the local-MCP theme ([plan PR jentic-one-plans#25](https://github.com/jentic/jentic-one-plans/pull/25); phase doc: [`themes/local-mcp/phase-2.md`](https://github.com/jentic/jentic-one-plans/blob/docs/local-mcp-plan/themes/local-mcp/phase-2.md), lane D). **Unblocked — no dependency on the Go lane** (the Go side sends the `jentic-mcp/` User-Agent later). Follow the two-plane pattern: property-free telemetry wire (closed-enum tags only), full-fidelity detail on internal events, which is what the UI reads.

## Work

- **`Origin.MCP`**: new member on the `Origin` enum (`shared/models/actors.py`) + a `jentic-mcp/` branch in `derive_origin` (`shared/web/deps.py`). That alone labels every MCP execution — `execution_records.origin` is already persisted and on the admin wire. Note the broker's own resolver derives origin from `User-Agent` in `broker/web/deps.py` too.
- **Execution-event origin tag**: `EVENT_TAGS` is single-valued today and `EXECUTION_FAILED` already maps `ErrorSource` — widen values to tuples, map `EXECUTION_FAILED: (ErrorSource, Origin)` and `EXECUTION_COMPLETED: Origin`, add `Origin` to the `EventTag` union. Test strategy: tag-validation regression tests over every existing tagged event + one `event_tag_dropped` case pinning the fallback. Fallback if the widening is rejected in review: a separate `mcp_execution` allowlist event.
- **`mcp_session_started`**: new `EventType` constant (+ `ALL` frozenset) + `TelemetryEventName` member. Telemetry side: at most a closed `McpClient` tag (`claude|cursor|codex|other`). Internal event `data`: full clientInfo name/version, `transport` (`stdio|http`), and the `session_id` UUID (the dedupe key); emitted **with `actor_id` = the agent** (mandatory here — the UI's per-agent read keys on it). Emit once per session UUID on the first authenticated request carrying `X-Jentic-Session-Id` **with a `jentic-mcp/` User-Agent** (the UA, not the header, marks the session as MCP — plain CLI calls carry the header too). Dedupe: one indexed events-table lookup on `data.session_id` before emitting, then a per-process seen-set; **both planes emit** (control plane and broker — broker-only sessions must still appear), sharing the table-backed dedupe. Decide the exact mechanics in the PR within these constraints.
- **`mcp_config_registered`** (per runtime, closed runtime tag): the CLI can't call `emit_event` — decide in this PR between a small authenticated report-registration endpoint (per-runtime fidelity; the 2-E3 CLI side would call it) or a server-side emit from an existing bootstrap call. Enum entries can land first.
- Everything rides `emit_event`'s existing consent gate — no MCP-specific consent plumbing.
- **Not in scope**: kill-switch hardening — #1136 was fixed by #1137; nothing remains here.

## Acceptance

- [ ] MCP executions carry `origin=mcp` on execution records and the admin wire
- [ ] Execution telemetry split by origin (or the fallback `mcp_execution` event)
- [ ] `mcp_session_started` emitted once per session UUID with `actor_id`, clientInfo, transport, session_id in `data`; dedupe correct across workers and both planes
- [ ] `mcp_config_registered` emit path decided and enum entries landed
- [ ] Tag-validation regressions green; telemetry wire stays property-free

Manuel-Jentic:
## Summary

Lane D of the local-MCP theme (`themes/local-mcp/phase-2.md`): telemetry & events groundwork for the upcoming MCP transport.

- **`Origin.MCP`**: new enum member; `derive_origin` (shared) and the broker's origin derivation now map a `jentic-mcp/` User-Agent prefix to it. `execution_records.origin` already persists the value — no schema work.
- **Execution-event origin tag**: `EVENT_TAGS` values widened to tuples of allowed tag types (`_validate_tags`'s isinstance loop already handles tuples). `EXECUTION_FAILED` now allows `(ErrorSource, Origin)`, `EXECUTION_COMPLETED` allows `(Origin,)`; `Origin` added to the `EventTag` union. All three execution emit sites (broker sync path, broker streaming persist, async worker handler) tag lifecycle events with the request-derived origin via a new `origin_or_none` coercion helper (garbage/absent origins degrade to no tag, never an error).
- **`mcp.session_started`**: new `EventType` + `TelemetryEventName` + closed `McpClient` tag enum (`claude|cursor|codex|other`). Emitted once per session UUID on the first authenticated request carrying both the `X-Jentic-Session-Id` header **and** a `jentic-mcp/` User-Agent (the UA marks it MCP; plain CLI calls carry the header too). Internal event `data`: full parsed clientInfo (name/version from `jentic-mcp/<version> (<client>/<clientversion>)`, parenthetical optional → client `unknown`), `transport: "stdio"`, `session_id`; `actor_id`/`actor_type` = the authenticated agent. Telemetry wire carries only the closed `McpClient` tag.
- **Dedupe**: per-process bounded seen-set short-circuit backed by one indexed events-table lookup (`EventRepository.exists_with_data_value`, riding `ix_events_type_created` on the `type` predicate). Both planes emit (shared `resolve_identity` and broker `require_broker_identity`) and share the table-backed dedupe, so execute-only sessions are still counted exactly once.
- **`mcp.config_registered`**: `EventType` + `TelemetryEventName` + closed `McpConfigRuntime` tag enum (`claude_code|cursor|codex|generic`) landed.

### Emit-path decision for `mcp.config_registered` (reviewer input wanted)

The plan leaves open: a small authenticated report-registration endpoint the CLI calls (per-runtime fidelity) vs a server-side emit from an existing bootstrap call. **This PR lands only the enum/event entries and defers the emit wiring.** Recommendation: a dedicated `POST /events/mcp-config-registered`-style authenticated endpoint is the cleaner fit — config registration happens client-side in the CLI (`jentic mcp install`-type flows), so only the CLI knows which runtime's config file was actually written; a server-side emit from bootstrap would have to guess the runtime and would fire on re-bootstraps rather than registrations. That endpoint touches the Go CLI (out of scope for this lane) and deserves its own small PR alongside the CLI change, per the `web-layer` rule's one-concern-per-endpoint guidance.

### Consent & wire discipline

All new events ride `emit_event`'s existing consent gate — no MCP-specific consent plumbing. Telemetry wire stays property-free: closed enums only, raw client strings live only in the internal event `data`.

### Tests

- Tag-validation regression suite over every (event, allowed-tag-type) pair in the widened `EVENT_TAGS` map + an `event_tag_dropped` case pinning the fallback.
- Origin derivation (shared + broker), `origin_or_none` coercion, execution-lifecycle tagging in broker service and async worker.
- MCP UA parsing, session-ID validation, seen-set + table-backed dedupe, both-plane emit wiring.
- Integration test for `EventRepository.exists_with_data_value` (SQLite + Postgres backends).

`uv run pytest tests/unit tests/arch`, integration suite (both backends), `make lint`, `make detect-secrets`, mypy: all green. (`make check`'s `score` target needs a `JENTIC_API_KEY` not available locally; no OpenAPI surface changed.)

Closes #1177

### DIFF
--- a/src/jentic_one/admin/core/schema/events.py
+++ b/src/jentic_one/admin/core/schema/events.py
@@ -32,6 +32,22 @@ class Event(AuditableMixin, AdminBase):
             postgresql_where=text("trace_id IS NOT NULL"),
         ),
         Index("ix_events_actor", "actor_id", "actor_type"),
+        # Cross-worker exactly-once backstop for mcp.session_started: the emit
+        # path (shared/events/mcp_session.py) is check-then-insert, so two
+        # workers can both pass the dedupe lookup before either commits. This
+        # partial unique index turns the loser's insert into an IntegrityError
+        # (tolerated there), and doubles as the read-path index for
+        # EventRepository.exists_with_data_value on Postgres. Scoped to the one
+        # event type so other events carrying a session_id stay unconstrained.
+        # The SQLite migration (f2a3b4c5d6e7) creates the same index with
+        # json_extract(data, '$.session_id') instead of the Postgres operator.
+        Index(
+            "uq_events_mcp_session_started_session",
+            "type",
+            text("(data ->> 'session_id')"),
+            unique=True,
+            postgresql_where=text("type = 'mcp.session_started'"),
+        ),
     )
 
     id: Mapped[str] = mapped_column(
--- a/src/jentic_one/admin/repos/event_repo.py
+++ b/src/jentic_one/admin/repos/event_repo.py
@@ -66,8 +66,15 @@ class EventRepository:
 
         One narrow lookup used for table-backed dedupe (e.g. one
         ``mcp.session_started`` per ``data.session_id`` across workers and
-        surfaces). The ``type`` predicate rides the ``ix_events_type_created``
-        index; the JSON comparison then filters the (small) per-type slice.
+        surfaces — a pair additionally *enforced* by the partial unique index
+        ``uq_events_mcp_session_started_session``; this lookup is the cheap
+        first check, the index is the guarantee). On Postgres that same index
+        covers the ``(type, data->>'session_id')`` read here. For other
+        type/key pairs the ``type`` predicate rides ``ix_events_type_created``
+        and the JSON comparison filters the (small) per-type slice. On SQLite
+        the JSON comparison compiles to a ``json_extract`` path form that does
+        not match the index expression, so the lookup stays a per-type scan —
+        acceptable for the embedded single-file target.
         """
         stmt = (
             select(Event.id)
--- a/src/jentic_one/admin/repos/event_repo.py
+++ b/src/jentic_one/admin/repos/event_repo.py
@@ -55,6 +55,30 @@ class EventRepository:
         return await session.get(Event, event_id)
 
     @staticmethod
+    async def exists_with_data_value(
+        session: AsyncSession,
+        *,
+        event_type: str,
+        key: str,
+        value: str,
+    ) -> bool:
+        """Return whether any event of ``event_type`` has ``data[key] == value``.
+
+        One narrow lookup used for table-backed dedupe (e.g. one
+        ``mcp.session_started`` per ``data.session_id`` across workers and
+        surfaces). The ``type`` predicate rides the ``ix_events_type_created``
+        index; the JSON comparison then filters the (small) per-type slice.
+        """
+        stmt = (
+            select(Event.id)
+            .where(Event.type == event_type)
+            .where(Event.data[key].as_string() == value)
+            .limit(1)
+        )
+        result = await session.execute(stmt)
+        return result.scalar_one_or_none() is not None
+
+    @staticmethod
     async def list_all(
         session: AsyncSession,
         *,
--- a/src/jentic_one/broker/services/execution/service.py
+++ b/src/jentic_one/broker/services/execution/service.py
@@ -35,6 +35,7 @@ from jentic_one.shared.events.repeated_failure import maybe_emit_repeated_failur
 from jentic_one.shared.executions import record_execution
 from jentic_one.shared.metrics import get_meter
 from jentic_one.shared.models import ExecutionStatus
+from jentic_one.shared.models.actors import origin_or_none
 from jentic_one.shared.models.events import ErrorSource, EventSeverity, EventTag, EventType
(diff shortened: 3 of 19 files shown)


Manuel-Jentic:

An independent reviewer agent reviewed this PR (jointly with #1190, which stacks on it). Verdict here was **request-changes** (1 MAJOR — a live-reproduced duplicate-emit race — plus 2 MINOR); all findings addressed in 8142a689. Record below.

### What held under attack
- Two-plane integrity: telemetry wire stays property-free — only `(wire_name, valid_tags, actor_type)` leaves; raw clientInfo never reaches the sink; unknown clients collapse to `OTHER`.
- `EVENT_TAGS` tuple widening complete with all pre-existing entries migrated; the dropped-tag fallback pinned.
- Origin threading is retry-safe (persisted in the job payload at enqueue); both planes hook the emit; failures never fail the request; the non-MCP fast path is negligible.

### Findings → fixes (8142a689)
|---|---|---|---|
| 1 | MAJOR | Dedupe was check-then-insert with no in-flight tracking and no DB uniqueness — two concurrent first requests for one session id emitted twice (**reproduced live**); exactly the orchestrator-set `$JENTIC_SESSION_ID` scenario the plan designed for | Both layers: synchronous `_in_flight` set before `create_task` (discarded in done callback so failed emits retry), plus a partial unique index `(type, data->>'session_id') WHERE type='mcp.session_started'` on both dialects with the losing insert treated as already-emitted — true exactly-once; concurrency + lost-race tests |
| 2 | MINOR | Untrusted UA fields persisted unbounded to the events table | 128-char cap on `server_version`/`client_name`/`client_version`; oversized-UA test |
| 3 | MINOR | `exists_with_data_value` degraded to a growing per-type scan | The unique index doubles as the read index on Postgres (one index, not two); deliberate SQLite scan documented |

Migration `f2a3b4c5d6e7` written per the house dialect-branch style, cycled up→down→up clean. 133 unit + 13 integration (SQLite) + 283 arch tests, mypy, ruff all green at 8142a689 (Postgres coverage via offline-rendered DDL; Docker daemon unavailable locally).

PR #1179, opened 2026-08-28
refactor(cli): extract agentops from cmd for reuse

Manuel-Jentic:
## Summary

Extracts the UX-free execute core out of the cobra command tree into `cli/internal/agentops` (local-MCP theme, phase-0 item 0.2; plan: `jentic-one-plans/themes/local-mcp/phase-0.md`). **Pure refactor — no behavior change**; part of #1174, unblocks #1175.

## Changes

- `cli` — new **`cli/internal/agentops/`**, the reusable core a future MCP handler can call without cobra:
  - `types.go`: `ExecuteRequest` / `ExecuteResult` (+ `Envelope()`), `Denial`, `Operation`, `KV`, the narrow **`Inspector`** seam (`Inspect(ctx, target, revision, format) ([]byte, error)` — satisfied by the existing generated-SDK wrapper, shared by execute-resolve and `jentic inspect`), and `HTTPError` (moved from `api`, aliased there so every existing call site is untouched).
  - `execute.go`: `ParseMethodPath` / `ResolveOperation` (de-cobra'd — takes `ctx` + `Inspector`), `BuildRequest` (path-param substitution, query append, broker catch-all URL, SEC-1 secure-transport guard, bearer/correlation/idempotency headers), `Do` (SDK broker transport, bounded read), and one-call `Execute`.
  - `classify.go`: `IsBrokerDenial` / `ParseAgentDirective` lifted intact; `Classify` unfuses the old `executeOutput`'s classification from printing.
- `cli` — **`cli/internal/cli/ux/execute.go`**: the shared success envelope (`ux.ExecuteEnvelope`, replacing the inline map; fields declared alphabetically so the emitted bytes stay frozen) and the directive / synthesized-denial renderers (command code no longer composes stderr text; the app keeps only the stream choice).
- Dependency rules held: `agentops` receives token strings and a **resolved** broker target. Session acquisition, broker precedence + SEC-21 machine-mode pin + remote-CP fail-closed guard, the stdin fallback, `--data`/`--raw`, rendering, and exit-code mapping all stay cobra-side. No `io.Writer`, no cobra, no exit codes, no TTY, no stdin in `agentops`.
- **Out of scope**: no MCP SDK dependency (phase-1 PR 1-A), no behavior change of any kind — the goldens exist to prove that.

## Risk & rollback

[...]

Issue #1175, opened 2026-08-28
local MCP phase 1 (lane A): jentic mcp stdio server — skeleton, discovery/execute tools, hardening, skills resources, access-loop tail


Manuel-Jentic:
## Context

Part of the local-MCP theme ([plan PR jentic-one-plans#25](https://github.com/jentic/jentic-one-plans/pull/25); phase doc: [`themes/local-mcp/phase-1.md`](https://github.com/jentic/jentic-one-plans/blob/docs/local-mcp-plan/themes/local-mcp/phase-1.md)). Lane A is the serial Go spine: **PR 1-A → 1-B → 1-C → 1-D**, plus the access-loop tail (2-E1) which queues behind 1-C. Blocked on the phase-0 `agentops` extraction; product sign-off (phase-0 0.1) is **not** a gate for this lane.

## PR 1-A — command skeleton + pre-auth surface

- New cobra command `jentic mcp` registered via `cmdcore.AddGrouped(root, "client", …)` (gets `--context`/`$JENTIC_CONTEXT`, state injection, fencing, migrate gate for free). Default stdio via `github.com/modelcontextprotocol/go-sdk` v1.7.x (dep lands here). Machine-safe: never prompt.
- Flags: `--read-only`, `--exclude-tools`, `--log-file` (default under XDG state dir — stdout is the wire).
- **Always boots**: `tools/list` works with no/invalid config. Pre-auth tools: `get_started` (diagnoses no-config / pending approval / instance unreachable, names the fix verbatim; never auto-starts the instance) and `whoami`.
- Identity-stamp plumbing: TTL-cached `GET /instance` (`GetInstanceWithResponse` exists uncalled) with the degraded `backend: "unreachable"` form.
- `User-Agent: jentic-mcp/<version>` on every backend call, upgraded with clientInfo from request `_meta` when present; session UUID via the shipped `X-Jentic-Session-Id` seam (orchestrator-set `$JENTIC_SESSION_ID` wins; per-process UUID fallback). This needs a new exported `clictx` transport hook composing the UA/session RoundTripper **over** the CA-pinned client (wrap, never displace; `RequestEditorFn` chain is the fallback if rejected in review).

## PR 1-B — discovery tools

- `search_apis` → `SearchOperationsWithResponse`, envelope passthrough.
- `inspect_operation` → the `agentops` raw-inspect wrapper (full body, `json` format, redaction kept).
[...]

PR #1181, opened 2026-08-28
feat(cli): add jentic mcp stdio server skeleton with pre-auth tools

Manuel-Jentic:
> **Stacked on #1179; retarget to main after it merges. Part of #1175.**

Implements local-MCP phase 1, **PR 1-A**: the `jentic mcp` command skeleton and pre-auth surface.

## What

- **`jentic mcp`** — an MCP stdio server (official Go SDK `modelcontextprotocol/go-sdk` v1.7.0, dep added here per phase-0 item 0.3), registered via `cmdcore.AddGrouped(root, "client", …)` so it inherits `--context`/`$JENTIC_CONTEXT`, the interceptor's state injection, fencing, and the migrate gate. Machine-safe: never prompts; stdout is the JSON-RPC wire (banner skipped, logs to `--log-file`, default `<XDG state>/jentic/logs/mcp.log`). Flags: `--read-only`, `--exclude-tools`, `--log-file`. Marked long-running so the interceptor's 60s agent deadline doesn't kill the session; each tool call bounds its own context instead.
- **Always boots**: `tools/list` (and the SDK's legacy-`initialize` fallback) works with no/invalid config. Exactly two tools:
  - `get_started` — pre-auth diagnosis over a closed state enum (`no_config` / `not_registered` / `pending_approval` / `instance_unreachable` / `ready`), each carrying the operator instruction verbatim from the V2 no-config error and the skill's step-1 branches; the same-host unreachable branch names `jenticctl start` and the server never auto-starts the instance.
  - `whoami` — the existing GET /me agent surface, envelope passthrough.
- **Identity stamp**: TTL-cached (60s) `GET /instance` via `clictx.GetControlClient`; every tool result gets a top-level sibling `instance` key `{backend | "unreachable", host, instance_id, fetched_at}` — never a wrapper, never `_meta`. Degraded form keeps the last-known identity with the *real* last-success `fetched_at`. `get_started`'s reachability probe forces a refresh through the same cache, so diagnosis and stamp can't disagree.
- **`clictx` transport hook** (new seam): `clictx.WithTransportHook` composes a caller RoundTripper over the transport the SEC-20 resolution produced — **wrap, never displace** the CA pinning; the SDK's retry transport stays outermost so headers ride every attempt. The MCP server's hook sets `User-Agent: jentic-mcp/<version>`, upgraded to `… (<client>/<clientversion>)` once a request's `_meta` carries `io.modelcontextprotocol/clientInfo`, and stamps a per-process session UUID into `X-Jentic-Session-Id` **only when absent** — the SDK's session editor attaches an env-set `$JENTIC_SESSION_ID` first, so the orchestrator always wins.

## Hook shape rationale

The context-carried `TransportHook` was chosen over the generated client's `RequestEditorFn` chain: the hook decorates the *transport*, so the headers survive the SDK's retry loop on every attempt and cover any client the clictx constructors build (including 1-C's broker leg via the same seam), whereas editors run once per logical request and would need threading through every constructor signature. The hook receives the already-resolved base transport, making "wrap, never displace" structural rather than conventional.

## Tests

- `get_started` branch selection (pure `diagnoseSetup` table, incl. loopback-vs-remote unreachable wording and the no-auto-start decision).
- Instance stamp: fresh, TTL expiry, degraded-with-history (real `fetched_at`), degraded-with-no-history, probe-forces-refresh.
- Attribution RoundTripper: UA base + clientInfo upgrade + sanitization; session precedence (pre-set header wins); request non-mutation.
- In-process MCP session (SDK client, in-memory transport): `tools/list` and `get_started` on a machine with no config; `--exclude-tools` filtering.
- `make -C cli check` green (lint, vet, race tests, arch tests, vendored specs).


Manuel-Jentic:

An independent reviewer agent attacked this PR's core safety properties. Verdict was **request-changes** (3 MAJOR, 5 MINOR); all findings addressed in 143f255c. Record below.

### What held under attack
- stdout purity is structurally sound — the only stdout writer in the serving path is the SDK's stdio transport (banner double-gated, slog on stderr, SDK logging on the file handler, panics to stderr).
- The transport hook wraps and never displaces (CA pinning stays innermost; attribution re-stamps every retry, including the 401 re-exchange arm).
- `req.Clone` before stamping; orchestrator-wins session-id precedence is real; `uaToken` sanitization closes header smuggling; pre-auth surface green under `-race`; both fencing exemptions narrow and justified.

### Findings → fixes (143f255c)
|---|---|---|---|
| 1 | MAJOR | `instanceCache` held its mutex across network I/O — a hung control plane head-of-line blocks every concurrent tool call (~60 s worst case), plus double-dial per `get_started` and no negative caching | Fetch moved outside the lock with singleflight; waiters honor their own deadlines; 5 s negative cache kills the double dial and per-call re-dials; 4 concurrency tests under `-race` |
| 2 | MAJOR | Revoked/kill-switched identity (wire 401 surviving the re-exchange) surfaced as `INTERNAL_ERROR` with no `next_tool` — violating the §3.7 row written for exactly this case | `mcpCoded()` maps `*HTTPError` 401/403 → `NOT_AUTHENTICATED` with the `get_started` pointer; session-level test |
| 3 | MAJOR | The load-bearing SDK behavior (context values reaching handlers) and the concurrency/annotation contracts were untested — an SDK upgrade could silently regress every tool to `no_config` with CI green | End-to-end session test pinning context propagation + attribution headers; concurrency tests; `LongRunningAnnotation` presence + interceptor-honors tests |
| 4 | MINOR | Degraded stamp claimed `backend: "unreachable"` for answered HTTP errors; §3.7.4 refresh-on-auth-error absent | `backendError` marker distinguishes answered-with-error (last-known identity kept); `softError` invalidates the cache on auth codes |
| 5 | MINOR | Marshal-failure fallback could emit invalid JSON | Fallback is itself a `json.Marshal` of a string map |
| 6 | MINOR | Config read error misdiagnosed as "not registered" (comment claimed the opposite of the code) | `configErr` carried into the probe; dedicated diagnosis branch surfaces the cause |
| 7 | MINOR | Token-mint calls bypass the attribution hook and CA pinning (predates this PR) | Recorded as a known gap at the hook site; follow-up issue to be filed |
| 8 | MINOR | Windows stance ("confirm and document in 1-A") undelivered | `cli/README.md` MCP section states the decided scope |

`GOWORK=off make -C cli check` fully green at 143f255c (lint, vet, race).

PR #1183, opened 2026-08-28
feat(cli): add search_apis + inspect_operation MCP discovery tools

Manuel-Jentic:
## Summary

Adds the two discovery tools to the `jentic mcp` stdio server (local-MCP phase 1, PR 1-B — plan: `jentic-one-plans/themes/local-mcp/phase-1.md`): `search_apis` and `inspect_operation`, completing the read half of the whoami → search → inspect → execute flow. Both pass their envelopes through unchanged from the CLI surface, carry the sibling `instance` stamp on every result, and tolerate the parameter drift models actually produce via one shared normalizer.

Stacked on #1181 (which stacks on #1179); retarget after they merge. Part of #1175.

## Changes

- `cli`: `search_apis` tool → the generated SDK's `SearchOperationsWithResponse` (the same call `searchE` makes), with the CLI's `{schema_version, data, has_more, next_cursor}` list envelope passed through one page per call (the model paginates with `next_cursor`; no `--all`-style auto-follow, which would flood its context). Inputs mirror the CLI surface: `query` + `apis`/`limit`/`cursor`. The 501 "search not enabled" mapping is kept.
- `cli`: `inspect_operation` tool → the `agentops` Inspector seam (raw full inspect document, format fixed to `json`), body redacted through the same `ux.RedactBytes` funnel as `inspectE`. A 404 maps to the shared `RESOLVE_FAILED` soft error — pointed at `next_tool: search_apis` (the identity resolved; the id didn't), while the code-keyed default for auth/config failures stays `get_started`.
- `cli`: shared param normalizer (`mcp_params.go`): `uuid`/`id` → `operation_id`, `params`/`parameters` → `inputs` (declared now per the plan; consumed by 1-C's execute tools), string↔list coercion (comma-split like the CLI's repeatable flags, one-element unwrap), numeric-string ints, stringified-object decode. Conflicting alias values and genuinely malformed arguments return JSON-RPC invalid-params, per the §3.7 error-mapping table.
- `cli`: both tools carry `readOnlyHint: true`, respect the 1-A `--read-only`/`--exclude-tools` filters, and their docstrings encode the flow per master §3.2 (whoami-first "don't execute to probe access", search → inspect → execute) with concrete argument examples.
- Out of scope (1-C): `execute`/`execute_read`, `get_execution_result`, response-size cap, broker-leg CA-pinning routing.

## Risk & rollback

- Low risk: additive tool registrations behind the unreleased `jentic mcp` command; no command-tree, envelope, or auth-surface changes. Revert the squash commit to roll back.

## Test plan

- `GOWORK=off make -C cli check` green (lint + race tests + arch + golden).
- New table-driven normalizer unit tests (aliases, coercions, conflicts, malformed input).
- Handler tests against an httptest control plane: envelope passthrough incl. fresh `instance` stamp, alias/coercion reaching the wire, empty `data` serializing as `[]`, 404 → soft `RESOLVE_FAILED` + `next_tool: search_apis`, no-context → soft `RESOLVE_FAILED` + `next_tool: get_started`, missing args → JSON-RPC invalid-params, secret redaction in inspect bodies.
- 1-A's in-process MCP session test extended: tools/list serves exactly the 4 read-only tools, `--read-only`/`--exclude-tools` filtering, and a full search → inspect round trip over a real client session (including the `id` alias on the wire).

### DIFF
--- a/cli/internal/cli/api/mcp.go
+++ b/cli/internal/cli/api/mcp.go
@@ -104,11 +104,11 @@ func (a *app) mcpE(cmd *cobra.Command, opts *mcpOptions) error {
 	case isClientDisconnect(err):
 		// The MCP client closed (or killed) the stdio pipe — the normal end of
 		// a stdio session, even when it arrives mid-flight. Not an error.
-		logger.Info("mcp client disconnected", "cause", err.Error())
+		logger.Info("mcp client disconnected", "cause", redactedErr(err))
 		err = nil
 	}
 	if err != nil {
-		logger.Error("mcp server exited", "error", err)
+		logger.Error("mcp server exited", "error", redactedErr(err))
 		return reportCoded(aud, err)
 	}
 	return nil
--- a/cli/internal/cli/api/mcp_discovery.go
+++ b/cli/internal/cli/api/mcp_discovery.go
@@ -38,7 +38,11 @@ var searchAPIsParams = []paramSpec{
 	{name: "query", kind: paramString},
 	{name: "apis", aliases: []string{"api"}, kind: paramStringList},
 	{name: "limit", kind: paramInt},
-	{name: "cursor", kind: paramString},
+	// next_cursor is aliased because it's the single most predictable drift:
+	// a model copying the response key straight back as the argument name.
+	// Without the fold the key would be silently dropped and the tool would
+	// re-serve page 1 forever.
+	{name: "cursor", aliases: []string{"next_cursor"}, kind: paramString},
 }
 
 func (s *mcpServer) handleSearchAPIs(ctx context.Context, req *mcp.CallToolRequest) (*mcp.CallToolResult, error) {
@@ -57,7 +61,7 @@ func (s *mcpServer) handleSearchAPIs(ctx context.Context, req *mcp.CallToolReque
 
 	client, err := s.app.controlClient(cctx)
 	if err != nil {
-		s.logger.Warn("search_apis failed", "error", err)
+		s.logger.Warn("search_apis failed", "error", redactedErr(err))
 		return s.softError(cctx, err), nil
 	}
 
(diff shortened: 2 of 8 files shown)


Manuel-Jentic:

An independent reviewer agent attacked this PR's claims (redaction funnel, invalid-params reachability, `softErrorNext` behavior preservation — all verified against go-sdk v1.7.0 sources and `-race` runs). Verdict was **approve-with-nits** (8 MINOR); all addressed in d9b4c040. Record below.

### What held under attack
- No redaction leak path found: every payload — inspect bodies and soft errors included — rides the same `ux.RedactBytes` funnel `inspectE` uses; the structured pass catches non-string secrets at depth.
- Invalid-params is reachable exactly as claimed (raw `AddTool` never validates; `*jsonrpc.Error` propagates untouched).
- `softErrorNext` is provably behavior-preserving for 1-A; empty `data` serializes as `[]`; post-coercion alias-conflict comparison is correct.

### Findings → fixes (d9b4c040)
| # | Finding | Fix |
|---|---|---|
| 1 | `next_cursor` not aliased to `cursor` — the most predictable model drift silently re-serves page 1 | Alias added + table case |
| 2 | `has_more:true` + empty `next_cursor` representable → page-1 pagination loop | `has_more` derived from cursor presence (ux.NewList's rule); empty `next_cursor` omitted; test |
| 3 | Schema comment claimed server-side `required` rejection that raw `AddTool` never does | Comment corrected (client-side validators reject; handler checks are the enforcement) |
| 4 | `limit` docstring promised 1–100 but forwarded anything → server 422 as misleading `INTERNAL_ERROR` | Handler validates bounds → `invalidParams`; tests for -5 and 500 |
| 5 | Leak test only covered the shape the weakest (byte-regex) layer catches | Non-string secret case (`client_secret` object) pinned through the MCP marshal path |
| 6 | 501 mapping + session-level error propagation untested | httptest 501 → `errSearchUnsupported`; session `CallTool({})` → protocol error |
| 7 | MCP log file wrote unredacted error strings (HTTPError embeds raw response bodies) | `redactedErr` helper at all 7 log sites (incl. 1-A's) + log-redaction test |
| 8 | `coerceObject("null")` yielded present-but-nil vs JSON `null` = absent | Nil decoded map now absent; table case |

`GOWORK=off make -C cli check` fully green at d9b4c040.

Issue #1132, opened 2026-08-25
[epic] AWS Marketplace container listing: entitlement gate, ECR publish, listing readiness


ren-jentic:
## Goal

List the `jentic-one-app` container image (plus its Helm deployment artifacts under `deploy/`) as a **paid AWS Marketplace container product** (contract pricing, dimensions `users` + `executions`).

Tracking issue for the engineering side of the listing. Detailed plan and decision record live in the internal plans repo (`jentic-one-plans`, research: `aws-marketplace-readiness*`).

## Done

- [x] CI hygiene: Dependabot docker ecosystem + Trivy CVE gate on CI and release — #1032 (the gate has already caught and driven a real fix: #1060)
- [x] Helm: Marketplace values overlay (`deploy/helm/values/aws-marketplace.yaml`), required DB password guards, no baked-in defaults — #1039, shipped in v0.31.0
- [x] Entitlement gate: config, dependency-free AWS credential chain + SigV4 clients (License Manager `CheckoutLicense` / Metering `RegisterUsage`), cached checker with grace window, pure-ASGI lockout middleware — #1041 (**merged 2026-08-25**; contract pricing is the config default, one `CheckoutLicense` call gates on all configured dimensions)
- [x] ECR repo path recorded + `<seller>` placeholder filled (`.../jentic/jentic-one-app`) — #1133
- [x] ECR publish workflow — #1134 (**merged 2026-08-26**). IAM OIDC role created in the seller account the same day; `MARKETPLACE_ECR_ROLE_ARN` + `MARKETPLACE_ECR_IMAGE` repo variables set
- [x] **First image published to Marketplace ECR** (2026-08-26): the v0.32.1 release publish job was re-run after the variables were set — OIDC login, digest-equal copy of the signed multi-arch index, and cosign + SBOM attestation on the ECR image all green. Every future release publishes automatically
[...]

PR #1184, opened 2026-08-28
fix(entitlement): use Count entitlements and release the checkout seat

ren-jentic:
## Summary

Live testing against the real Marketplace license (buyer test 2026-08-28, #1132) proved the contract-pricing checker would lock out **every legitimate buyer**: the two provisional constants flagged in the code were wrong in exactly the way the comments anticipated. This PR fixes the client (verified live: ENTITLED for the subscribed test account, NOT_ENTITLED for a wrong SKU) **and enables the entitlement gate in the Marketplace overlay with the live listing's IDs** — the two must ship together, so they ride one PR.

## Changes

- `LicenseManagerClient._body`: entitlements sent as `Unit: "Count"` + `Value: "1"` — the live license carries counted dimensions (`users`/`executions`, `Unit: "Count"`, `MaxCount`), and a `Unit: "None"` checkout is rejected with `NoEntitlementsAllowedException` (i.e. the current shape reads as not-entitled for an entitled account)
- **Check the seat back in**: a successful `PROVISIONAL` checkout consumes a seat from `MaxCount` for up to 60 min (`MaxTimeToLiveInMinutes`). A 1-seat contract + app/broker pods re-checking hourly = permanent self-lockout — observed live: one manual checkout locked the test account's license for the full TTL. Every successful checkout now immediately calls `CheckInLicense` with the returned `LicenseConsumptionToken` (allowed: `AllowCheckIn: true`), making the check a pure probe. Check-in failure only logs (the seat self-expires) and never changes the verdict
- **One short-delay retry on NOT_ENTITLED**: seat contention and a genuinely missing license return the *identical* `NoEntitlementsAllowedException`, and a false NOT_ENTITLED is cached for a whole refresh interval; a single 1.5s-delayed retry disambiguates a transient collision (e.g. app + broker booting together)
- **`deploy/helm/values/aws-marketplace.yaml`: entitlement env uncommented** on app + broker with the live listing's IDs (product code `dwhxz1v53k58ew6pr7qzieumd`, product ID/SKU `prod-dd2p2s65dysv6`, dimensions `users,executions` — public identifiers, not secrets). Deliberately in this PR: a chart carrying the env without this client fix locks entitled buyers out
- Base client: extracted `_signed_post` + `_on_entitled` hook (mechanical; metering variant unchanged)
- Tests: unit — Count shape, checkout→check-in sequencing + token propagation, check-in-failure tolerance, contention-retry recovery, exactly-one-retry on persistent denial; render — the enforcing env lands on both pods with the exact IDs

Dimension semantics (decided 2026-08-28): quantities are **commercial terms** for this version — the gate is a boolean entitlement probe; nothing meters executions or counts users at runtime. App-side quota enforcement (license `MaxCount` is readable) is a scoped follow-up.

## Risk & rollback

- **This makes the next listed release the enforcing one**: a non-entitled deployment of that version locks its HTTP surface (503, health excepted) after grace. Entitled buyers are unaffected — verified live against the real license.
- The gate stays fully off for self-host installs (config default `enabled=false`; only the Marketplace overlay sets the env).
- New AWS call surface: `license-manager:CheckInLicense` — already granted by the listing's launch policy (`AWSLicenseManagerConsumptionPolicy`).
- Rollback: revert the squash commit (or re-comment the env block to disable enforcement only).

## Test plan

- `uv run pytest tests/unit/integrations/aws_marketplace/ tests/smoke/test_helm_render.py --no-cov` — 66 passed
- Live against the real listing (buyer account, SKU `prod-dd2p2s65dysv6`): old shape → `NoEntitlementsAllowedException`; `Count`+`Value` → ENTITLED; wrong SKU → NOT_ENTITLED; repeated checkout without check-in → observed seat exhaustion (the motivating bug)
- `ENTITLEMENT_REAL_AWS=1` smoke: negative control passed; entitled case currently blocked by the seat my earlier manual probe consumed (auto-expires ~60 min) — will re-run and report on this PR

Part of #1132

### DIFF
--- a/deploy/helm/values/aws-marketplace.yaml
+++ b/deploy/helm/values/aws-marketplace.yaml
@@ -79,16 +79,16 @@ app:
   # secret. Safe here because global.appSecrets generates real values.
   extraEnv:
     JENTIC_ENV: "production"
-  # Entitlement gate wiring (plan PR 4). The listing is contract-priced with
-  # dimensions `users` + `executions` (decided 2026-08-20); `contract` is the
-  # config default so only the IDs need filling in. DO NOT uncomment until the
-  # deployed image ships the entitlement config: AppConfig rejects unknown
-  # config sections (extra="forbid"), so these env vars crash older images at
-  # boot.
-  #   JENTIC__ENTITLEMENT__ENABLED: "true"
-  #   JENTIC__ENTITLEMENT__PRODUCT_CODE: "<product code from the portal>"
-  #   JENTIC__ENTITLEMENT__LICENSE_SKU: "<product ID from the portal>"
-  #   JENTIC__ENTITLEMENT__LICENSE_DIMENSIONS: "users,executions"
+    # Entitlement gate (contract pricing, dimensions users + executions —
+    # decided 2026-08-20; `contract` is the config default so only the IDs are
+    # set). IDs verified against the live listing 2026-08-28. Enabled together
+    # with the checkout-shape/check-in fix — a release carrying this env MUST
+    # ship that client (Count+Value entitlements, CheckInLicense after
+    # checkout) or entitled buyers lock themselves out.
+    JENTIC__ENTITLEMENT__ENABLED: "true"
+    JENTIC__ENTITLEMENT__PRODUCT_CODE: "dwhxz1v53k58ew6pr7qzieumd"
+    JENTIC__ENTITLEMENT__LICENSE_SKU: "prod-dd2p2s65dysv6"
+    JENTIC__ENTITLEMENT__LICENSE_DIMENSIONS: "users,executions"
 
 broker:
   enabled: true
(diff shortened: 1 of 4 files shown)


ren-jentic:
Real-AWS entitlement smoke is now fully green (14:45 UTC+1, after the pre-fix seat expired), run from the subscribed buyer account with the live listing IDs (product code `dwhxz1v53k58ew6pr7qzieumd`, SKU `prod-dd2p2s65dysv6`, dimensions `users,executions`):

- entitled case → `ENTITLED`
- negative control (wrong SKU) → `NOT_ENTITLED`, never `UNKNOWN`
- back-to-back runs succeed — `CheckInLicense` releases the seat, so repeated probes no longer self-lockout (the exact failure that blocked the earlier run for 60 min)

That was the last open verification — this PR is ready to merge, and the next release becomes the enforcing listed version.

PR #1186, opened 2026-08-28
feat(cli): add execute + get_execution_result MCP tools with broker hardening

Manuel-Jentic:
Stacked on #1183 (→ #1181 → #1179); retarget after they merge. Part of #1175.

## What

PR 1-C of the local-MCP theme: the execute surface for `jentic mcp`.

- **`execute`** (+ **`execute_read`**, GET/HEAD-only) via the `agentops` core: CLI envelope `{schema_version, status, headers, body, execution_id}` + sibling `instance` stamp. `execute` is the only tool carrying `destructiveHint`; `execute_read` rejects non-GET/HEAD operations post-resolve and never takes a body.
- **`get_execution_result(job_id)`** wired to the live `GET /jobs/{id}` + `/jobs/{id}/result` routes — the poll path for held (202) executes. The 202-held envelope passes through as a normal result with its `agent_directive` intact ("poll, never re-send" is encoded in both docstrings); the broker's ask-tier behavior is mocked in tests only.
- **Stdin structural safety**: tool handlers pass an explicit body (or none) into `agentops`; the MCP path cannot reach `executeE`'s stdin fallback. A test swaps `os.Stdin` for a data-laden pipe and proves not a byte is consumed.

## Hardening decisions (for reviewers)

- **CA pinning (§3.7.2)**: new `clictx.BrokerHTTPClient` builds the broker-leg client — SEC-20 CA-pinned when the environment sets `ca_cert_path` (fail closed), with the session `TransportHook` (UA + session header) composed over it, wrap-never-displace. `agentops.DoWith(hc, req)` is the seam; the SDK broker retry policy still decorates the outside, so hooked headers ride every retry. Broker target resolution is a strict subset of `executeE`'s (environment `broker_url` or loopback default — no flag overrides), keeping the SEC-21 pin structural; the malformed-broker_url and remote-CP-without-broker fail-closed guards port unchanged.
- **Response-size cap**: default **128 KiB** (`--max-result-bytes`), applied to relayed execute bodies and job result documents. Far below the 64 MiB transport bound — MCP has no chunking, so this is context protection. Truncation envelope: `{truncated: true, total_bytes, execution_id}` with the leading bytes kept, so the model can act deliberately.
- **Singleflight (§3.7.1)**: `auth.BearerToken`'s mint path dedupes per `(base URL, identity, environment)` via `golang.org/x/sync/singleflight` — N concurrent tool calls at one expiry instant share one exchange. Efficiency only; mint-fresh stays last-writer-wins correct.
- **Error mapping (§3.7 table)** — `next_tool` pointers chosen per failure class:
  - broker denial → soft `BROKER_DENIED`, verbatim `agent_directive` passthrough, `retryable: false`, `next_tool: whoami` (check bindings, never execute to probe); directive-less denials get the UX7-synthesized status-keyed hint.
  - transport → soft `TRANSPORT_ERROR`, `retryable: true`, `next_tool: get_started` (diagnoses a down instance).
  - resolve → soft `RESOLVE_FAILED`, `next_tool: search_apis` (the id is wrong, not the setup); unknown `job_id` points back at `get_execution_result` itself.
  - auth → coded soft errors with the default `get_started` pointer.
  - malformed args → JSON-RPC invalid-params.
  - upstream 4xx/5xx and 202-held → normal results (the upstream's answer is data, not an error).
- Both execute tools respect `--exclude-tools`; `--read-only` withholds exactly `execute` (everything else is `readOnlyHint`).

## Tests

Success envelope + stamp; denial with directive passthrough + synthesized-hint variant; 202-held passthrough (mocked); live-route `get_execution_result` round trip (+ pending, 404, truncation); size-cap truncation; stdin structural safety; CA-pin + hook on the broker leg (TLS httptest, fail-closed check); broker-target guards; read-only/exclude filtering; invalid-params rows; singleflight concurrency (with `-race`); in-process MCP session test extended to the full whoami → search → inspect → execute round trip. `GOWORK=off make -C cli check` green.

Not in scope (deliberate): `include_paths` output filtering (phase 4); ask-tier live end-to-end (7b, cross-plan).

### DIFF
--- a/cli/client/auth/middleware.go
+++ b/cli/client/auth/middleware.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net/http"
 	"time"
+
+	"golang.org/x/sync/singleflight"
 )
 
 // Credentials is the minimal, UX-free input the auth middleware needs. The
@@ -87,16 +89,40 @@ func BearerToken(creds Credentials) (string, error) {
 	tokens, err := ReadTokens(creds.IdentityRef())
 	needsExchange := err != nil || tokens == nil || time.Now().After(tokens.ExpiresAt)
 	if needsExchange {
-		newTokens, xerr := performOAuthExchange(creds)
+		// Singleflight (local-MCP §3.7.1): N concurrent callers hitting one
+		// expiry instant share ONE mint instead of racing N parallel
+		// exchanges. Mint-fresh is already concurrency-safe (V2 has no
+		// refresh tokens; SaveTokens is documented last-writer-wins), so this
+		// is purely an efficiency guard for long-lived concurrent embedders
+		// (the `jentic mcp` server) — every waiter gets the winner's
+		// independently valid token.
+		v, xerr, _ := mintGroup.Do(mintKey(creds), func() (any, error) {
+			newTokens, xerr := performOAuthExchange(creds)
+			if xerr != nil {
+				return nil, fmt.Errorf("failed to authenticate: %w", xerr)
+			}
+			if serr := SaveTokens(creds.IdentityRef(), newTokens); serr != nil {
+				return nil, serr
+			}
+			return newTokens, nil
+		})
 		if xerr != nil {
-			return "", fmt.Errorf("failed to authenticate: %w", xerr)
+			return "", xerr
 		}
-		if serr := SaveTokens(creds.IdentityRef(), newTokens); serr != nil {
-			return "", serr
-		}
-		tokens = newTokens
+		tokens = v.(*TokenSet)
 	}
 	return tokens.AccessToken, nil
+}
+
+// mintGroup dedupes concurrent token exchanges per (base URL, identity,
+// environment). Package-level: BearerToken is stateless per call, so the
+// dedup scope must be the process.
+var mintGroup singleflight.Group
+
+// mintKey scopes the singleflight to one logical credential: the same
+// identity pair against a different base URL must never share a mint.
+func mintKey(creds Credentials) string {
+	return creds.BaseURL + "\x00" + creds.IdentityName + "\x00" + creds.EnvironmentName
 }
 
 // RefreshBearerToken drops any cached token and forces a fresh assertion
--- a/cli/client/auth/middleware.go
+++ b/cli/client/auth/middleware.go
@@ -96,6 +96,13 @@ func BearerToken(creds Credentials) (string, error) {
 		// is purely an efficiency guard for long-lived concurrent embedders
 		// (the `jentic mcp` server) — every waiter gets the winner's
 		// independently valid token.
+		//
+		// Waiters are uncancellable today: BearerToken takes no context, so a
+		// caller whose own deadline fires still blocks here until the
+		// winner's exchange finishes (bounded by the exchange client's own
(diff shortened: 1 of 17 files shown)


Manuel-Jentic:

This PR was flagged as the riskiest of the stack and got the deepest scrutiny from an independent reviewer agent. Verdict was **request-changes** (2 MAJOR, 6 MINOR); all findings addressed in f24ee8af. Record below.

### What held under attack
- Stdin isolation is genuinely structural (no `os.Stdin` reference in `agentops`; the only sniffing lives cobra-side, unreachable from MCP).
- CA pinning: the MCP path always routes through the fail-closed `clictx` client; hook composed inside, retry policy outside — attribution rides every retry.
- SEC-21/SSRF: broker target admits only the environment's `broker_url` or loopback default; scheme smuggling, path traversal, CRLF, and absolute-URL tricks all dead-end; `job_id` percent-escaped.
- The §3.7 error table verified row-for-row; `execute_read`'s post-resolve gate airtight on body/method; singleflight key correct; `get_execution_result` wired to live routes.

### Findings → fixes (f24ee8af)
|---|---|---|---|
| 1 | MAJOR | Response headers escaped the 128 KiB cap entirely — up to ~10 MiB could ride around the context-protection this PR claims | 8 KiB aggregate header budget clamped to the cap; over-budget headers dropped whole with `headers_truncated: true`; normal payloads byte-identical (shared goldens untouched); fat-header test |
| 2 | MAJOR | 30 s per-call deadline silently bounded the broker leg (advertised 60 s ceiling unreachable), and mid-flight timeouts said `retryable: true` — inviting double-execution of unkeyed POSTs | Execute family gets a 90 s deadline (60 s broker ceiling now binding on both surfaces; control-plane tools keep 30 s); `retryable: true` only for GET/HEAD, `Idempotency-Key`, or provably pre-send failures (`TransportFailurePreSend` whitelist over the preserved cause chain); mid-flight unkeyed mutations get "may already have reached the upstream" + `get_execution_result`/idempotency-key recovery; 6 tests |
| 3 | MINOR | Poll-tool transport failures surfaced as `INTERNAL_ERROR` with no hints | Classified to `TRANSPORT_ERROR`, `retryable: true`, `next_tool: get_started` |
| 4 | MINOR | Case-sensitive method gate rejected registry `"get"` | Method normalized once at resolve time |
| 5 | MINOR | "VERBATIM" directive passthrough was struct-projected — unknown future broker fields silently dropped | Raw relay chosen: `Denial.DirectiveRaw` carries the verbatim `agent_directive` JSON; CLI envelope/renderer untouched, goldens byte-identical |
| 6 | MINOR | `coerceJSON`'s string sniffing untested/undocumented | Table tests + normalizer test; reinterpretation stated in the `body` schema |
| 7 | MINOR | Truncation UTF-8 semantics untested; docstring overpromised | Rune-straddle + binary-body tests; docstring corrected |
| 8 | MINOR | Singleflight key isolation unpinned; waiters uncancellable | Two-identity test (two exchanges); `DoChan`+`ctx.Done()` note for a future ctx-aware `BearerToken` |

`GOWORK=off make -C cli check` fully green at f24ee8af (lint incl. gosec/nolintlint, race).

PR #1257, opened 2026-09-04
fix(mcp): answer 405, not the 401 challenge, to credential-less GET /mcp

### LINKED ISSUE #1256: Pre-auth whitelist on /mcp triggers mcp-remote's concurrent-auth race — return 405 to unauthenticated GET
## Verdict from a controlled experiment (2026-09-04)

Lab: minimal mock AS+MCP server with a posture switch, driving `mcp-remote@0.8.3` (latest) exactly as Claude Desktop does. Binary, 100%-deterministic result across 4 runs:

| Posture | Transport | Authorize flows | PKCE challenges per state | Outcome |
|---|---|---|---|---|
| **whitelist (ours)** | http-only | **4** | **4** | 2× invalid_grant, session never recovers |
| strict all-401 | http-only | 1 | 1 | token minted, everything works |
| **whitelist (ours)** | default | **4** | **4** | fails, never recovers |
| strict all-401 | default | 1 | 1 | works |

## Mechanism

Because `initialize`/`tools/list` succeed pre-auth (§3.3 whitelist in `src/jentic_one/mcp/app.py`), the proxy goes live unauthenticated; then **three 401 arms fire concurrently** (fallback-test GET stream, main GET stream on `notifications/initialized`, and the `tools/call` POST). mcp-remote's SDK provider has one fixed `state` per process and keys its verifier file only by state — concurrent flows clobber each other's PKCE verifier → guaranteed `invalid_grant`, plus dead-promise codes nothing consumes. Under strict, the first `initialize` 401s before any concurrency exists → single serial flow → works. (Defect is upstream in mcp-remote ≤0.8.3 — evidence bundle prepared for geelen/mcp-remote — but our posture is the deterministic trigger, and other clients may share the pattern.)

## Fix shape (experiment-verified lead)

Return **405** (not 401) to *unauthenticated* GET `/mcp`: a stateless server offers no unauthenticated stream anyway, and the MCP SDK accepts 405 for GET without entering its auth path. That removes two of the three racing arms, leaving the single POST-armed flow — the one that actually awaits the code. Bearer-carrying GETs keep today's behavior (401 challenge on bad/expired tokens). Preserves the §3.3 pre-auth discovery whitelist.

## Evidence

`/tmp/mcp-remote-race-lab/` — REPORT.md, mock server, driver, and per-run logs (mock.log.jsonl, mcp-remote stderr + debug logs, clobbered verifier files, summary.json).

## Live context

This is the root cause of the all-morning Claude Desktop connect failure (T7); workaround in place is out-of-band token minting. Refs: #1251 (retry-loop registrations), #1252 (revoked-grant wedge), T7 results in the plans repo.

Manuel-Jentic:
## Summary

`mcp-remote` (≤ 0.8.3, and potentially other MCP-TS-SDK-based bridges) deterministically fails OAuth against our `/mcp` mount: because the §3.3 pre-auth whitelist serves `initialize`/`tools/list` without a credential, the proxy goes live unauthenticated and its *only* 401s arrive mid-session and **concurrently** — the SDK fallback-test GET stream, the main GET stream, and the first `tools/call` POST. Each 401 launches an OAuth flow on the same provider, which holds one per-process `state` and a PKCE-verifier file keyed only by state → the flows clobber each other's verifier → guaranteed `invalid_grant` and orphaned auth codes. This PR answers **405 Method Not Allowed** (instead of the 401 challenge) to a `GET`/`HEAD` on `/mcp` that offers **no credential at all**, which removes both GET-armed flows without touching any authenticated behavior or the pre-auth whitelist itself.

## Mechanism (experiment-proven, 100% deterministic)

Lab: a mock server implementing OAuth discovery + a strictly-validating token endpoint + an `/mcp` with a `whitelist|strict` posture switch, driven by real `npx mcp-remote@0.8.3` fed a Claude-like stdio session:

| Run | Posture | Transport | authorize hits | states | PKCE challenges per state | token POSTs | PKCE match | `tools/call` |
|---|---|---|---|---|---|---|---|---|
| 1 | whitelist (ours) | http-only | **4** | 1 | **4** | 2 | ✗ ✗ (`invalid_grant`) | **error** |
| 2 | strict (all-401) | http-only | 1 | 1 | 1 | 1 | ✓ | ok |
| 3 | whitelist (ours) | default | **4** | 1 | **4** | 2 | ✗ ✗ | **error** |
| 4 | strict (all-401) | default | 1 | 1 | 1 | 1 | ✓ | ok |

Root cause is upstream (`NodeOAuthClientProvider._state` fixed per process; verifier file keyed by state only; the `_startOrAuthSse` arm fires `auth()` with no code consumer — a dead promise), but our whitelist posture is the trigger: against a strict server the very first `initialize` 401s **serially**, before any concurrency exists, and everything works.

## Why 405 (fix rationale + spec angle)

- This endpoint is **stateless** (`StreamableHTTPSessionManager(stateless=True)`, JSON responses): it never serves a GET stream to *any* caller — an authenticated GET already answers 405 + `Allow: POST` (pinned pre-existing behavior). So a 401 on an unauthenticated GET invited a token that unlocks nothing.
- The MCP Streamable HTTP transport spec ("Listening for Messages from the Server") is explicit: on a client `GET`, *"the server MUST either return `Content-Type: text/event-stream` … **or else return HTTP 405 Method Not Allowed**, indicating that the server does not offer an SSE stream at this endpoint."* Our conditional 405 is spec-compatible — an unauthenticated GET can never receive a stream, so 405 is a truthful "no stream offered here". The SDK treats it exactly that way, **without entering its auth path** — removing two of the three racing arms and leaving only the POST-armed flow, the one that actually awaits and consumes the auth code (the strict runs show the single-flow path succeeds).
- `Allow: POST` advertises what the mount actually serves: the gate refuses GET, and the SDK transport refuses DELETE in stateless mode (no session to terminate).

## Changes

- `src/jentic_one/mcp/app.py` (control-plane `/mcp` mount): in the credential-less gate arm, `GET`/`HEAD` with **no** `Authorization` and no `X-Jentic-Api-Key` header now answer the same `_stream_refusal` (405 + `Allow: POST`; HEAD mirrors GET's status/headers with no body per RFC 9110 §9.3.2) that the authenticated GET arm already answered. Presence of *any* credential-bearing header — valid, expired, garbage, or a non-Bearer scheme like `Basic` — keeps today's contract exactly (401 challenge with `WWW-Authenticate` on failure).
- `tests/unit/mcp/test_mount_gate.py`: flipped the pre-#1256 pin `test_credential_less_get_stays_401_not_405` ("auth precedes method semantics") into `test_credential_less_get_is_405_stream_refusal` with the rationale and issue link inline; added HEAD-mirrors-GET and presented-bad-credential (garbage bearer / `Basic` / garbage API key → 401 challenge, unchanged) pins; `GET`/`HEAD` removed from the generic credential-less-non-POST challenge parametrize (DELETE/PUT/PATCH/OPTIONS unchanged).
- `docs/mcp-http-endpoint.md`: documented the credential-less GET/HEAD → 405 posture.
- **Deliberately out of scope:** the §3.3 pre-auth whitelist stays; POST/DELETE gate arms unchanged; the disabled-door 404 arm and the oauth-only discovery-challenge placeholder arm (both strict/serial — they don't race) unchanged; the Go CLI's `mcp-daemon` is unaffected: token-less mode already 405s GET, and the token-gated mode answers credential-less GET with a serial 401 before any transport concurrency exists — no race to fix there.

## What stays pinned (unchanged contracts, re-verified by existing tests)

- `mcp.enabled=off, oauth=off` → plain 404 on every method, no `Allow`/`WWW-Authenticate` tell.
- `mcp.enabled=off, oauth=on` → the placeholder's discovery challenge (401 + `resource_metadata`) on every method — including GET (strict posture; no transport exists there, and all-401 does not race).
- Unauth `POST initialize`/`tools/list` → whitelisted 200; unauth `tools/call` POST → 401 challenge; mixed batches → 401.
- Authenticated GET → 405 + `Allow: POST` (pre-existing MINOR-3 pin, now shared with the new arm).
- Presented-but-rejected credential → 401 + challenge on every method.

## Risk & rollback

- Touches the auth gate of the `/mcp` mount, but only the **credential-less GET/HEAD** path: no route, scope, credential-resolution, or whitelist change. A spec-following client that wanted the challenge from GET still gets it from any non-whitelisted POST (and the discovery documents are independently served).
- Deliberate contract flip: the "auth precedes method semantics on GET" pin is replaced (test updated with rationale, not deleted).
- Residual (documented in the experiment report): two *concurrent* unauthorized `tools/call`s could still race the POST arm upstream — but clients serialize calls in practice, and only upstream can fix its missing single-flight guard.
- Rollback: revert the squash commit.

## Test plan

- `make check` equivalent, all green: `ruff check` + `ruff format --check` clean, `mypy` clean (1221 files), `detect-secrets` clean, `make test-arch` **152 passed**.
- `make test-unit` → **3246 passed** (includes the 5 new/updated gate pins in `tests/unit/mcp/test_mount_gate.py`).
- `make test-integration-sqlite` → **731 passed, 13 skipped** (Postgres-only skips). Postgres integration lane left to CI (the local shared fixture DB serves live services and was off-limits).
- `make score` not run locally (requires `JENTIC_API_KEY`; the step is also disabled in CI) — no OpenAPI surface is touched.
- Root-cause evidence: 4 deterministic lab runs of real `mcp-remote@0.8.3` against a posture-switchable mock (whitelist → 4 same-state flows, PKCE mismatch, `invalid_grant`; strict → 1 flow, success), logs and machine-readable summaries retained.

Closes #1256

### DIFF
--- a/docs/mcp-http-endpoint.md
+++ b/docs/mcp-http-endpoint.md
@@ -38,7 +38,15 @@ Flipping the flag needs no rebuild; the gate is evaluated per request. The
 endpoint accepts only `POST` (stateless — no SSE stream, no `Mcp-Session-Id`),
 validates the `Origin` header strictly (403 on mismatch), and requires a
 credential for everything except surface discovery (`tools/list`, `ping`,
-the resource listings).
+the resource listings). A `GET`/`HEAD` that offers **no** credential answers
+`405 Method Not Allowed` (`Allow: POST`) rather than an auth challenge: no
+caller ever gets a stream here, and the MCP SDK reads the 405 as "no stream
+offered" without starting an OAuth flow — challenging that probe made
+OAuth-capable bridges (`mcp-remote` ≤ 0.8.3) launch concurrent flows that
+clobbered each other's PKCE verifier
+([#1256](https://github.com/jentic/jentic-one/issues/1256)). A request that
+presents a credential — valid, expired, or garbage — keeps the standard
+401-challenge contract on every method.
 
 ## Connecting an HTTP-capable MCP client
 
--- a/src/jentic_one/mcp/app.py
+++ b/src/jentic_one/mcp/app.py
@@ -32,7 +32,16 @@ that owns everything the platform — not the SDK — must decide:
   FastAPI dependency: a Starlette mount is a separate ASGI app, so parent
   route dependencies never run here. A missing/invalid credential answers the
   same 401 challenge contract as the placeholder (the ``resource_metadata`` pointer riding
-  only when the OAuth discovery surface is on to serve it).
+  only when the OAuth discovery surface is on to serve it) — EXCEPT a
+  credential-less GET/HEAD, which answers the stream refusal (405) instead:
+  this endpoint never offers a GET stream (stateless), so an unauthenticated
+  GET could never receive one even after authenticating, and challenging it
+  makes SDK-based proxy clients (``mcp-remote`` ≤ 0.8.3) launch concurrent
+  OAuth flows off their fallback-test/main GET-stream arms that clobber each
+  other's PKCE verifier — guaranteed ``invalid_grant``
+  (https://github.com/jentic/jentic-one/issues/1256). A request that DOES
+  present a credential (any ``Authorization`` header or API-key header, even
+  a garbage one) keeps the challenge contract on every method.
 - **The pre-auth whitelist**: ``tools/list``, the SDK's legacy ``initialize``
   fallback (+ ``notifications/initialized``), ``ping``, and the public
   resource listings are served without a credential — a client can always
@@ -136,6 +145,24 @@ def _unauthorized(ctx: Context, request: Request, method: str) -> Response:
     if method == "HEAD":
         return Response(status_code=401, headers=headers)
     return JSONResponse(status_code=401, content={"detail": "Unauthorized"}, headers=headers)
+
+
+def _stream_refusal(method: str) -> Response:
+    """The GET-stream refusal: 405 + ``Allow: POST`` (HEAD: headers, no body).
+
+    ``Allow`` advertises what the mount actually serves — POST only. The gate
+    itself refuses GET (stateless: no server-initiated stream to offer), and
+    the SDK transport refuses DELETE in stateless mode (no session to
+    terminate), so neither belongs in the list. HEAD mirrors GET's status and
+    headers without a body (RFC 9110 §9.3.2).
+    """
+    if method == "HEAD":
+        return Response(status_code=405, headers={"Allow": "POST"})
+    return JSONResponse(
+        status_code=405,
+        content={"detail": "Method Not Allowed"},
+        headers={"Allow": "POST"},
+    )
 
 
 def _is_loopback_origin_host(hostname: str) -> bool:
@@ -424,6 +451,27 @@ class McpMount:
                 return JSONResponse(status_code=413, content={"detail": "Request body too large"})
 
         if credential is None:
+            if method in ("GET", "HEAD") and not self._credential_offered(request):
+                # A request that offers NO credential at all (no Authorization
+                # header, no API-key header) gets the stream refusal, not the
+                # challenge: this stateless endpoint never serves a GET stream
+                # (the post-auth arm below answers the same 405), so there is
+                # nothing a token would unlock and no reason to invite one.
+                # Challenging here is what made SDK-based proxy clients
+                # (mcp-remote ≤ 0.8.3) launch OAuth flows off their
+                # fallback-test and main GET-stream arms concurrently with the
+                # tools/call arm — one per-process `state`, a verifier file
+                # keyed by state → flows clobber each other's PKCE verifier →
+                # guaranteed invalid_grant. The 405 is the MCP Streamable HTTP
+                # spec's own "server does not offer an SSE stream at this
+                # endpoint" answer, which the SDK handles WITHOUT entering its
+                # auth path — leaving only the POST-armed flow, the one that
+                # actually awaits and consumes the auth code.
+                # https://github.com/jentic/jentic-one/issues/1256
+                # A GET that DOES present a credential — valid, expired, or
+                # garbage — keeps today's contract exactly (challenge on
+                # failure, 405 after successful auth).
+                return _stream_refusal(method)
             methods = _body_methods(body) if body is not None else None
             pre_auth = (
                 method == "POST"
@@ -461,11 +509,7 @@ class McpMount:
             # ever ride and that never ends — each such request would pin a
             # connection + task for the life of the process. json_response
             # mode governs POST responses only, so the gate owns this refusal.
-            return JSONResponse(
-                status_code=405,
-                content={"detail": "Method Not Allowed"},
-                headers={"Allow": "POST"},
-            )
+            return _stream_refusal(method)
 
         if body is None:
             return receive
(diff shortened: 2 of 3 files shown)


Manuel-Jentic (commented):
## Adversarial review — verdict: **MERGE-READY** (3 non-blocking notes)

Reviewed `88d3e45c` with maximum suspicion: every claim verified against source, spec, or a live experiment rather than trusted. Full report retained at `/tmp/mcp-reviews/review-1257.md`.

### 1. The security inversion is sound
- Only new bit an anonymous GET learns is `Allow: POST` — but the pre-auth whitelist already serves anonymous `POST initialize`/`tools/list` (full tool surface), and the RFC 9728 PRM doc advertises the mount. The old "auth precedes method semantics" pin protected a secret that stopped existing when the whitelist shipped.
- Spec verified verbatim (Streamable HTTP, "Listening for Messages from the Server"): on GET the server **MUST** either open the stream *"or else return HTTP 405 Method Not Allowed"*. No auth-precedence carve-out; since no caller ever gets a stream here (the pre-existing authenticated-GET 405 pin **is** on main, `test_mount_gate.py:214`), 405 is truthful for every credential-less GET. RFC 9728 §5 constrains what a 401 carries, not which requests must 401; RFC 9110's `Allow`-on-405 requirement is met.
- SDK premise verified in the vendored bundle (`mcp-remote/dist/chunk-6QT6R6YJ.js`, `_startOrAuthSse`): `401 → _authThenStart()`; **`405 → return` — silent, no auth path**. The 405-triggered *transport fallback* in `connectToRemoteServer` can't fire off this arm either (the stream 405 is swallowed, never thrown to connect) — confirmed empirically below. No real client starts auth discovery from GET: TS SDK GET arms fire only after a successful POST connect; the 2024-11-05 back-compat chain POSTs first; an `sse-only` client was already a dead end post-auth (authenticated GET 405'd before this PR too).

### 2. Header-presence gate — probed live, both oauth arms
Credential-less GET/HEAD → 405 `Allow: POST` no challenge; scheme-only bearer, empty-token bearer, garbage bearer, `Basic`, garbage API key, whitespace `Authorization` → all 401 + challenge; valid bearer GET → 405; duplicate `Authorization` headers → no bypass; DELETE/PUT/PATCH/OPTIONS unchanged 401. Cookies/query params are not credential channels the mount honors, so no third shape can disagree with `_credential_offered`. A client whose header an intermediary strips still gets its 401 from the first non-whitelisted POST — the arm that actually drives auth. No trap, no new tell.

### 3. Acceptance test — the leg the PR only reasoned about, now run
Extended the #1256 lab with a `whitelist405` posture mimicking this PR exactly (fresh port/dirs, real `npx mcp-remote@0.8.3`, Claude-like stdio session, browser stubbed):

| Run | Transport | authorize hits | PKCE challenges/state | token POSTs | PKCE | tools/call | recovery |
|---|---|---|---|---|---|---|---|
| whitelist (baseline, #1256) | both | **4** | **4** | 2 | ✗✗ `invalid_grant` | error | error |
| **whitelist405** | http-only | **1** | **1** | 1 | ✓ | **ok** | **ok** |
| **whitelist405** | default (http-first) | **1** | **1** | 1 | ✓ | **ok** | **ok** |
| **whitelist405** (clean-cache rerun) | http-only | **1** | **1** | 1 | ✓ | **ok** | **ok** |

All four former GET-arm hits answer 405 and start **zero** flows; the single surviving POST-armed flow mints the token and the session completes. **The mitigation works, deterministically, on both transport strategies.**

### 4. Tests, gates, hygiene
- Flipped pin preserves the old rationale verbatim + links #1256, parametrized over both oauth arms; new HEAD-mirror and 3× presented-bad-credential pins are honest and pass.
- Gates re-run in the worktree — counts match the PR body exactly: arch **152**, unit **3246** (cov 79.34%), SQLite **731 + 13 skipped**, ruff/format/mypy (1221 files) clean.
- Docs accurate; commit message convention clean; no drive-bys (3 files, all in scope).

### Non-blocking notes
| # | Severity | Finding |
|---|---|---|
| N1 | MINOR, pre-existing | **Authenticated HEAD** falls through the gate (post-auth arm tests `method == \"GET\"` only) to the SDK transport → 405 with `Allow: GET, POST, DELETE`, disagreeing with the gate's `Allow: POST` and advertising a GET nobody is served. Same on main — one-line follow-up: `method in (\"GET\", \"HEAD\")` on the post-auth refusal arm. |
| N2 | INFO | Empty-string `Authorization`/`X-Jentic-Api-Key` count as credential-less → 405. Defensible (nothing was presented; the SDK sets the header only when tokens exist). |
| N3 | MINOR, doc accuracy | PR-body out-of-scope bullet *"the Go CLI's mcp-daemon already 405s GET"* holds only for the **token-less** daemon; the token-gated daemon pins credential-less GET → **401** (`cli/internal/cli/api/mcp_http_gate_test.go:131`). No race there (no OAuth discovery surface), so out-of-scope stands — but please correct the wording **before squash** (the body becomes the squash commit body), and consider a cross-surface-consistency follow-up issue. |

**MERGE-READY.** Squash + merge after the N3 body edit.

PR #1259, opened 2026-09-04
fix(auth): omit unset optional members from OAuth responses

### LINKED ISSUE #1253: /oauth/token emits id_token: null on the MCP arm — latent strict-SDK rejection (same class as #1250)
## Latent defect (review follow-up N1 from PR #1250)

`POST /oauth/token` emits `"id_token": null` on every MCP authorization-code exchange (the D11 arm). The MCP TypeScript SDK's current `OAuthTokensSchema` **rejects null members** — clients only tolerate this today because shipped SDK builds (e.g. Cursor's) predate the `id_token` schema member. The upstream leniency fix ([modelcontextprotocol/typescript-sdk#2462](https://github.com/modelcontextprotocol/typescript-sdk/pull/2462)) is still open, so the moment MCP clients ship a current SDK, every token exchange against us fails with the same invalid_type loop that #1250 fixed for the DCR door.

## Fix shape

Same class as #1250: omit unset optional members from the token response per RFC 6749 §5.1 (`refresh_token`/`id_token` when absent). Check the response-construction arms; note the pre-G11 byte-identical pin belongs to `/oauth/revoke`, not this endpoint (PR #1250 body originally misattributed it — corrected).

## Refs
- #1250 (DCR-door fix, merged `749fa185`), review report /tmp/mcp-reviews/review-1250.md (N1)
- Live evidence: Cursor completes token exchange today (SDK predates the member) — this is a time bomb, not a current outage.

Manuel-Jentic:
## Summary

`POST /oauth/token` serializes unset optional members as JSON `null` — `"id_token": null` on every grant-bearing (agent-channel) code exchange (D11: those codes deliberately mint no id_token) and on every `refresh_token` exchange. RFC 6749 §5.1 and OIDC Core define these members as **omitted when absent, never null**. Strict clients reject the null and drop the whole token response — mcp-remote 0.8.3's zod schema and Cursor's MCP SDK both do — which live-broke Claude Desktop's refresh exchange during T7 testing. Same defect class as the anonymous DCR door fixed in #1250; same cure: `response_model_exclude_none=True` on the route decorator, with an RFC-citing comment.

## Changes

- `auth` — `POST /oauth/token` (`routers/oauth.py`): `response_model_exclude_none=True`. Unset `id_token`/`refresh_token` are now omitted from the body (RFC 6749 §5.1 + OIDC Core §3.1.3.3). This is the fix for the live breakage.
- `auth` — `POST /oauth/introspect` (`routers/oauth.py`): same decorator, per the same-defect audit. RFC 7662 §2.2: the inactive-token body is now exactly `{"active": false}`; previously it carried `sub/scope/exp/token_type: null`.
- `auth` — `POST /register` (`routers/registration.py`): same decorator. RFC 7591 §3.2.1 posture; `claim_token` is documented as "omitted on the OSS single-user default" but was serialized as `null`.
- Tests: new `tests/unit/auth/web/test_oauth_response_shape.py` pins the raw JSON body (agent-channel code exchange with no id_token, refresh exchange, set-id_token retention, introspect inactive shape); `test_registration_router.py` gains omitted-vs-set `claim_token` tests. All follow the #1250 test pattern (`"key" not in data` + `b"null" not in resp.content`).

Audited and **deliberately left alone**:
- `POST /oauth/mint` — `MintResponse` has no optional members; nothing to omit.
- `POST /oauth/revoke` — returns an empty 200 (JSON arm) / hand-built RFC 6749 §5.2 error bodies (form arm); no Pydantic response model, no nulls.
- Discovery documents (`routers/discovery.py`) — hand-built dicts with no null members; the root RFC 8414 doc is pinned byte-identical by a golden test.
- `GET /auth/idp` — emits `provider: null` when IdP login is disabled, but it is an internal SPA capability descriptor, not an RFC document; the SPA reads the key.
- Platform admin APIs (`agents.py`, `service_accounts.py`, `oauth_grants.py`, `identity.py`) — nullable fields there are the platform's normal JSON contract (UI TS client models them as nullable); no absence-vs-null spec semantics.

## Risk & rollback

- Touches the OAuth token-plane response shapes (`/oauth/token`, `/oauth/introspect`, `/register`). Omission of previously-null keys is safe for tolerant readers (the Go CLI's `encoding/json` treats absent and null identically; the UI TS client models these as optional) and is what strict clients require.
- Generated OpenAPI spec is unchanged (`make openapi` produced no drift — `exclude_none` affects serialization, not schema).
- Revert: revert the squash commit.

## Test plan

- `make check` — ruff, ruff format, mypy all green (the `score` step needs a `JENTIC_API_KEY`/Docker image not available in this environment; the spec it scores is untouched by this PR).
- `make detect-secrets` — clean.
- `make test-arch` — 153 passed.
- `make test-unit` — 3256 passed, coverage 79.34%.
- Empirical verification on the live stack (hot-patched worktree): with the decorator fix, both the authorization_code and refresh_token legs return bodies with no null keys, and Claude Desktop's refresh exchange succeeds.

Closes #1253

### DIFF
--- a/src/jentic_one/auth/web/routers/oauth.py
+++ b/src/jentic_one/auth/web/routers/oauth.py
@@ -226,6 +226,13 @@ _TOKEN_REQUEST_BODY: dict[str, object] = {
     "/oauth/token",
     dependencies=[Depends(_check_token_rate_limit)],
     openapi_extra=_TOKEN_REQUEST_BODY,
+    # RFC 6749 §5.1 + OIDC Core §3.1.3.3: optional members the grant did not
+    # mint (id_token on grant-bearing agent-channel codes per D11, refresh_token
+    # on grants that don't rotate one) are omitted from the response, never
+    # emitted as JSON null — strict clients (mcp-remote's zod schema, Cursor's
+    # MCP SDK) reject nulls and drop the whole token response. Same posture as
+    # the DCR door's RFC 7591 omission (oauth_client_registration.py).
+    response_model_exclude_none=True,
 )
 async def token_endpoint(
     request: Request,
@@ -566,7 +573,13 @@ async def revoke_endpoint(
 router.include_router(revocation_router)
 
 
-@router.post("/oauth/introspect")
+@router.post(
+    "/oauth/introspect",
+    # RFC 7662 §2.2: members the server has no value for are omitted from the
+    # introspection response, never emitted as JSON null (the inactive-token
+    # body is exactly `{"active": false}`).
+    response_model_exclude_none=True,
+)
 async def introspect_endpoint(
     body: IntrospectRequest,
     identity: Identity = get_current_identity(allow_expired_password=True),
--- a/src/jentic_one/auth/web/routers/registration.py
+++ b/src/jentic_one/auth/web/routers/registration.py
@@ -35,7 +35,15 @@ def _extract_rat(request: Request) -> str:
     return auth[7:]
 
 
-@router.post("/register", status_code=201)
+@router.post(
+    "/register",
+    status_code=201,
+    # RFC 7591 §3.2.1: optional response members without a value (claim_token
+    # on the OSS single-user default — no claim-token minter installed) are
+    # omitted, never emitted as JSON null. Same posture as the anonymous DCR
+    # door (oauth_client_registration.py) and /oauth/token.
+    response_model_exclude_none=True,
+)
 async def register_endpoint(
     body: RegisterRequest,
     reg_svc: RegistrationService = Depends(get_registration_service),
--- a/tests/unit/auth/web/test_oauth_response_shape.py
+++ b/tests/unit/auth/web/test_oauth_response_shape.py
@@ -1,0 +1,194 @@
+"""Unit tests pinning the raw JSON shape of OAuth token-plane responses.
+
+RFC 6749 §5.1 / OIDC Core §3.1.3.3 (token endpoint) and RFC 7662 §2.2
+(introspection): optional members the server has no value for are OMITTED
+from the response body — never serialized as JSON ``null``. Strict clients
+(mcp-remote 0.8.3's zod schema, Cursor's MCP SDK) reject a null ``id_token``
+or ``refresh_token`` and drop the whole token response, which live-broke
+Claude Desktop's refresh exchange. Pinned on the raw JSON body, mirroring
+the anonymous DCR door's omission tests
+(test_oauth_client_registration_router.py, PR #1250).
+"""
+
+from __future__ import annotations
+
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
+
+from jentic_one.auth.services.errors import AuthServiceError
+from jentic_one.auth.web.errors import service_error_handler
+from jentic_one.auth.web.routers import oauth
+from jentic_one.shared.auth.identity import Identity
+from jentic_one.shared.config import AuthConfig, PlatformClientConfig
+from jentic_one.shared.web.deps import resolve_identity
+
+_PLATFORM_CLIENT_ID = "jentic-one-spa"
+_PUBLIC_CLIENT_ID = "oc_public_mcp_client"
+
+
+def _fake_identity() -> Identity:
+    return Identity(sub="usr_test", email="test@example.com", permissions=[])
+
+
+@pytest.fixture()
+def client() -> TestClient:
+    app = FastAPI()
+    app.include_router(oauth.router)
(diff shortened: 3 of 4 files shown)


Manuel-Jentic (commented):
## Adversarial review — verdict: **MERGE-READY** (3 non-blocking notes)

Reviewed `6797da42` with maximum suspicion in a dedicated worktree: every claim verified against source, spec text, or an experiment rather than trusted. Full report retained at `/tmp/mcp-reviews/review-1259.md`.

### 1. The core `exclude_none` risk — no null-meaningful member can be dropped
- **TokenResponse**: `access_token`/`expires_in` required non-Optional; `token_type: str` non-Optional and explicitly set at all four construction arms — `exclude_none` is structurally unable to remove any of them. The only Optional members (`refresh_token`, `id_token`) are exactly RFC 6749 §5.1 / OIDC Core §3.1.3.3's omitted-when-absent members.
- **IntrospectResponse**: `active` required; `TokenService.introspect` always sets `sub/scope/exp/token_type` for a found token (`scope` degenerates to `""`, never None) — only the not-found arm returns bare `{"active": false}`, which is the RFC 7662 §2.2 SHOULD shape.
- **RegisterResponse**: only `claim_token` is Optional; `grant_types`/`token_endpoint_auth_method` have non-None defaults and stay present. The Go-CLI claim **verified in code**: `cli/client/auth/register.go` `ClaimToken string` — `encoding/json` leaves `""` for both JSON null and absent key (absent ≡ null, exactly as the PR body claims), and `register_flow.go` branches on `!= ""`.

### 2. Introspection consumers — none can break
No internal service calls `/oauth/introspect` over HTTP (verification goes through the in-process resolvers / broker `CachedTokenValidator`); the enterprise overlay has zero matches; the Go CLI and UI have only *generated* bindings (Go: `*string` + `omitempty`, absent ≡ nil) with **no call sites**; integration tests hit the service method pre-serialization. The previously-present null keys have no reader anywhere.

### 3. Tests — genuinely raw-body, and they actually pin the fix
- Assertions are on the client-side parse of the raw HTTP bytes plus `b"null" not in resp.content` — the #1250 pattern, faithfully. Refresh leg, agent-channel (D11) omission, set-id_token retention, introspect inactive shape, and register omit+keep are all covered.
- **Mutation check run**: with `src/` reverted to `origin/main` the 4 omission tests fail (retention tests still pass); with the fix, 11/11 pass. No vacuous assertions.

### 4. Audit-table spot-checks — every "deliberately left alone" claim verified
`/oauth/mint` (`MintResponse` all-required ✓), `/oauth/revoke` (bare 200 / hand-built §5.2 dicts with both members always set ✓), all discovery documents (hand-built, no None members ✓), `GET /register/{agent_id}` (`RegistrationStatusResponse` has no Optional members ✓), `/oauth-clients` already fixed by #1250 ✓. `GET /auth/idp`'s `provider: null` is correctly *left alone* — the SPA (`ui/src/shared/api/idp.ts`) reads that key; blanket application would have broken it. Sweep of the remaining auth-plane routers found no missed same-class defect (authorize.py is redirects/HTML only; the platform admin routers are non-RFC surfaces with null-tolerant generated clients).

### 5. Gates — re-run in the review worktree, counts match the PR body exactly
ruff + format + mypy (**1223 files**) clean · detect-secrets clean · arch **153 passed** · unit **3256 passed, cov 79.34%** · `make openapi` — no drift (only the `.secrets.baseline` timestamp churned; spec untouched, confirming `exclude_none` is serialization-only).

### 6. Hygiene
Single `fix(auth)` commit (correct type — real production breakage), 4 files, no drive-bys; decorator comments RFC-citing per the #1250 precedent; PR body matches the diff and the Risk section names the touched surfaces. The ":8000 live-verified" bullet is author-claimed (that stack is out of bounds for this review); my raw-body reproduction of both legs is consistent with it.

### Non-blocking notes
| # | Severity | Finding |
|---|---|---|
| N1 | MINOR, pre-existing | `TokenResponse` has no `scope` member at all — RFC 6749 §5.1 REQUIRES `scope` in the response when the issued scope differs from the requested one, and the platform *does* downscope (client `allowed_scopes` ceiling, grant-scope intersection). Not an `exclude_none` issue; suggest a follow-up issue. |
| N2 | INFO, pre-existing | Introspection `token_type` returns `"access_token"`/`"refresh_token"`; RFC 7662 §2.2 defines it as the RFC 6749 §7.1 token type (`"Bearer"`) — current values look like `token_type_hint` values. No consumer today, harmless, but worth fixing before one appears. |
| N3 | INFO | No raw-shape test for the introspect **active** arm (all optional members set). The service always sets them and the token-endpoint retention tests cover the keeps-set-members direction — a one-line add if you want symmetry. |

**MERGE-READY.** Squash + merge; none of N1–N3 blocks.

PR #1261, opened 2026-09-04
fix(auth): dedupe software_id-less DCR registrations by name + redirect set

### LINKED ISSUE #1251: Pending-approval DCR clients retry-loop and mint unbounded duplicate registrations (G13)
## Observed (live, Cursor 2026-09-04)

A client whose DCR registration lands `pending` (the default posture since #1247) gets the D7 awaiting-approval page from `/authorize`. No OAuth callback ever fires, so strict clients (Cursor) treat the auth attempt as failed, clear their cached registration, and **re-register** — every retry mints a fresh `pending` row. Observed 16 duplicate 'Cursor' rows in the approval queue within minutes; several within one 25-second burst (inside the registration rate limit). Approving one row does not stop the loop: the client is already retrying under a different client_id.

## Why dedupe misses

Cursor sends no `software_id`, so the D8 dedupe key `(software_id, redirect_uris fingerprint)` never matches (no `software_id` → no dedupe, by design).

## Impact

- Approval queue floods with duplicates; the admin cannot tell which row the client is currently using.
- First-connect UX for the now-default approval-first posture is effectively broken for clients with this retry behavior.

## Design candidates (needs the G12/G13 design pass, gap register in plans repo)

1. Extend the dedupe fingerprint for software_id-less registrations, e.g. `(client_name, redirect_uris fingerprint)` — weigh self-chosen-name collision/squat risk.
2. Tighten the registration bucket so a retry burst can't mint N rows in seconds.
3. Approval-queue UI affordance that collapses/expires duplicate pending rows.
4. Consider the awaiting-approval page's no-callback shape: it's what triggers client retries (related G12 design pass).

Refs: gap register G13 (plans repo, `themes/local-mcp/README.md`), D7 amendment #1247, D8 dedupe in `auth/services/oauth_dcr_service.py`.

Manuel-Jentic:
## Summary

With the approval-first default (`auto_approve_clients=false`, #1247), a DCR client parked `pending` gets the D7 awaiting-approval page from `/authorize` — no OAuth callback ever fires, so strict clients (Cursor 3.19.7 observed) treat auth as failed, drop their cached registration, and re-register on every retry. Cursor sends no `software_id`, so the D8 dedupe key `(software_id, redirect_uris_fingerprint)` never matched and every retry minted a fresh pending row: **34 duplicate "Cursor" `oauth_clients` rows accumulated in one morning of live testing** (G13, gap register in the plans repo; approach approved by Manuel 2026-09-04). Each duplicate also broke continuity: a client that lost its local cache came back as a brand-new client, forcing a fresh approval and a fresh consent.

This PR extends the DCR dedupe with a fallback identity for software_id-less registrations, so a re-registering client re-attaches to its existing row instead of minting a new one.

## Changes

- `auth` — `OAuthDcrService.register`: registrations **with** a `software_id` keep the existing D8 key untouched; registrations **without** one now dedupe on the fallback key `(client_name, redirect_uris_fingerprint)`, matched only against DCR-source rows that *also* carry no `software_id`. Winner selection is shared with the D8 path (approved > pending > denied, then oldest — the F6 race rule).
- `auth` — falsy/whitespace-only `software_id` (`""`, `"   "`) is normalized to `None` at the service boundary, before both the key-space branch and the insert (review F1): `""` passes schema validation but previously took the `IS NULL` lookup while *storing* `""` — a row in neither key space that re-opened the #1251 loop for empty-default serializers. The response echoes the normalized value. Normalizing (not rejecting) follows RFC 7591's ignore-don't-reject posture.
- `auth` — every dedupe re-attach now writes an **audit row** (actor `dcr`, reason `"anonymous DCR re-attach (dedupe hit)"`, `after={client_id, approval_status}`, origin `mcp`) so anonymous re-attaches — including probes of the 200-oracle and denied-row retries — are admin-visible (review F2). Deliberately **no event**: retries must not spam the approval queue, which is the point of the fix.
- `admin` — `OAuthClientRepository.list_dcr_by_name_dedupe_key`: the fallback lookup. `software_id IS NULL` keys into the existing `(software_id, redirect_uris_fingerprint)` covering index — **no new index, no migration** (the columns already exist).
- Response contract unchanged: 201 on create, 200 echoing the existing row's `client_id` on a dedupe hit — the pre-existing D8 contract. RFC 7591 §3.2.1 prescribes 201 for a *new* registration; a re-attach is deliberately not one, and Cursor's MCP SDK accepts the 200 shape (proven live by the D8 path since 3a).
- Docstring/comment updates in the router, model, and lifecycle test where they asserted "no software_id → no dedupe"; regenerated `openapi/` + `ui/openapi.json` (route description only).
- Out of scope (tracked separately in the G13 design candidates): registration rate-limit tightening (stays as-is per the approved design), the approval-queue collapse-duplicates UI affordance, the P2 approval-in-flow page that removes the retry trigger itself, and the reviewer's two hardening options (require `software_id` for approved re-attach; tighter registration bucket) — follow-up issue material.

## Risk & rollback

- Auth surface (anonymous DCR door, `POST /oauth-clients`). **Security analysis of re-attach** — the fallback key is client-chosen data, so assume an attacker replays an approved client's `client_name` + exact redirect URIs:
  - They receive the existing `client_id` — public information per RFC 6749 §2.2 (it travels in browser URLs and is echoed to any registrant).
  - They get **no** `client_secret`: this door only mints public clients (`token_endpoint_auth_method=none`); auth is PKCE-bound per flow.
  - They **cannot mutate** the row: a dedupe hit is read-only — approval status, stored name, scopes, and redirect set are preserved verbatim; a differing redirect set is a different fingerprint and therefore a *new* registration, so an approved client's redirect URIs can never be widened by re-registration.
  - **The honest D7 delta (review F3):** pre-PR, an exact-replay impersonator landed `pending` and stalled at the admin gate; post-PR they skip it and can drive `/authorize` under the victim's name. This is accepted deliberately because (a) consent remains unconditional for registered clients — the user must log in, see the consent page, and actively pick an agent; (b) the authorization code is delivered only to a URI in the **victim's** registered set — for the loopback URIs this door's clientele uses, interception requires co-residency on the victim's machine, and a co-resident attacker can generally read the real client's cached registration anyway; (c) the pre-PR gate had no discriminating power against exact replays — an impersonator's fresh registration was content-identical to the real client's, and the 34-duplicate queue made rubber-stamping the realistic outcome. The gate's real value there was friction + visibility; the F2 audit row restores the visibility half, which is the recoverable one.
  - The two key spaces never cross-match: a software_id-less replay cannot adopt a row registered *with* a `software_id`, and vice versa (`test_dedupe_key_spaces_never_cross_match`); with F1's normalization, `""`/whitespace spellings land in the fallback key space rather than escaping both. The `registration_source='dcr'` filter keeps admin-created rows out of reach of the anonymous door entirely (pinned adversarially).
  - Residual: a 200-vs-201 existence oracle for a (name, redirect-set) pair — now audit-logged per probe (F2) and bounded by the unchanged per-IP registration rate limit.
- **Denied rows re-attach** (same policy as D8): a denied client stays denied instead of getting a fresh pending second chance. **Pre-squat variant (review F4):** an attacker can pre-register a not-yet-installed client's (name, well-known redirect set), eat a denial, and the real client will later silently re-attach to the denied row. Recovery is admin-actioned: the admin **re-approves** the denied row (deny is reversible — the recovery arm is covered by `test_denied_then_approved_recovery_emits_approved_event`) or **deletes/deactivates** it so the next registration mints a fresh pending row. The F2 audit trail is what surfaces the stuck client: repeated `"anonymous DCR re-attach (dedupe hit)"` entries against a denied row are the signal to look.
- Known dedupe misses (accepted): an admin renaming a software_id-less row takes it out of the fallback key — a client re-registering after that mints a new pending row (treated as a distinct identity, which the rename arguably made it); a client that later starts sending `software_id` re-registers once. Concurrent same-key registrations can still race to two rows (no unique index — consistent with no-migration); the winner rule deterministically converges later retries (review F5, accepted as-is).
- No migration, no config change. Rollback: revert the squash commit.

## Test plan

- New integration tests (`tests/integration/auth/test_oauth_dcr_registration.py`, real DB per the no-DB-mocking rule; replaces the now-obsolete `test_no_software_id_never_dedupes`): same name + redirect set → same `client_id`, no new row, no second admin alert; different name → new row; different redirect set → new row; dedupe hit preserves pending/inactive/name/redirects/scopes verbatim; approved-row re-attach stays approved + active; software_id ↔ name key spaces never cross-match; **F1 regression** — two `software_id=""` registrations dedupe to one row and `""`/whitespace/absent spellings share one key space with `software_id IS NULL` stored; **F2** — dedupe hit writes the audit row (reason, actor, after) with no event; **F6(b)** — fallback-arm denied re-attach stays denied; **F6(c)** — an admin-source row with matching name + fingerprint is never adopted by the anonymous door.
- Existing D8 suite (software_id dedupe, F6 race preference, denied re-attach, metadata echo minimization) unchanged and passing; router unit tests already pin the 200-dedupe response shape.
- Gates run locally at `1f38c46b`: `make lint` (ruff check + format + mypy, 1222 files) clean, `make detect-secrets` clean, `make test-arch` **153 passed**, `make test-unit` **3250 passed**, `make test-integration-sqlite` **740 passed, 13 skipped**. `make openapi` regenerated with no drift.

## Review guide

- Start with `OAuthDcrService.register` (the normalization line, the two-key branch, and the audited dedupe arm), then `list_dcr_by_name_dedupe_key` (the `software_id IS NULL` + `registration_source='dcr'` scoping is the security seam), then the adversarial tests.

Closes #1251

### DIFF
--- a/cli/client/generated/control/client.go
+++ b/cli/client/generated/control/client.go
@@ -5987,8 +5987,12 @@ type ClientInterface interface {
 	// Register a public OAuth client anonymously (RFC 7591 subset).
 	//
 	// Returns 201 with the new ``client_id``, or 200 with the **existing** row's
-	// ``client_id`` on an exact (``software_id`` + redirect-URI set) dedupe match
-	// (D8). No client_secret is ever issued here and no registration_access_token
+	// ``client_id`` on an exact dedupe match (D8, extended per G13/#1251):
+	// (``software_id`` + redirect-URI set), falling back to (``client_name`` +
+	// redirect-URI set) for registrations without a ``software_id`` — so a
+	// pending client's awaiting-approval retry loop re-attaches instead of
+	// minting duplicate rows. No client_secret is ever issued here and no
+	// registration_access_token
 	// is returned (D12). New rows await admin approval unless the deployment
 	// auto-approves registrations (D9). The ``server.mcp.oauth.enabled`` gate
 	// lives on the route class — a disabled door 404s before this handler,
@@ -6004,8 +6008,12 @@ type ClientInterface interface {
 	// Register a public OAuth client anonymously (RFC 7591 subset).
 	//
 	// Returns 201 with the new ``client_id``, or 200 with the **existing** row's
-	// ``client_id`` on an exact (``software_id`` + redirect-URI set) dedupe match
-	// (D8). No client_secret is ever issued here and no registration_access_token
+	// ``client_id`` on an exact dedupe match (D8, extended per G13/#1251):
+	// (``software_id`` + redirect-URI set), falling back to (``client_name`` +
+	// redirect-URI set) for registrations without a ``software_id`` — so a
+	// pending client's awaiting-approval retry loop re-attaches instead of
+	// minting duplicate rows. No client_secret is ever issued here and no
+	// registration_access_token
 	// is returned (D12). New rows await admin approval unless the deployment
 	// auto-approves registrations (D9). The ``server.mcp.oauth.enabled`` gate
 	// lives on the route class — a disabled door 404s before this handler,
@@ -9642,8 +9650,12 @@ func (c *Client) UpdateNote(ctx context.Context, noteId string, params *UpdateNo
 // Register a public OAuth client anonymously (RFC 7591 subset).
 //
 // Returns 201 with the new “client_id“, or 200 with the **existing** row's
-// “client_id“ on an exact (“software_id“ + redirect-URI set) dedupe match
-// (D8). No client_secret is ever issued here and no registration_access_token
+// “client_id“ on an exact dedupe match (D8, extended per G13/#1251):
+// (“software_id“ + redirect-URI set), falling back to (“client_name“ +
+// redirect-URI set) for registrations without a “software_id“ — so a
+// pending client's awaiting-approval retry loop re-attaches instead of
+// minting duplicate rows. No client_secret is ever issued here and no
+// registration_access_token
 // is returned (D12). New rows await admin approval unless the deployment
 // auto-approves registrations (D9). The “server.mcp.oauth.enabled“ gate
 // lives on the route class — a disabled door 404s before this handler,
@@ -9669,8 +9681,12 @@ func (c *Client) RegisterOauthClientEndpointWithBody(ctx context.Context, conten
 // Register a public OAuth client anonymously (RFC 7591 subset).
 //
 // Returns 201 with the new “client_id“, or 200 with the **existing** row's
-// “client_id“ on an exact (“software_id“ + redirect-URI set) dedupe match
-// (D8). No client_secret is ever issued here and no registration_access_token
+// “client_id“ on an exact dedupe match (D8, extended per G13/#1251):
+// (“software_id“ + redirect-URI set), falling back to (“client_name“ +
+// redirect-URI set) for registrations without a “software_id“ — so a
+// pending client's awaiting-approval retry loop re-attaches instead of
+// minting duplicate rows. No client_secret is ever issued here and no
+// registration_access_token
 // is returned (D12). New rows await admin approval unless the deployment
 // auto-approves registrations (D9). The “server.mcp.oauth.enabled“ gate
 // lives on the route class — a disabled door 404s before this handler,
@@ -21690,8 +21706,12 @@ type ClientWithResponsesInterface interface {
 	// Register a public OAuth client anonymously (RFC 7591 subset).
 	//
 	// Returns 201 with the new ``client_id``, or 200 with the **existing** row's
-	// ``client_id`` on an exact (``software_id`` + redirect-URI set) dedupe match
-	// (D8). No client_secret is ever issued here and no registration_access_token
+	// ``client_id`` on an exact dedupe match (D8, extended per G13/#1251):
+	// (``software_id`` + redirect-URI set), falling back to (``client_name`` +
+	// redirect-URI set) for registrations without a ``software_id`` — so a
+	// pending client's awaiting-approval retry loop re-attaches instead of
+	// minting duplicate rows. No client_secret is ever issued here and no
+	// registration_access_token
 	// is returned (D12). New rows await admin approval unless the deployment
 	// auto-approves registrations (D9). The ``server.mcp.oauth.enabled`` gate
 	// lives on the route class — a disabled door 404s before this handler,
@@ -21707,8 +21727,12 @@ type ClientWithResponsesInterface interface {
 	// Register a public OAuth client anonymously (RFC 7591 subset).
 	//
 	// Returns 201 with the new ``client_id``, or 200 with the **existing** row's
-	// ``client_id`` on an exact (``software_id`` + redirect-URI set) dedupe match
-	// (D8). No client_secret is ever issued here and no registration_access_token
+	// ``client_id`` on an exact dedupe match (D8, extended per G13/#1251):
+	// (``software_id`` + redirect-URI set), falling back to (``client_name`` +
+	// redirect-URI set) for registrations without a ``software_id`` — so a
+	// pending client's awaiting-approval retry loop re-attaches instead of
+	// minting duplicate rows. No client_secret is ever issued here and no
+	// registration_access_token
 	// is returned (D12). New rows await admin approval unless the deployment
 	// auto-approves registrations (D9). The ``server.mcp.oauth.enabled`` gate
 	// lives on the route class — a disabled door 404s before this handler,
@@ -38885,8 +38909,12 @@ func (c *ClientWithResponses) UpdateNoteWithResponse(ctx context.Context, noteId
 // Register a public OAuth client anonymously (RFC 7591 subset).
 //
 // Returns 201 with the new “client_id“, or 200 with the **existing** row's
-// “client_id“ on an exact (“software_id“ + redirect-URI set) dedupe match
-// (D8). No client_secret is ever issued here and no registration_access_token
+// “client_id“ on an exact dedupe match (D8, extended per G13/#1251):
+// (“software_id“ + redirect-URI set), falling back to (“client_name“ +
+// redirect-URI set) for registrations without a “software_id“ — so a
+// pending client's awaiting-approval retry loop re-attaches instead of
+// minting duplicate rows. No client_secret is ever issued here and no
+// registration_access_token
 // is returned (D12). New rows await admin approval unless the deployment
 // auto-approves registrations (D9). The “server.mcp.oauth.enabled“ gate
 // lives on the route class — a disabled door 404s before this handler,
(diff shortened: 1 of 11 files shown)


Manuel-Jentic (commented):
**Verdict: NEEDS-WORK** — one empirically confirmed bug against the PR's own goal (F1), plus a required observability fix on the new re-attach path (F2). The headline design question (full re-attach incl. approved rows) I conclude is *acceptable with those conditions* — analysis under F3. Gates re-run in a clean worktree and all counts confirmed.

---

## F1 — HIGH (bug): `software_id: ""` defeats the fix and mints unreachable-key orphan rows

`OAuthClientRegistrationRequest.software_id` has no `min_length`, so `""` passes schema validation. In `OAuthDcrService.register` the branch is truthiness-based (`if software_id:`), so `""` takes the **fallback** arm and looks up `software_id IS NULL` — but `create(..., software_id=software_id)` then stores `""` (NOT NULL). The row lands in **neither** key space:

- a second `software_id: ""` registration finds nothing (`IS NULL` ≠ `""`) → **new row every retry — the exact #1251 loop this PR closes**;
- a later `software_id: null` registration also can't adopt the `""` row.

Reproduced against this branch with a scratch integration test (real DB, no mocking):

```
software_id='' re-registration minted a duplicate row: 2 rows, stored software_id values ['', '']
```

Any client that emits `"software_id": ""` (empty-default serializers are common) re-opens the queue-spam bug, silently. Two PR-body claims are also falsified by this case: "the two key spaces never cross-match" (a `""` row is in *neither* space) and "registrations without a software_id now dedupe".

**Fix (small):** normalize falsy/whitespace-only `software_id` to `None` at the top of `register()` (before both the branch and the `create`), and/or `min_length=1` on the schema field. Add the NULL-vs-empty-string test — `test_dedupe_key_spaces_never_cross_match` covers NULL vs values but not `""`.

## F2 — MEDIUM (required): a dedupe re-attach leaves no audit trail — the approved-client_id oracle is invisible

A dedupe hit returns before `record_audit`/`emit_event_best_effort` (`oauth_dcr_service.py`, the `return winner, False` arm) — the only trace is a structlog INFO line in the router. Keeping the *event* silent is correct (that's the point: no queue spam), but no **audit row** means:

- an anonymous attacker probing the 200-vs-201 oracle (replay name + redirect set → learn the approved `client_id`) leaves zero forensic trace in the audit system this codebase otherwise treats as reconstruction-grade (cf. `record_consent_decision`'s own docstring);
- a real client bouncing off a **denied** row retries forever and the admin never learns it's still alive (deny is silent by design, dedupe is silent, no new pending row anymore — this PR removed the one signal that existed).

This pattern is inherited from D8, but this PR is the moment the oracle becomes reachable for the name-only client population, so it should land with it. **Fix (small):** write an audit entry (actor `dcr`, reason e.g. "anonymous DCR re-attach (dedupe hit)", `after={client_id, approval_status}`) on the dedupe arm — no event, queue stays clean.

## F3 — Design analysis (advisory, no change requested): full re-attach of approved rows is an acceptable D7 trade-off — but only because of what F2 fixes

I worked the full chain for an attacker who replays an approved client's `client_name` + exact redirect set:

1. **What they get:** the approved `client_id` (200-dedupe), anonymously, rate-limit-bounded. The row itself is untouched (verified: dedupe arm is read-only; differing redirect set → different fingerprint → new *pending* row, so approved redirect sets can't be widened — pinned by `test_no_software_id_different_redirect_set_creates_new_row`).
2. **What the client_id enables:** `/authorize` under the victim's *name* on the consent page (`app_name` is the H1). Because the row is approved, the D7 awaiting-approval page never renders — this is the real delta: **pre-PR an exact-replay impersonator landed pending and stalled at the admin gate; post-PR they skip it.**
3. **What still stands between them and tokens:** consent is unconditionally required for registered clients (`needs_consent = oauth_client is not None` in `authorize.py` — no consent-skip on prior grants), the user must log in and actively pick an agent, and the code is delivered only to a URI in the **victim's** registered set. For the loopback URIs this door's clientele uses (`http://localhost:PORT/...`), interception requires **co-residency** on the machine where the victim's browser runs — and a co-resident attacker who can bind the victim's ports can generally also read the real client's cached registration, i.e. already has the client_id. For https redirects, the attacker must control the victim client's own host. PKCE is the attacker's own in their flow, so it's neutral here — but it does mean intercepting the *real* client's flow stays dead.
4. **Was D7 protecting against this before?** Only nominally: an impersonator's fresh registration is *content-identical* to the real client's (same name, same redirects, anonymous door) — an admin cannot distinguish them, and the pre-PR 34-duplicate queue made rubber-stamping the realistic outcome. The gate's value against exact-replay was friction + visibility, not discrimination. This PR removes both; F2 restores the visibility half, which is the recoverable one.

Conclusion: with F2 in place, full re-attach is defensible (and pending-only genuinely wouldn't fix the cache-loss → re-consent half of #1251). Two hardening options worth a follow-up issue, not this PR: require `software_id` for approved re-attach (costless for the D8 population, but does nothing for Cursor — the population that matters here), and tightening the registration bucket (already tracked in the G13 candidates).

## F4 — MEDIUM (advisory): denied re-attach + fallback key = targetable pre-squat brick

New with the name key: an attacker can pre-register a *not-yet-installed* popular client's (name, well-known loopback redirect set), eat a denial, and the real client later silently re-attaches to the denied row — permanently stuck on the awaiting-approval page with **no admin signal** (see F2) and no fresh-pending second chance. Recovery is admin-actioned, but the admin has to know to look. This is the D8-inherited policy and I'm not asking for fresh-pending-on-denied — but F2's audit row is what makes this survivable, and the PR body's "Risk" section should name the pre-squat variant explicitly (it currently only covers denied-stays-denied for the *same* client).

## F5 — LOW: race / no unique constraint

Two concurrent same-key registrations can still mint two rows (no unique index was added — consistent with "no migration", and a partial unique on `(name, fingerprint) WHERE software_id IS NULL` would need one). The F6 winner rule (approved > pending > denied, then oldest; `min` is stable over the repo's `created_at, id` ordering) deterministically converges later retries, matching the documented D8 posture. Fine as-is; noting a stray duplicate row lingers until an admin culls it.

## F6 — LOW: test gaps

The 6 new tests do pin the advertised properties (dedupe + single row + single event, order-insensitivity, name-alone ✗, set-alone ✗, read-only hit incl. scopes, approved re-attach, both cross-match directions). Missing: (a) the `""` case (F1); (b) a **fallback-arm** denied re-attach test (the existing denied test is D8/software_id-only); (c) a guard that an *admin-source* row with matching name + fingerprint is never adopted (the `registration_source == 'dcr'` filter is the only line standing between the anonymous door and admin-created rows — worth pinning adversarially).

## Verified

- **Fingerprint semantics (clean):** SHA-256 over `"\n".join(sorted(set(uris)))` — order/duplicate-insensitive, otherwise exact-string (no case/slash/port normalization: divergence can only cause a *missed* dedupe, never a wrong adoption). Same function at create (repo) and dedupe (service). Newline-join is unambiguous — the RFC 3986 charset gate rejects raw control chars on every door. Collisions can't cause adoption: the service re-verifies `set(candidate.redirect_uris) == requested_set` against the **live** row, and the fingerprint is maintained on admin `redirect_uris` updates (repo line ~233). Duplicate URIs are rejected pre-hash.
- **Key-space separation:** correct in code (`IS NULL` vs `==`) and pinned both directions by `test_dedupe_key_spaces_never_cross_match` — modulo the `""` hole (F1). Name matching is consistent at both ends (`.strip()` on register and dedupe; ` Cursor ` ↔ `Cursor` dedupe verified empirically). Pre-3a rows (NULL fingerprint, admin source) can't participate.
- **Kill switches hold:** a dedupe hit never resurrects `active=false`, and `/authorize` + `exchange_code` re-check `active` + `approval_status` live.
- **Gates re-run in a clean worktree at e2cfc4c4:** `make lint` clean (ruff + format + mypy, 1222 files), `test-arch` **153 passed**, `test-unit` **3250 passed**, `test-integration-sqlite` **736 passed, 13 skipped** — all match the PR body. OpenAPI diff is the route description only.
- **Hygiene:** single conventional commit, `fix(auth)` is legitimate (bug live on `main`), no drive-bys; PR body is accurate except the two `""`-falsified claims noted in F1.

**Requested changes: F1 + F2 (both small), plus the F6(a)/(b) tests. Everything else is advisory.**

PR #1262, opened 2026-09-04
fix(auth): include the RFC 6749 §5.1 scope member in every token response

### LINKED ISSUE #1260: Token responses omit the RFC 6749 §5.1 REQUIRED scope member when the issued scope differs from the request
Found during the adversarial review of #1259 (N1, non-blocking there).

RFC 6749 §5.1: the token response MUST include the \`scope\` member when the issued scope differs from the scope requested by the client — it is OPTIONAL only when they are identical.

Our platform *does* downscope: the effective scope of a minted token is the intersection of the client ceiling and the consent grant's scopes (and the D2 triple intersection on the agent channel). But \`TokenResponse\` (src/jentic_one/auth/web/schemas/oauth.py) has no \`scope\` member at all, so a client that requested a broader scope never learns it received a narrower one. Strict clients cache the *requested* scope as authoritative (mcp-remote pins \`requested_scope\`) and may later present tokens for scopes they don't hold, producing confusing 403s downstream instead of an honest downscope signal at mint time.

Fix sketch: add \`scope: str | None = None\` to \`TokenResponse\` and populate it whenever the effective scope != requested scope (always populating it is also spec-legal and simpler). With #1259's \`response_model_exclude_none=True\` the member is cleanly omitted when unset.

Related: #1253 (null-vs-omitted, fixed by #1259); N2 from the same review — introspection \`token_type\` returns \`access_token\`/\`refresh_token\` instead of RFC 7662's \`Bearer\` (harmless today, zero HTTP consumers; fold into the same pass if touched).

Manuel-Jentic:
## Summary

`POST /oauth/token` responses omitted the `scope` member, so clients never learned when the platform downscoped them. RFC 6749 §5.1 makes `scope` REQUIRED whenever the granted scope differs from the requested one — and this platform downscopes routinely: enforcement is the live intersection (live `actor_scope_grants` ∩ client `allowed_scopes` ceiling ∩ consent-grant scopes) recomputed on every resolve by both `TokenService.resolve_access_token` and the broker's `InProcessTokenResolver`, and the agent channel mints from the consent-time D2 triple intersection (`auth/web/routers/authorize.py::_effective_agent_scopes`). Strict clients (mcp-remote pins `requested_scope`) then presented tokens for scopes they were never granted and hit confusing downstream 403s. Every grant leg now reports the effective scope **computed the same way the resolvers enforce it** — reported == enforced by construction.

## Changes

- `auth`: `TokenResponse` (`auth/web/schemas/oauth.py`) gains `scope: str | None` (space-delimited per RFC 6749 §3.3), populated on every leg that mints a non-empty effective set and OMITTED when the set is empty (see Risk below for why).
- `auth`: new shared helper `resolve_effective_scopes()` in `auth/services/token_service.py` — the single source of truth for "what does this token actually enforce" (live grants ∩ client ceiling ∩ grant scopes for non-ephemeral AGENT/SA actors; snapshot ∩ ceilings for USER/ephemeral). Consumed by:
  - `TokenService.resolve_access_token` (the enforcement path — refactored onto the helper),
  - `TokenService.refresh` (reports the live intersection at rotation time, NOT the family's mint-time snapshot, which enforcement ignores for AGENT/SA),
  - `AuthorizeService.exchange_code` agent channel (reports the live intersection at exchange time — a scope revocation landing inside the code TTL is reflected, consistent with the "consent-time snapshot is not trusted across the TTL" posture).
  The broker resolver re-implements the same math over raw SQL by design (no `auth` imports); the grant-channel integration suite pins all three against each other.
- `auth`: jwt-bearer / client_credentials legs report the live `actor_scope_grants` set stamped at mint; the user channel reports the authorize-time snapshot (documented in place: OIDC passthrough scopes are consumed by the id_token, never enforced as platform permissions — standard OIDC posture).
- Token/audit rows are untouched: rotation and exchange still stamp the family/consent snapshot; only the *reported* member changed.
- Regenerated codegen artifacts: `openapi/control/control.openapi.yaml`, `ui/openapi.json`, UI client (`scope?: string`), Go CLI client (`Scope *string \`json:"scope,omitempty"\``).
- Out of scope: `/oauth/mint` (`MintResponse`, caller supplies the exact scopes) and RFC 7662 introspection's empty-string `scope` posture.

## Risk & rollback

- Public API shape change: `/oauth/token` 200 bodies gain an optional `scope` member. Additive and §5.1-legal; verified tolerant in mcp-remote 0.8.3's vendored zod schema (`scope: z.string().optional()`).
- **Empty-set posture (deliberate):** when the effective set is empty the member is omitted — never `""`, which RFC 6749 §3.3's ABNF (`scope-token` is `1*NQCHAR`) forbids. Omission cannot contradict a request at this endpoint: no grant leg's token request carries a `scope` parameter, and the consent flow fails closed on an empty D2 intersection (`no_grantable_scopes`), so "requested scopes, granted none" tokens are not mintable via consent. Empty sets are reachable only for zero-grant agents/SAs (nothing requested) or when every scope grant is revoked post-mint (refresh leg) — the latter is a live, reversible administrative state the platform deliberately distinguishes from revocation (grant revoke / actor disable already fail the exchange closed with `invalid_grant`); erroring `invalid_scope` there would conflate narrowing-to-zero with revocation and break recovery-by-regrant.
- Scope *enforcement* semantics are unchanged (`resolve_access_token` is a pure refactor onto the shared helper; the resolver-parity integration tests pass unchanged). Reporting is point-in-time: grants can still change between rotation and use — inherent, and the resolvers re-derive live either way.
- Revert: revert the squash commit and re-run `make openapi` + client codegen.

## Test plan

- Reviewer's repro pinned in `tests/integration/auth/test_oauth_grant_channel.py`: consent `apis:read apis:write` → revoke `apis:write` → refresh: response reports exactly `apis:read` and BOTH resolvers enforce the same set (`test_refresh_reports_live_narrowed_scope_not_snapshot`); same for a revocation inside the auth-code TTL (`test_exchange_reports_live_narrowed_scope_not_consent_snapshot`). Suite run: `JENTIC_TEST_BACKEND=sqlite uv run pytest tests/integration/auth -m integration` — 89 passed.
- `tests/unit/auth/test_token_service.py::test_refresh_reports_live_agent_scopes_not_snapshot`: agent refresh returns the live set while the minted rows keep the snapshot.
- `tests/unit/auth/web/test_oauth_response_shape.py` (raw-bytes style): every grant leg carries `scope` with the effective value; `test_downscoped_exchange_reports_narrowed_scope_not_requested` pins requested > granted → NARROWED value only; `test_empty_effective_scope_set_omits_the_scope_member` pins empty set → member absent, never `""`/null.
- Full local gate: `ruff format --check` + `ruff check` clean, `mypy` clean (1223 files), `make detect-secrets` clean, `make test-arch` 153 passed, `make test-unit` 3264 passed (coverage 79.36%).

Closes #1260

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -164,21 +164,21 @@
         "filename": "cli/client/generated/control/client.go",
         "hashed_secret": "36c46a098c7543cc8569553d4a8946bc7df395c9",
         "is_verified": false,
-        "line_number": 3851
+        "line_number": 3854
       },
       {
         "type": "Secret Keyword",
         "filename": "cli/client/generated/control/client.go",
         "hashed_secret": "dc0acf6e95f2dcba19616d63c275e8a749126f89",
         "is_verified": false,
-        "line_number": 5134
+        "line_number": 5137
       },
       {
         "type": "Secret Keyword",
         "filename": "cli/client/generated/control/client.go",
         "hashed_secret": "8cf6cb4d35e002c25f800782aab43af23d9e1f2c",
         "is_verified": false,
-        "line_number": 7889
+        "line_number": 7892
       }
     ],
     "cli/client/generated/control/spec.yaml": [
@@ -187,7 +187,7 @@
         "filename": "cli/client/generated/control/spec.yaml",
         "hashed_secret": "5ccd78467ff52d61471c0cf0f32543d0d5d14367",
         "is_verified": false,
-        "line_number": 19568
+        "line_number": 19573
       }
     ],
     "cli/internal/cli/api/admin_providers.go": [
@@ -492,7 +492,7 @@
         "filename": "openapi/control/control.openapi.yaml",
         "hashed_secret": "5ccd78467ff52d61471c0cf0f32543d0d5d14367",
         "is_verified": false,
-        "line_number": 19568
+        "line_number": 19573
       }
     ],
     "release-please-config.json": [
@@ -1712,7 +1712,16 @@
         "filename": "tests/unit/auth/web/test_oauth_client_credentials.py",
         "hashed_secret": "bee7a0aa6017e98bbfbae6b7b1dfb5c64105172b",
         "is_verified": false,
-        "line_number": 135
+        "line_number": 136
+      }
+    ],
+    "tests/unit/auth/web/test_oauth_response_shape.py": [
+      {
+        "type": "Secret Keyword",
+        "filename": "tests/unit/auth/web/test_oauth_response_shape.py",
+        "hashed_secret": "9251c77b74cbfe6c2e1fbf5f7eeb6e14bd1ba37a",
+        "is_verified": false,
+        "line_number": 294
       }
     ],
     "tests/unit/broker/test_execute_credential_name.py": [
@@ -2171,5 +2180,5 @@
       }
     ]
   },
-  "generated_at": "2026-09-04T13:21:33Z"
+  "generated_at": "2026-09-04T15:25:51Z"
 }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -187,7 +187,7 @@
         "filename": "cli/client/generated/control/spec.yaml",
         "hashed_secret": "5ccd78467ff52d61471c0cf0f32543d0d5d14367",
         "is_verified": false,
-        "line_number": 19573
+        "line_number": 19574
       }
     ],
     "cli/internal/cli/api/admin_providers.go": [
@@ -492,7 +492,7 @@
         "filename": "openapi/control/control.openapi.yaml",
         "hashed_secret": "5ccd78467ff52d61471c0cf0f32543d0d5d14367",
         "is_verified": false,
-        "line_number": 19573
+        "line_number": 19574
       }
     ],
     "release-please-config.json": [
@@ -1721,7 +1721,7 @@
         "filename": "tests/unit/auth/web/test_oauth_response_shape.py",
         "hashed_secret": "9251c77b74cbfe6c2e1fbf5f7eeb6e14bd1ba37a",
         "is_verified": false,
-        "line_number": 294
+        "line_number": 299
       }
     ],
     "tests/unit/broker/test_execute_credential_name.py": [
@@ -2180,5 +2180,5 @@
       }
     ]
   },
-  "generated_at": "2026-09-04T15:25:51Z"
+  "generated_at": "2026-09-04T16:04:57Z"
 }
--- a/cli/client/generated/control/client.go
+++ b/cli/client/generated/control/client.go
@@ -2986,8 +2986,8 @@ type TokenResponse struct {
 	IdToken      *string `json:"id_token,omitempty"`
 	RefreshToken *string `json:"refresh_token,omitempty"`
 
-	// Scope Space-delimited effective scopes of the minted access token (RFC 6749 §3.3). Always present (RFC 6749 §5.1): the platform downscopes (client ceiling / consent-grant intersection), so the granted set may be narrower than requested and clients must not assume they got what they asked for.
-	Scope     string  `json:"scope"`
+	// Scope Space-delimited effective scopes of the minted access token (RFC 6749 §3.3), computed the way the platform's resolvers enforce them (live scope grants ∩ client ceiling ∩ consent-grant scopes for agent and service-account tokens), so the granted set may be narrower than requested and clients must not assume they got what they asked for. Present on every response whose token carries at least one scope; OMITTED (never the ABNF-invalid empty string) only when the effective set is empty — reachable solely on legs where the client requested no scopes at the token endpoint (the token request carries no scope parameter, and consent fails closed on an empty intersection).
+	Scope     *string `json:"scope,omitempty"`
 	TokenType *string `json:"token_type,omitempty"`
 }
 
--- a/cli/client/generated/control/client.go
+++ b/cli/client/generated/control/client.go
@@ -2985,7 +2985,10 @@ type TokenResponse struct {
 	ExpiresIn    int     `json:"expires_in"`
 	IdToken      *string `json:"id_token,omitempty"`
 	RefreshToken *string `json:"refresh_token,omitempty"`
-	TokenType    *string `json:"token_type,omitempty"`
+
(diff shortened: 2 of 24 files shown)


Manuel-Jentic (commented):
## Adversarial review — verdict: **NEEDS-WORK**

(Posted as a comment review — GitHub refuses request-changes from the PR author's account.)

Reviewed from a clean worktree at `1c3aa527` (rebased on `25f10b70`). I re-derived the scope *source* vs the scope the minted token actually *enforces* on every leg, re-ran all codegen, reproduced the full local gate, and checked the strict-client story against the actual vendored mcp-remote 0.8.3 bundle. One major finding blocks merge; everything else is clean or minor.

### Findings, by severity

#### 1. MAJOR — refresh leg reports a scope the minted access token does not enforce (agent/SA tokens)

`TokenService.refresh` reports `list(rt.scopes)` — the rotated family's mint-time snapshot. But both live resolvers deliberately **ignore that snapshot** for non-ephemeral AGENT/SERVICE_ACCOUNT tokens and recompute per request:

- `TokenService.resolve_access_token` (`src/jentic_one/auth/services/token_service.py` ~L495-596): *"scopes are resolved live from the actor's current ActorScopeGrant rows rather than the frozen snapshot stored on the token … the token row's `scopes` column is only a mint-time snapshot"* — then intersects live grants ∩ client `allowed_scopes` ceiling ∩ grant scopes.
- Broker `InProcessTokenResolver.resolve_access_token` (`src/jentic_one/broker/repos/token_resolver.py`) enforces the identical live quadruple intersection.

So what the refreshed access token enforces is **live grants ∩ client ceiling ∩ grant scopes at request time** — never `rt.scopes`. With `refresh_ttl_seconds = 604800` (7 days), live grants routinely change across a family's lifetime; immediate-effect scope edits are the platform's own headline design property.

Concrete repro: consent grants agent scopes `a b` → admin revokes the agent's `b` grant (takes effect immediately by design) → client refreshes → response says `"scope": "a b"` but every request with the new token resolves `permissions = [a]` → the strict client (mcp-remote pins the scope it holds) presents `b` and hits exactly the confusing downstream 403 that #1260 was filed about. **The PR defeats its own purpose on the leg where long-lived MCP clients spend their entire lives.** The mismatch runs both directions: narrowed grants → over-report; widened grants → the resolver re-broadens live, so the response *under*-reports.

Fix: in `refresh`, compute the reported set the same way the resolvers do — for non-ephemeral AGENT/SA rows: live `actor_scope_grants` ∩ client ceiling ∩ grant scopes at rotation time (you already hold the session and re-check the grant, client, and actor live in the same transaction); USER rows keep the snapshot. Also update the PR body — *"the response merely reports what was already minted"* is misleading: the row snapshot is dead weight for enforcement on these actor types.

#### 2. MEDIUM — authorization_code agent channel: same class, smaller window

`exchange_code` re-checks grant status, client active/approval, and agent status live at exchange — but **not the agent's live scope grants** — then reports `grant.scopes` (the consent-time D2 intersection). Enforced from the token's first use is `live ∩ ceiling ∩ grant.scopes`. With `auth_code_ttl_seconds = 300`, a scope revocation landing between consent and exchange makes the response over-report at the moment of issuance. Recommend recomputing the intersection with live grants at exchange (one extra query, consistent with the "consent-time snapshot is not trusted across the code TTL" posture the surrounding code already states).

#### 3. MINOR — empty scope `""` is reachable and forced by the REQUIRED field

A freshly approved agent (or SA) with zero `actor_scope_grants` rows exchanges jwt-bearer / client_credentials → `scopes = []` → `"scope": ""`. RFC 6749 §3.3 ABNF (`scope-token *( SP scope-token )`) does not permit an empty value; §5.1 would rather you omit it. Verified the actual consumers tolerate it: mcp-remote 0.8.3's vendored `OAuthTokensSchema` is `scope: z.string().optional()` on a `.strip()` object (checked the shipped 0.8.3 bundle — unknown keys stripped, `""` passes), Go client `Scope string` and UI TS `scope: string` are fine. But a spec-strict third-party validator could reject `""`. Consider `str | None` + omission when empty (rides #1259's `response_model_exclude_none`), or document the `""` posture explicitly in the field description.

#### 4. NIT — user channel reports OIDC passthrough + pre-ceiling snapshot

The user leg reports `auth_code.scopes` verbatim (validated ⊆ allowlist ∪ OIDC at /authorize), while enforcement is snapshot ∩ *live* client ceiling — so `openid`/OIDC scopes are reported but never enforced as permissions, and a mid-flow ceiling narrowing over-reports. Standard OIDC posture and an edge case; a code comment would do.

### Verified clean (no findings)

- **Ordering/determinism**: jwt-bearer & client_credentials are sorted (`ORDER BY scope` in `ActorScopeGrantRepository.list_for_actor`); grant/user channels follow request order and are stable per family (refresh reuses the snapshot). No byte-wise flapping across mints of the same family.
- **Information disclosure**: every leg reports the authenticated caller's own effective set; D11 keeps user identity off the agent channel; no leg reveals another principal's scopes.
- **Strict clients**: mcp-remote 0.8.3 confirmed tolerant from the vendored schema (see finding 3) — no regression from the new member.
- **Codegen honesty**: re-ran `make openapi`, `ui: npm run codegen`, and `GOWORK=off make -C cli generate-api` in a clean worktree — **zero drift** on all three. (The `go.work`→primary-clone caveat that motivates `GOWORK=off` in a worktree is documented in the workspace `docs/how-it-works.md` §9.)
- **Gates reproduced**: `ruff format --check` + `ruff check` clean, `mypy` clean (643 files), `make detect-secrets` clean, `make test-arch` **153 passed**, `make test-unit` **3262 passed, coverage 79.36%** — matching the PR's stated counts. Raw-body shape suite (`tests/unit/auth/web/test_oauth_response_shape.py`): 7 passed, including the downscoped-exchange and space-delimiting pins.
- **Hygiene**: single conventional commit, `fix(auth)` justified (bug exists on main), no drive-bys (`.secrets.baseline` churn is mechanical line-number shifts + the new test file's fake secret), PR body follows the canonical skeleton and its test plan maps to real commands.

### Bottom line

The schema change, codegen, tests, and three of the five legs are solid. But the review bar here is *reported scope == enforced scope*, and the refresh leg — the one strict MCP clients exercise hourly for a week per family — reports a dead snapshot while enforcement is live. Fix findings 1 (and ideally 2), update the PR body's risk framing, and this is merge-ready. **Do not merge as-is.**

PR #1264, opened 2026-09-04
feat(auth): approval-in-flow page for pending OAuth clients on /authorize

Manuel-Jentic:
## Summary

A registered-but-unapproved OAuth client at `GET /authorize` used to park the browser on a static awaiting-approval page — a dead stop. The human saw a dead page, the MCP client never learned when the admin approved, and strict clients (Cursor) treated the stall as failure and spammed re-registrations. This PR implements the **D-approval-in-flow (P2)** decision (Manuel, 2026-09-04 — plans repo, `themes/local-mcp` README, "Authorize-UX optimization wave (2026-09-04 PM)"): the page is now live — it polls for the decision, lets an admin decide inline, and auto-continues the original authorize request the moment the client is approved. No client-side retry, no protocol change on the wire for machine clients.

The re-registration spam itself is fixed separately in `fix/dcr-dedupe-registrations` (related: #1251 — not closed here).

## Changes

- **auth**: `/authorize` with a pending/denied client renders an approval-pending page (same server-rendered style/templating as the consent page). It embeds the ORIGINAL authorize params in a signed state blob — same HMAC signer as the IdP-leg `state`, but a distinct derived key AND `_purpose` discriminator (`approval`), so blobs can never cross legs. TTL 600 s, bound to `iat`, and `_verify_payload` now rejects any blob lacking `iat` outright — the TTL check is not skippable (both mint sites always stamp `iat`, pinned by test). The page self-heals past expiry by re-running `/authorize`, which mints a fresh blob; that re-mint arm is clamped to 3 per tab (sessionStorage, reset by any healthy poll), so a clock-skewed verifier degrades to a terminal message instead of a reload loop.
- **auth**: new `GET /oauth/approval/status?st=<blob>` — anonymous poll, verifies the blob and returns ONLY `{"status": "pending" | "approved" | "denied"}`. No names, redirect URIs, or metadata; a bare `client_id` is never accepted (no registry enumeration). `approved` requires the full D7 gate (approved AND active) so a kill-switched client is never announced ready; a vanished row reads as `pending` (no deletion oracle). Rate limited per-IP in its own bucket namespace like the DCR door (`auth.oauth_rate_limit.approval_status_rpm`/`_burst`, defaults 120/60 — one tab polls at 12 rpm, so the defaults hold ~10 concurrently pending tabs behind one NAT; the page also honors `Retry-After` on 429 with exponential backoff clamped to [5 s, 60 s], so saturation degrades to a slower cadence, never a thundering retry. The knob is documented in `docs/oauth-clients.md` as the operator remedy for larger fleets).
- **auth**: new `POST /oauth/approval/decision` — a thin wrapper for the page's inline admin controls. Authorization is identical to `POST /admin/oauth-clients/{id}:approve`/`:deny` (`get_current_identity(required_permissions=["oauth-clients:write"])`) and the decision is the SAME `OAuthClientService.approve`/`deny` calls (same audit records, same D7 status/active coupling) — no second approval code path. CSRF posture matches the consent POST: no ambient credential is honored; the page must explicitly present the operator SPA's bearer token, which a cross-site form cannot do.
- **page behaviour**: the admin panel renders hidden and is revealed only after page script confirms the same-origin SPA session via `/me` (admin or `oauth-clients:write`). Because the display name is self-chosen at anonymous registration, the panel renders **row-derived, HTML-escaped registration facts** next to the decision buttons — registered redirect-URI origins, `client_id`, and `software_id` when present, plus the approval-queue deep link — mirroring the consent page's origin-as-phishing-counter posture; the details come from the server row, never from request input; anonymous browsers see "ask your admin" plus a copyable deep link to the Settings approval queue (`/app/settings?tab=queue`). On `approved` → client-side redirect re-runs the original `/authorize` → 302 to the IdP. On `denied` → `error=access_denied` redirect to the client's redirect_uri only when that URI is in the client's registered set and parses as http(s); otherwise a rendered terminal message. The initial render stays byte-identical for pending vs denied.
- **hardening**: all dynamic values reach page script through one `\u003c`-escaped JSON `<script type="application/json">` block; the client's self-chosen name only lands in HTML-escaped text nodes; consent-grade security headers (frame-ancestors none, no-store) kept.
- **config**: `auth.oauth_rate_limit.approval_status_rpm`/`approval_status_burst` (+ config-schema and CLI generated struct regen).
- **docs**: new "Approval (self-registered clients)" section in `docs/oauth-clients.md`; `docs/mcp-http-endpoint.md` approval paragraph updated. OpenAPI spec + endpoint reference regenerated (`approvalStatusEndpoint` classified public/signed-state like the OAuth callbacks).
- **Out of scope**: DCR dedupe of duplicate pending rows (`fix/dcr-dedupe-registrations`, #1251); the no-IdP local-login fallback (G12/P3) and the consent-page create-agent CTA (P4), which stack after this per the plan.

## Risk & rollback

- Touches the auth surface's `/authorize` browser leg and adds two endpoints; the machine-client wire contract is unchanged (pending client ⇒ HTML page as before).
- New anonymous endpoint (`/oauth/approval/status`): keyed by signed expiring state, tri-state-only response, own rate bucket — reviewed adversarially for enumeration/oracle/DoS above.
- Deliberate semantic change: a **denied** client can now receive OAuth closure (`access_denied`) via the page's poll, where D7 previously kept deny fully silent. Deny remains reversible; the initial page render still does not distinguish pending from denied.
- **Stale-page window (kill-switch re-arm)**: within the 600 s blob TTL, an admin's still-open pending page can one-click Approve a client that another admin approved and then kill-switched (`:approve` sets both `approval_status` and `active`, per the service docstring). This is admin-gated and identical in power to misusing the existing admin API; the row-derived details panel gives the deciding admin the facts, and the queue remains the source of truth for full history.
- No migrations, no permission changes, no new secrets. Rollback: revert the squash commit.

## Test plan

- `make lint` (ruff + format + mypy strict): clean.
- `make detect-secrets`: clean (baseline line-number drift only).
- `make test-arch`: 153 passed.
- `make test-unit`: 3276 passed (79.4% coverage).
- New `tests/unit/auth/web/test_authorize_approval_flow.py` (24 tests): page render with poll/resume wiring + hidden admin panel + deep link; blob verifies under the `approval` purpose and binds the client; deny-redirect safety (registered URI only); XSS (name escaping + `<`-free JSON block); status endpoint tri-state-only body for all five row states incl. kill-switch and vanished row; 400 on tampered/expired/wrong-purpose/malformed blobs; rate limit 429 + Retry-After in its own bucket (authorize bucket unaffected); decision endpoint 401 anonymous / 403 non-admin / approve+deny delegate to `OAuthClientService` with the caller identity / 400 bad state / 422 unknown action; approved → authorize re-run 302s to the IdP; admin-panel details are row-derived (never request input), escaped, software_id row conditional; purpose separation tested in BOTH directions (approval blob rejected at `/oauth/callback`); blob without `iat` rejected.
- Updated `test_authorize_awaiting_approval.py`: pending/denied render byte-identical under a frozen clock. `test_authorize_router.py`: missing-`iat` blobs now pinned as rejected.
- Manual: both page states rendered and screenshotted via headless Chrome (anon "ask your admin" + copy link; admin Approve/Deny panel) — consistent with the consent page styling.

## Review guide

- Start with the approval-in-flow helpers + endpoints in `src/jentic_one/auth/web/routers/authorize.py` (`_mint_approval_state`, `_approval_tri_state`, `_safe_deny_redirect`, `_render_approval_pending_page`, the two endpoints); the page template/script above them is presentation. Then the tests. Everything else is regenerated artifacts.

Related: #1251 (registration-spam symptom of the same dead-stop; fixed separately by DCR dedupe — this PR removes the dead-stop itself).


Manuel-Jentic (commented):
## Adversarial review — verdict: **NEEDS-WORK**

(Posted as a comment review — GitHub refuses request-changes from the PR author's account.)

Reviewed from a clean worktree at `921193c3` (PR head, `feat/authorize-approval-in-flow` vs `origin/main`). All gates re-run locally; all generator artifacts re-generated and diffed. Focus: the two new anonymous/admin endpoints, the signed-state blob, and every claim in the PR body — with maximum suspicion.

### Findings (by severity)

**M1 — the inline admin decision is made on an attacker-chosen string alone.** The admin panel renders only `{app_name}` — the client's self-chosen name from **anonymous** DCR — and nothing verifiable: no registered redirect-URI origins, no `client_id`, no `software_id`. Attack: register a client named "Cursor" through the anonymous door, lure an admin to the `/authorize` URL (a plain link click; no credentials needed to render the page), and the admin sees "**Cursor** is awaiting administrator approval — you can decide now" with a one-click Approve. The consent page explicitly treats client-claimed names as untrusted and renders the redirect origin as a phishing counter (`_redirect_origin`, "the one client-controlled string the user can actually verify") — the approval page, which guards the *vetting* step for anonymously-registered clients, has no equivalent signal. The Settings queue (which shows full row detail) is deep-linked only in the **anonymous** panel. Impact is bounded (an approved malicious client still needs per-user consent, where the origin IS shown), but this measurably lowers the D7 gate. Fix is small: render row-derived (never request-derived), HTML-escaped registered redirect-URI origins + `client_id` in the admin panel, and/or put the queue deep link in the admin panel too ("review details before deciding").

**L1 — purpose separation is only tested in one direction.** `test_status_endpoint_rejects_wrong_purpose_blob` pins callback-blob → status = 400, but there is no test that an **approval** blob presented at `/oauth/callback` is rejected. Code-wise it holds by construction (distinct derived key `oauth-approval` vs `oauth-state` ⇒ signature fails, plus the `_purpose` discriminator as a second layer — verified both in `_approval_state_key`/`_mint_approval_state` and the callback's `_verify_payload(..., purpose="state")`), but this is exactly the invariant a future refactor could silently break. One test, please.

**L2 — `_verify_payload` skips the TTL check entirely when `iat` is absent** (`if iat is not None:`). Pre-existing helper, not introduced here, and unexploitable today (blobs are server-minted, always carry `iat`, and can't be forged) — but the approval purpose is a good excuse to make `iat` mandatory (missing `iat` ⇒ reject) rather than inherit the footgun into a new anonymous endpoint.

**L3 — page script ignores `Retry-After` on 429.** A 429 falls into the `resp.ok ? … : null` arm and the page just keeps its fixed 5 s cadence — so no thundering retry (good), but also no backoff. Combined with the per-IP key: 60 rpm / burst 30 saturates at ~5 concurrent pending tabs behind one NAT; degradation is graceful (delayed status, buckets refill, knobs exist) but worth knowing at deploy time. Optional: honor `Retry-After` before the next tick.

**I1 (info) — stale-page approve can re-arm a kill-switched client.** Within the 600 s blob window: client approved by admin B, then kill-switched (`active=false`); admin A's still-open pending page (poll now reads `pending` again) one-clicks Approve ⇒ `active=true` re-armed (`:approve` "sets both" per the service docstring's own warning). Admin-gated and identical to misusing the admin API, so no action required — noting for the record.

**I2 (info) — multi-instance clock skew loop.** `age < 0` ⇒ 400 ⇒ the page navigates to `resume_url` ⇒ fresh render ⇒ first poll 5 s later can 400 again if verify lands on the lagging instance — a reload every poll tick, churning the `/authorize` bucket. Same-instance deployments are immune; the pattern is pre-existing on the callback leg. The plain expiry-refresh loop is fine: one `/authorize` reload per 10 min, no interaction with its 30-burst bucket, and a vanished row terminates on the `/error` page rather than looping.

**I3 (nit) — PR body says 17 new tests; `test_authorize_approval_flow.py` collects 18.**

### Adversarial claims verified in code (no findings)

1. **Decision endpoint auth / CSRF**: the only credentials the stack accepts are the `x-jentic-api-key` header and `Authorization: Bearer` (`shared/web/auth.extract_credential`) — no cookie, no basic, no query-param token, anywhere. No CORS middleware exists in `src/` at all, so a cross-site `fetch` can't attach `Authorization` (preflight fails) and a cross-site form can't attach any credential ⇒ 401. The token comes from the SPA's same-origin `localStorage` (`jentic-one.access_token`, matches `ui/src/shared/api/token-store.ts`). **No ambient-credential CSRF is constructible.** The endpoint requires *both* a valid signed blob *and* the `oauth-clients:write` gate (org:admin implies it, enforced server-side via `compute_effective`); state-only unauthenticated decisions are impossible (401 before the handler runs), and an admin-token holder using it is strictly *narrower* than the existing `POST /admin/oauth-clients/{id}:approve` — same `OAuthClientService.approve/deny`, same audit, same permission.
2. **Blob binding**: the blob carries `client_id`; both endpoints read the client only from the verified payload — a blob for client A cannot poll or decide client B. Key derivation differs (`oauth-approval` vs `oauth-state` HMAC info strings) *and* the `_purpose` field is checked; future-dated blobs rejected (`age < 0`).
3. **Status-endpoint oracle claims**: all three verified in code + tests — `approved` requires the full D7 gate (approved AND active; kill-switched ⇒ `pending`), vanished row ⇒ `pending`, initial render byte-identical pending-vs-denied (frozen-clock test). The rate-limit dependency runs **before** blob verification, so 429s don't leak blob validity; the DB lookup runs identically on all tri-state arms (no timing split between arms). Existence-confirmation via blob-minting adds nothing: `/authorize` already distinguishes pending (HTML) from unknown (error redirect) — pre-existing, rate-limited behavior.
4. **Auto-resume**: original params live in the server-built `resume_url` (fixed `/authorize?` prefix, `urlencode`d) inside the JSON block; tampering only affects the attacker's own browser, and the resumed `/authorize` re-validates redirect_uri/response_type/S256/scope-allowlist server-side (verified: `test_resume_after_approval_proceeds_to_idp_redirect`).
5. **Denied semantics**: `_safe_deny_redirect` checks the request URI against the **current row's** `redirect_uris` plus an http(s) scheme parse; the fallback is a static text node — no reflection of the attacker URI anywhere in HTML.
6. **XSS**: every sink checked — `client.name` only in `html.escape`d text nodes; `queue_url` attribute-escaped; page `<title>` static; the JSON block is `\u003c`-escaped (kills `</script>` breakout and `<!--`; entities aren't decoded inside `<script>`, so `JSON.parse(textContent)` round-trips correctly); `deny_redirect` is additionally scheme-checked on both sides. The name never enters the JSON block at all.
7. **Config/codegen/docs**: re-ran `make openapi`, `make config-schema`, `make endpoints` — **zero drift** (`git status` clean). Both config-schema copies + the generated Go struct (defaults 60/30) are consistent; `.secrets.baseline` is line-drift only, as claimed. `approvalStatusEndpoint` is correctly registered in `PUBLIC_OPERATION_IDS` with a rationale; the decision endpoint carries the identity dependency so the route-auth arch test holds.
8. **Gates reproduced**: `make lint` clean, `make test-arch` **153 passed**, `make test-unit` **3270 passed @ 79.40%**, `make detect-secrets` clean — all matching the PR body. The 18 new tests genuinely pin the adversarial properties (tamper/expiry/wrong-purpose/malformed blobs, kill-switch + vanished-row states, tri-state-only body, own rate bucket + Retry-After, 401/403/400/422 on the decision endpoint, escaping + `<`-free JSON) — not just happy paths. Gap: L1 above.
9. **Hygiene**: single conventional commit, no drive-bys, PR body accurate against the diff (modulo I3), matches D-approval-in-flow in the plans repo.

### Ask

M1 (row-derived client detail in the admin panel) and L1 (reverse-direction purpose test) before merge; L2/L3 at author's discretion. Everything else is solid — the threat modeling in this PR is genuinely good, and every security claim in the body checked out against the code.

Issue #1249, opened 2026-09-04
[request] No simple way to find a deployment's broker URL — the UI and API never expose it


Killian-Jentic:
### What's the problem, friction, or need?

There is no simple way for an operator or a user to find out the broker (data plane) URL of a deployment. I had a control-plane URL (`https://<release>.<env>.example.net`) and needed the matching broker URL to configure `jentic`. Nothing in the product would tell me:

- **The UI never shows it.** The MCP setup snippet on the agent detail page prints a literal placeholder instead. From `ui/src/modules/agents/components/detail/McpPanel.tsx`:

  ```
  // On a remote install the broker lives on its own host and is never derived
  // from the control-plane URL — without --broker-url the environment has no
  // broker and `jentic execute` fail-closes (register.go). The UI can't know
  // the broker URL, so the snippet carries an explicit placeholder.
  ```

  The rendered snippet ends in `--broker-url <broker-url>`, with help text telling the user to "ask your operator for the broker (data plane) URL".

- **The API never returns it.** `server.mcp.broker_url` (`src/jentic_one/shared/config.py`) is server-side config only. It is read by `src/jentic_one/mcp/execute.py` and `src/jentic_one/mcp/tools.py`, but no HTTP endpoint exposes it — not `/me`, not `/.well-known/*`, not `llms.txt`.

- **The gateway does not route it.** `deploy/helm/jentic-one/charts/gateway/templates/configmap.yaml` proxies admin, control and registry paths only, so there is no "same host, different path" convention to guess from either.

The result is that the value is only discoverable by asking whoever deployed the instance, or by guessing hostnames. In my case I found it by DNS-probing a `-broker` sibling of the control-plane hostname and confirming with `GET /health` (`{"status":"ok","surface":"broker","version":"0.38.0"}`). That worked, but guessing at DNS names is not a reasonable discovery mechanism, and it only worked because that deployment happened to use that naming convention.

This bites agents as well as humans: an agent that has been given a control-plane URL and a token cannot self-configure `execute`, because the one value it still needs is the one value the platform refuses to tell it.

### What would you like instead? (optional)

The control plane already knows `server.mcp.broker_url`. Expose it read-only so both the UI and agents can discover it:

1. Return it from an existing discovery surface — e.g. a `broker_url` field on `/me`, and/or in `/.well-known/` + `llms.txt` so agents can self-onboard from the base URL alone (the pattern established by #809).
[...]

Issue #1280, opened 2026-09-05
[epic] Upstream seams for native gate integrators


ren-jentic:
## Goal

Make jentic-one fully consumable by **native/desktop integrators** — apps that connect to an arbitrary deployment, sign users in with the standards-track native-app flow, and observe/scope traffic through the broker — via four small, neutral, default-preserving seams. Driving consumer: the macOS menu-bar gate app (third seam consumer after the enterprise overlay), but every seam is generic integrator surface with no product logic.

## Scope — sub-issues

- #1276 — Local-account path for `GET /authorize` (`auth.local_login`, default off): makes RFC 8252 auth-code+PKCE sign-in universal, not IdP-only. (M–L; ships as an XS `AuthService.authenticate` prep refactor + the feature PR.)
- #1277 — `on_operation_not_found` hook on `AppContainer`: supported interception point for broker discovery misses (monitor modes, audit sinks, catalog suggestions). (S)
- #1278 — Identity-scoped `GET /governed-hosts` digest endpoint: minimum-knowledge host set + ETag change-poll, replacing `/apis` enumeration. (M)
- #1279 — Public capability document (`GET /capabilities`): deployment self-description (auth methods, broker URL, features) for one-URL client onboarding; absorbs the discovery half of #1249. (M)

## Rules of engagement

Every PR: default-preserving (no config/wiring change → byte-identical behaviour), own tests, `make openapi` where routes change, seam-table row in `docs/development/extending-jentic-one.md` where a seam is added, compliance base in `jentic_one.testing` where a protocol is added.

## Suggested order

#1277 (smallest, purely additive) → #1278 (independent) → #1276 prep refactor → #1276 feature (security-sensitive, lands with maintainer eyes on the login form/CSRF/consent branch) → #1279 (can land any time; grows a `local_login` flag when #1276 ships).

Manuel-Jentic:
Design note for the #1276 leg, from the local-MCP live-testing wave (the same one that produced #1261/#1262/#1264):

When `/authorize` grows human-facing login (as #1285 does), the resolution order we converged on is an **identity ladder**:

1. **Existing platform browser session** → skip login entirely, straight to consent. Today the SPA session and `/authorize` are two unconnected worlds, and the seam is masked in production only by IdP-side SSO — a logged-in admin connecting an MCP client is asked to authenticate *again*, which consistently confuses people in live testing.
2. No session + IdP configured → IdP hop (today's flow).
[...]

PR #1284, opened 2026-09-05
feat(shared): public capability document (GET /capabilities)

### LINKED ISSUE #1279: [request] Public capability document (GET /capabilities) — deployment self-description for clients
## Motivation

A client connecting to an arbitrary jentic-one deployment (native desktop app, CLI against a remote server, SPA, MCP client) cannot discover what that deployment supports. The information exists, but scattered and incomplete:

| Exists today | Where |
| --- | --- |
| Backend locality, `canonical_base_url`, `mcp_enabled` | `GET /instance` |
| IdP login hint (`{enabled, provider}`) | `GET /auth/idp` — already documented in-code as a *"Public capability hint"*; the SPA login page renders its SSO button from it |
| Auth endpoints + grant types | `GET /.well-known/oauth-authorization-server` (RFC 8414) |
| Mounted-route inventory | `GET /reference/endpoints.json` |

What a client **cannot** learn: the **broker URL** (#1249 — today it's transmitted human-to-human and the CLI fail-closes on remote installs without `--broker-url`), whether the anonymous OAuth-client DCR door is open, whether service accounts / claim tokens / agent self-approve are available, surface composition (`config.apps`), and deployment feature flags in general. RFC 8414 is the *mechanical* truth about endpoints but cannot express product semantics ("password login form exists", "DCR requires approval").

Prior art has converged on exactly this pattern: the MCP authorization spec makes discovery **mandatory** (RFC 9728 `/.well-known/oauth-protected-resource` — which this server already implements for `/mcp`); Matrix clients build their entire login UX from `/versions` + `/capabilities` + `/auth_metadata`; GitLab ships `/api/v4/metadata`. The in-tree `GET /auth/idp` is already a narrow instance of the pattern — this proposal generalises it.

## Proposed shape

`GET /capabilities` — public, unauthenticated, **minimal** (no exact version string; that stays behind the authed `GET /system/version`, per the ASVS fingerprinting posture). A `shared/web/` router mounted next to the instance router in both `create_surface_app` and `create_combined_app`; the broker opts out exactly as it does for `/instance`. Sketch:

```json
{
  "instance": { "backend": "remote", "canonical_base_url": "https://jentic.example.com" },
  "surfaces": ["registry", "admin", "control", "auth"],
  "urls": {
    "broker": "https://broker.example.com",
    "authorization_server_metadata": "/.well-known/oauth-authorization-server"
  },
  "auth": {
    "methods": {
      "idp": { "enabled": true, "provider": "google" },
      "local_login": { "enabled": false },
      "oauth_client_dcr": { "enabled": true, "approval": "manual" },
      "agent_dcr": { "enabled": true },
      "service_accounts": { "enabled": true }
    }
  },
  "features": { "mcp": true },
  "capabilities_version": 1
}
```

- `urls.broker` closes #1249 and enables one-URL client onboarding (broker URL rides the document; config source: the value `server.mcp.broker_url` already carries, or a dedicated advertised-URL key — see open questions).
- `auth.methods` is the login-picker contract: clients render exactly the sign-in options the deployment supports (the Matrix/Element model), instead of probe-and-guess.
- **Extensible by downstream packages** via a contributor registry (the `set_admission_policy` / `set_claim_token_minter` seam shape): OSS defines the route + registry; an overlay registers contributions at import time (e.g. additional `features.*` keys). Route override is not an option by design — `extra_routers` mount after built-ins and never shadow.
- `capabilities_version` lets clients hard-fail on incompatible future shapes.

## Default-preserving guarantee

Purely additive: one new public route; no existing route, schema, or config key changes meaning. Clients treat the document as additive — absent (older server) they fall back to today's probes (`/instance`, `/auth/idp`, RFC 8414).

## Test plan

Golden-pin the body shape (the RFC 8414 discovery test pattern); per-flag toggles (`auth.idp.enabled`, `server.mcp.enabled`, `server.mcp.oauth.enabled`, surface composition) reflected correctly; contributor-registry entry surfaces in `features`; broker app serves no `/capabilities`; `make openapi`.

## Open questions

1. New `/capabilities` route vs. extending `InstanceIdentityResponse` (precedent: `mcp_enabled` was added there)? Proposal: new route — `/instance` stays a pinned identity primitive with its own cache posture.
2. Fully public vs. split (public auth-methods core + authed feature detail)?
3. Does advertising `oauth_client_dcr.enabled` weaken the deliberate 404-unobservability of `server.mcp.oauth.enabled`? (Lean: no — the door's presence is already observable by POSTing to it.)
4. Where does the advertised broker URL come from on multi-node deployments (config key vs. `server.mcp.broker_url` reuse)?
5. Collision rule for downstream-contributed `features.*` keys (prefix? registry validation?).

Closes the discovery half of #1249.

ren-jentic:
## Summary

Implements #1279 (part of epic #1280): one public, unauthenticated `GET /capabilities` document so a client connecting to an arbitrary deployment (native desktop app, CLI against a remote server, SPA, MCP client) can discover — from one URL — which sign-in methods exist, where the broker is, which surfaces are mounted, and which optional features are on. Generalises the in-tree `GET /auth/idp` "public capability hint" into the pattern MCP authorization (RFC 9728), Matrix `/capabilities`, and GitLab `/metadata` converged on. Also closes the discovery half of #1249 (`urls.broker`).

**Based directly on `main`** (which already contains #1282/#1285): the branch was recut after review so it carries only the capabilities work — `git diff main -- src/jentic_one/auth/` is empty. The document publishes the *live* `local_login` capability rather than a hardcoded placeholder.

## What's in here

- **`shared/web/capabilities.py`**: response models, `resolve_capabilities` (config-only, DB-free), and `get_capabilities_router(enabled_apps)` — mounted next to the instance router in **both** app factories; the broker data plane opts out exactly as it does for `/instance`.
- **The login-picker contract** (`auth.methods`): `idp` (mirrors `/auth/idp`), `local_login` (the **effective** offer: `auth.local_login.enabled` AND no external IdP — the `/authorize` flow only reaches the form when no IdP is configured, "IdP always wins", so a raw config echo would advertise an unreachable method), `oauth_client_dcr` (`server.mcp.oauth.enabled` + `approval: auto|manual` from `auto_approve_clients`), `agent_dcr` / `service_accounts` (present iff the auth surface is mounted). Surface-derived flags are documented as describing *the process answering this request*.
- **`urls`**: `broker` comes from the new config key `server.advertised_broker_url` (default `""` → published as `null`). There is **no fallback** to `server.mcp.broker_url` on any backend: the hop URL is topology-private (compose service name, internal listener), so publishing it on an unauthenticated endpoint would be wrong for routing *and* an internal-topology leak. The value is validated http(s)-with-host and stripped of userinfo/query/fragment before publishing; invalid values are dropped with a boot warning. The document also publishes the other public doors — `authorize`, `token`, `agent_registration`, `oauth_client_registration`, and (gated on `server.mcp.oauth.enabled`) `authorization_server_metadata_mcp` + `protected_resource_metadata` — absolute whenever a base URL is known.
- **Contributor seam**: `register_capability_contributor` — same process-global import-time registry posture as `set_claim_token_minter`. Contributors receive a **frozen `CapabilityView`** (never the live `Context`) and run **once at app build time**, never on the request path. Registration validates the contributor (callable, arity, duplicates) and logs the registrant; each contribution is isolated (`try/except` + str→bool key/value validation). Contributions extend `features.*` additively; collisions with built-ins (or earlier contributors) are **logged and dropped**, so a downstream package can never rewrite OSS semantics (open question 5). Seam-table row in `docs/development/extending-jentic-one.md` + `BaseCapabilityContributorComplianceTest` in `jentic_one.testing`.
- **Caching**: strong ETag, `If-None-Match` → 304, `Cache-Control: public, max-age=60`.
- **Versioning**: `CAPABILITIES_VERSION` is tied to the sorted field-pointer set by a shape-pin test; the contract is additive (bumps only on remove/rename/retype; clients branch on key presence).
- **Posture**: fully public but minimal (open question 2) — no exact version string (ASVS fingerprinting; pinned by test). Advertising the DCR flag does not weaken the route gate's 404-unobservability: the door's presence is already observable by POSTing to it (open question 3, issue lean). The request path is log-silent; registration and router build each log one-shot events (`capability_contributor_registered`, `capabilities_resolved`, loopback-broker warning).
- `getCapabilities` in `PUBLIC_OPERATION_IDS`; tagged **Discovery**; `features` is typed `dict[str, bool]` end-to-end (Go client emits `Features map[string]bool`) and ships two built-ins: `mcp` (`server.mcp.enabled`) and `governed_hosts` (true iff the registry surface — which serves #1283's `GET /governed-hosts` — is mounted on the answering process); regenerated `make openapi` / `make endpoints` / `make config-schema` / `cli make generate-config` / UI codegen artifacts all included.

## Default-preserving guarantee

Purely additive: one new public route + one new config key (default `""` publishes `urls.broker: null`). No existing route, schema, or config key changes meaning. Clients treat the document as additive and fall back to today's probes when absent.

## Risk & rollback

- **Risk**: one new unauthenticated read-only route; the document is config-derived, DB-free, and computed at app build, so it cannot touch the DB or block the event loop at request time. Contributors run at boot — a broken contributor fails visibly at startup, not in production traffic.
- **Rollback**: revert the PR; the route disappears and `server.advertised_broker_url` becomes an ignored (still schema-valid) key. No migrations, no data.

## Tests

`tests/unit/shared/test_capabilities.py` (35 tests): byte-shape **golden pin** of the default body (the RFC 8414 golden pattern); per-flag toggles (`auth.idp.enabled`, `auth.local_login.enabled` incl. IdP-wins yield, `server.mcp.enabled`, `server.mcp.oauth.enabled` + `auto_approve_clients`, surface composition); broker-URL validation/stripping and no-fallback gating; contributor seam isolation, registration validation, case-shadowing, build-time freeze; ETag/304; version shape pin; standalone-surface mount; **no `/capabilities` on the broker app**; no version string in the body; schema-visible + public (no BearerAuth) + Discovery tag. Plus the compliance-harness self-test in `tests/unit/testing/test_compliance_oss.py`.

`make lint` (ruff + mypy, 1239 files), `make test-unit` (3418 passed), `make test-arch` (300 passed), `make detect-secrets` green; full Go CLI build+test green with the regenerated client (`GOWORK=off go build ./... && go test ./...`).

Closes #1279


ren-jentic:
## Review response — all seven laps triaged; branch merged with today's `main`

Full disposition of the lap 1–7 findings. The branch was **recut from `main`** after the review (head now carries only the three capabilities commits; `git diff main -- src/jentic_one/auth/` is empty), and the hardening commit `90bea782` addressed the design findings. Today's push (`49be1b4a`) merges current `main` (post-#1370) and fixes the one red check.

### Blocking set (lap 7): all resolved

| Finding | Disposition |
|---|---|
| **A-01 / C-01 / M-02** — branch carried a pre-review copy of #1282/#1285 that reverted main's two-sided `/login` gate; recut required | **Done — recut.** The branch is based on `main` past #1281/#1285/#1286; no `auth/` delta remains. The `--theirs` trap is gone because the conflicting `auth/` files no longer exist on this branch. |
| **I-01** — unauthenticated route reached a contributor seam holding the live mutable `Context` | **Done.** Contributors receive a **frozen `CapabilityView`** (backend, canonical URL, surfaces, mcp flag — nothing writable) and run **once at app build**, never on the request path. Registration validates callable/arity/duplicates and logs the registrant; each contribution is isolated (`try/except`, str→bool validation, case-insensitive first-writer-wins). A hostile or broken contributor can no longer touch config, 500 the route, or block the event loop. |
| **M-01** — `testing/` union with #1281 | **Done.** #1281 is merged and the branch is based past it; `BaseUnregisteredUrlHandlerComplianceTest` and `BaseCapabilityContributorComplianceTest` both present in `__all__`. |

### Majors

| Finding | Disposition |
|---|---|
| A-07/I-04 — `urls.broker` fallback gated on the self-declared `backend` hint; query-string credential survived | **Done.** No fallback to `server.mcp.broker_url` on any backend — only the explicit `server.advertised_broker_url` is published, validated http(s)-with-host, userinfo/query/fragment stripped, invalid → `null` + boot warning. |
| L-05/N-04/A-07 — advertised methods with no endpoints; relative AS-metadata URL; wrong DCR door | **Done.** `urls` now publishes `authorize`, `token`, `agent_registration`, `oauth_client_registration`, and (gated on `server.mcp.oauth.enabled`) `authorization_server_metadata_mcp` + `protected_resource_metadata`; all absolute whenever a base URL is known (canonical URL, else the request origin). The mcp-vs-root registration_endpoint split is documented on the fields. |
| Split-deployment misreporting | **Done.** Field docs scoped to "the process answering this request"; `surfaces` documents that a sibling tier may serve what's absent here. |
| B-01/B-02 — unenforced `capabilities_version` | **Done.** `test_capabilities_version_is_tied_to_the_document_shape` pins the version to the sorted field-pointer set; the contract is additive (bump only on remove/rename/retype). |
| H-04 — per-request collision warning (unauthenticated log amplifier) + silent registration | **Done.** Request path is log-silent; registration and router build emit one-shot `capability_contributor_registered` / `capabilities_resolved`, plus the loopback-broker boot warning. |
| G-04 — no caching headers | **Done.** Strong ETag, `If-None-Match` → 304, `Cache-Control: public, max-age=60`. |
| L-06 — Go `Features map[string]interface{}` | **Done.** `features` is `dict[str, bool]` end-to-end; Go client emits `map[string]bool`. |
| K-03/K-06 — PR body scope claims | **Done.** Body rewritten for the recut branch; the stacking claim is gone. |

### Deferred / not this PR

- **C-03 `features.governed_hosts`** — deferred exactly as lap 7 prescribed ("#1283 must exist for it to be truthful"); #1283 is still open. Fast-follow once it lands.
- **E-01/E-02/E-03 arch-gate erosion** — re-attributed to `main` by the review; now filed as #1375 so it isn't lost.
- **A-12 stale arch exemption** — re-attributed to #1285 on `main` (lap 6, M-03).

### Today's push (`49be1b4a`)

- Merged current `main` (post-#1370 toolkit removal). All four conflicts were generated artifacts, resolved by regenerating from merged source (`make openapi/endpoints/config-schema/config-reference`, `cli make generate-config/generate-api`, UI codegen, secrets baseline). One real semantic touch: main's #1338 added an `/instance` call site for the old private `_sanitized_url_parts` name — repointed to the public `sanitized_url_parts` this branch exports. `/capabilities` keeps its stricter no-fallback posture for `urls.broker`, unchanged.
- Fixed the red **Go CLI lint & test** check: the 1H sensitive-annotation sweep flagged `CapabilitiesUrlsResponse.token` (the RFC 8414 token-*endpoint URL*, not a credential). Rather than exempting every `token` field, the allowlist now accepts schema-qualified entries and the exemption is scoped to `CapabilitiesUrlsResponse.token` only.

**Verification on the merged head:** `make lint` clean (ruff + mypy, 1278 files); `tests/arch` 320 passed; `tests/unit` 3710 passed with one failure — `test_oneshot_cache_is_keyed_on_source_identity_not_path` from #1286, which fails identically on a pristine `origin/main` checkout on macOS (`/dev/fd` semantics; green on Linux CI) — pre-existing, unrelated; `GOWORK=off go build ./... && go test ./...` fully green.

ren-jentic:
Addendum to the review response above: #1283 merged while this was in flight, so the one deferred item (**C-03**) is now in rather than fast-follow — `0336be01` merges the new `main` and adds `features.governed_hosts` (true iff the registry surface, which serves `GET /governed-hosts`, is mounted on the answering process — mount-derived and process-scoped exactly like `agent_dcr`/`service_accounts`). Golden pin, dedicated toggle test, and all generated artifacts extended; PR body updated to match.

The PR is now **MERGEABLE** against `main`. Verified on the final head: `make lint` clean, `tests/arch` 321 passed, `tests/unit/shared/test_capabilities.py` 35 passed, `GOWORK=off go build ./... && go test` green (including the 1H sweep).

PR #1317, opened 2026-09-09
feat(cli): support multipart/form-data bodies in execute

### LINKED ISSUE #1316: [cli] jentic execute cannot invoke operations with multipart/form-data bodies (no file-part flag) — image/file-upload ops are uninvokable
## Summary

`jentic execute` cannot invoke any catalogued operation whose request body is `multipart/form-data`. The command only accepts a **JSON** body (`-d/--data` or `--data-file`), so operations that require a binary file part (image/document upload) are effectively uninvokable through the CLI — even when the API is registered, the credential is bound, and the broker path is healthy.

## Repro

Registered API: `facecheck-id/facecheck-id/v1.02` (host `facecheck.id`), credential bound (`Authorization` header), broker up.

FaceCheck's search flow is two steps:
1. `POST /api/upload_pic` — **`multipart/form-data`**, field `images` (`type: string, format: binary`) → returns an `id_search`
2. `POST /api/search` — `application/json`, requires that `id_search`

Step 2 works fine via the CLI:

```
$ jentic execute op_a5a5a5fd9a2de907f3edb9fa8a89ceeb620c4be7 --broker-scheme http -d '{"demo":true}' --raw
{"id_search":null,...,"error":"id_search parameter is missing","code":"BAD_PARAMETERS",...}
```

(That's a real `200` from FaceCheck — credential + broker routing confirmed working.)

Step 1 has no CLI path:
- `jentic execute op_<upload_pic> --dry-run` produces a request with the correct auth header and URL but **no body** — there is no flag to attach a file part:
  ```json
  { "payload": { "headers": { "Authorization": "[REDACTED]", ... },
      "method": "POST", "url": "http://127.0.0.1:8100/https://facecheck.id/api/upload_pic" } }
  ```
- `-d` / `--data-file` are JSON-only; there is no `--form`, `--file`, or `multipart` equivalent.
- A raw `curl -F images=@file` straight at the broker (`http://127.0.0.1:8100/https://facecheck.id/api/upload_pic`) is rejected `401 unauthorized` — the broker requires a Jentic **caller** token that the CLI mints internally and does not expose, so bypassing `execute` isn't a workaround either.

## Impact

Any operation with a `multipart/form-data` request body cannot be executed via the CLI: image search (FaceCheck), document/attachment uploads, OCR, media ingestion, etc. The operation is discoverable and its credential bindable, but the capability is dead-ended at execution. There is no error either — the request is just silently sent with no body and fails validation upstream.

## Expected

`jentic execute` (and the underlying broker execute contract) should support `multipart/form-data` request bodies — e.g. a repeatable `--form key=value` and `--form-file key=@path` (mirroring `curl -F`), with the broker assembling the multipart payload and still injecting credentials as it does for JSON bodies.

## Notes

- Distinct from the spec-import upload issues (#645, #689, #876), which are about *importing an OpenAPI document* in the UI/registry. This is about *executing* a catalogued operation whose body is multipart.
- Discovered while wiring a real FaceCheck reverse-image-search example; verified on stack `v0.38.2`, CLI `0.32.1`.

sophie-jentic:
## What

Adds `curl -F`-style flags to `jentic execute` so operations with a `multipart/form-data` request body (image/file upload, OCR, media ingestion) become invokable via the CLI:

- `--form key=value` (repeatable) — text field parts
- `--form-file key=@path` (repeatable) — file parts read from disk

Mutually exclusive with `--data`/`--data-file`.

Closes #1316.

## Why it's CLI-only

The broker and the SDK broker transport (`cli/client` `BrokerTransport`) are already **byte-transparent** for the request body and forward the inbound `Content-Type` verbatim; credential injection is header-based and independent of the body. The only blocker was the CLI: it had no way to build a multipart body and `agentops.BuildRequest` hard-defaults `Content-Type: application/json` on any body, which strips a multipart boundary.

## How

- `cli/internal/cli/api/execute.go`: new `--form`/`--form-file` flags; a `buildMultipartBody` helper assembles the body with `mime/multipart` and returns the boundary `Content-Type`. That content-type is **prepended to the header KVs**, so `BuildRequest`'s existing precedence suppresses the JSON default and applies it — and an explicit `--header Content-Type=…` still wins (later headers win the merge). **No change to `agentops`/broker.**
- Mutual-exclusion guard against `--data`/`--data-file`; validation that `--form-file` is `key=@path`.

## Tests (all passing locally)

- `TestExecuteCmdMultipartBody` — builds a multipart body, asserts a `multipart/form-data; boundary=…` content-type and that both a text field and the file part (name + bytes) arrive intact.
- `TestExecuteCmdMultipartRejectsRawBody` — `--form` + `--data` errors with `CodeMissingArgument`, broker not called.
- `TestExecuteCmdFormFileBadSpec` — `--form-file` without `@path` errors.
- `TestExecuteCmdSendsBody` unchanged (raw-body JSON default preserved).

Verified: `go build`, `go vet`, `gofmt`, `golangci-lint` (0 issues), `go test -race ./internal/cli/api/`, plus `tests/arch` and `tests/golden` suites.

## Manual E2E

FaceCheck ID two-step flow (multipart `POST /api/upload_pic` → JSON `POST /api/search`) documented in the plan doc as the canonical end-to-end check.

See `docs/development/plans/cli-multipart-execute.md` for full detail and out-of-scope notes.


ren-jentic (changes requested):
Let me know if you want me to do changes etc and take this off your plate, or you want to do it - happy either way @sophie-jentic

ren-jentic (approved):

PR #1361, opened 2026-09-11
fix(cli): don't block on idle non-TTY stdin in execute (#1354)

### LINKED ISSUE #1354: jentic execute hangs on streamed response body although broker completes the fetch (HTTP 200)
## Summary

`jentic execute` hangs indefinitely on a large/streamed upstream response even though the broker has already fetched the body successfully (HTTP 200). The broker completes the upstream request in well under a second and logs `status=completed`, but the CLI never returns — it holds the response stream open and eventually hits the caller's timeout. The fetched data never reaches stdout.

Reproduced repeatedly this session against Alpha Vantage's `NONFARM_PAYROLL` endpoint (~89 KB JSON body). The **first** call of the session returned normally; every subsequent call hung, which suggests a connection/stream state issue rather than a purely payload-size one.

## Environment

- `jentic` CLI: **0.38.2** (commit `848f2ce0132235b76f19994c88d951f317074f71`, built 2026-09-04)
- jentic-one app + broker image: `ghcr.io/jentic/jentic-one-app:0.32.1`
- Local install (`jenticctl install`), broker on `http://127.0.0.1:8100`, app on `127.0.0.1:8000`
- Platform: macOS (darwin 24.6.0)

## Steps to reproduce

Alpha Vantage imported into the local registry (single `/query` op; NFP selected via `function=NONFARM_PAYROLL`). Broker credential provisioned.

```bash
jentic execute "GET:https://www.alphavantage.co/query" \
  --query function=NONFARM_PAYROLL \
  --broker-scheme http --broker-host 127.0.0.1:8100 --raw
```

**Expected:** the ~89 KB JSON body streams to stdout and the command exits 0.

**Actual:** the command hangs. `--raw`, `--json`, and piping to another process all hang the same way. When piped (`| python3 slice.py`), the downstream reader parses the JSON fine but never sees EOF, so it blocks too — i.e. the body arrives but the CLI never closes the stream.

## Evidence the broker side succeeds

Broker container log for the hung call — upstream returns 200 in 421 ms and the execution is recorded as completed:

```
HTTP Request: GET https://172.66.170.11/query?function=NONFARM_PAYROLL&apikey=*** "HTTP/2 200 OK"
INFO: 172.19.0.1 - "GET /https%3A//www.alphavantage.co/query?function=NONFARM_PAYROLL HTTP/1.1" 200 OK
receive_response_body.started ... stream_id=1
receive_response_body.complete
response_closed.started stream_id=1
response_closed.complete
execution_recorded duration_ms=421 status=completed stream=True ...
```

Note `stream=True` on the recorded execution — the streaming response path looks like the culprit.

## Workaround

Calling the broker directly over HTTP (with an explicit timeout) returns the full body immediately, confirming the data and broker are fine and the hang is CLI-side:

```python
import json, urllib.request
tok = json.load(open("~/.local/state/jentic/<...>_127_tokens.json"))["access_token"]
url = "http://127.0.0.1:8100/https://www.alphavantage.co/query?function=NONFARM_PAYROLL"
req = urllib.request.Request(url, headers={"Authorization": "Bearer " + tok})
body = urllib.request.urlopen(req, timeout=20).read()   # 88,968 bytes, status 200, sub-second
```

## Impact

`severity:major` — `execute` is the primary way to run an operation, and it is unusable for any endpoint that returns a sizeable/streamed body. There is a workaround (bypass the CLI, hit the broker directly), but it defeats the point of the CLI and requires hand-managing the token and broker path.

## Suspected area

The `--raw`/streaming response path in `cli/` — the response body is received and forwarded but the stream is never closed / the CLI never flushes-and-exits, so stdout never sees EOF. Worth checking whether the first-call-succeeds / subsequent-calls-hang pattern points at a reused/leaked connection or an unclosed stream from a prior request.

sophie-jentic:
## Summary

`jentic execute <op>` with no request body hangs forever when stdin is a **non-TTY that never sends EOF** — a backgrounded process, or an interactive agent/harness whose stdin is a socket/pty. This is what made every body-less `execute` in an agent session hang.

## Root cause

`cli/internal/cli/api/execute.go` read the request body from stdin whenever `!term.IsTerminal(os.Stdin.Fd())`. That heuristic is true not only for a real `echo … | execute`, but also for an **inherited, idle, non-TTY fd**. On those, `io.ReadAll(os.Stdin)` blocks on an EOF that never arrives — **before** the broker request is ever built or sent.

Proven deterministically with a local stub broker (no network, no quota): a body-less `execute` hung, and a `SIGQUIT` stack dump showed the main goroutine parked in `io.ReadAll` at the stdin read. The same command with `< /dev/null` returned in 0s. The broker, response streaming/keep-alive, and the Alpha Vantage rate-limit body were all red herrings — the CLI never reached the response.

> Note: the issue title says "streaming" and an earlier comment blamed the broker leg. That diagnosis was wrong and is corrected in [this comment](https://github.com/jentic/jentic-one/issues/1354#issuecomment-5637457752). The earlier broker-leg mitigation is **not** part of this PR.

## Fix

Split the implicit stdin fallback from the explicit one and gate the implicit read on the fd's *shape* via `stdinHasPipedBody`:

- `-d -` (explicit) — read stdin, blocking to EOF. The caller opted in.
- No body flag — read stdin **only** when stdin is a named pipe (`ModeNamedPipe`) or a non-empty regular file. Every other shape (TTY, char device, idle inherited non-TTY fd) is skipped, so `io.ReadAll` can never block.

Preserves the two legitimate body inputs — `echo … | execute` and `execute < file` — while removing the hang on idle inherited stdin.

### Accepted trade-off

An explicit shell pipe (`producer | execute`) still classifies as a body source before the producer flushes, so a producer that never writes would still block — the correct, least-surprising behavior for `|` (same as `cat | grep`), and not the reported bug.

## Testing

- Stub repro, inherited non-TTY stdin: previously hung >120s → now returns in 0s, exit 0.
- `echo -n '{…}' | execute POST:/ok` — body still sent.
- `-d -` with piped data works; `-d -` with no data blocks by design.
- New unit test `TestStdinHasPipedBody` (pipe / non-empty file / empty file / idle fd / directory / nil).
- `go build ./...`, `go vet`, `gofmt`, and the `internal/cli/api` package tests pass.

Fixes #1354

### DIFF
--- a/cli/internal/cli/api/api.go
+++ b/cli/internal/cli/api/api.go
@@ -10,7 +10,6 @@ import (
 	"os"
 	"strings"
 
-	"github.com/charmbracelet/x/term"
 	"github.com/spf13/cobra"
 
 	"github.com/jentic/jentic-one/cli/client"
@@ -171,7 +170,23 @@ func emitAPIResponse(out io.Writer, resp *http.Response, failOnError bool) error
 // --data-file. Returns nil when there is no body.
 func resolveAPIBody(opts *apiOptions) (io.Reader, error) {
 	switch {
-	case opts.data == "-" || (opts.data == "" && opts.dataFile == "" && !term.IsTerminal(os.Stdin.Fd())):
+	case opts.data == "-":
+		// Explicit stdin body (`-d -`): the caller opted in, so block until EOF.
+		data, err := io.ReadAll(os.Stdin)
+		if err != nil {
+			return nil, fmt.Errorf("reading stdin: %w", err)
+		}
+		if len(data) == 0 {
+			return nil, nil
+		}
+		return bytes.NewReader(data), nil
+	case opts.data == "" && opts.dataFile == "" && stdinHasPipedBody(os.Stdin):
+		// #1354: implicit stdin fallback (no body flag), same contract as execute.
+		// Only read when stdin is a pipe or non-empty regular file (a real
+		// `echo … | jentic api`). An idle, inherited non-TTY fd (backgrounded
+		// process, or an agent/harness whose stdin is a socket/pty with no EOF)
+		// must NOT be read: io.ReadAll would block forever, hanging every
+		// body-less `jentic api`. The old !IsTerminal heuristic read those too.
 		data, err := io.ReadAll(os.Stdin)
 		if err != nil {
 			return nil, fmt.Errorf("reading stdin: %w", err)
--- a/cli/internal/cli/api/execute.go
+++ b/cli/internal/cli/api/execute.go
@@ -14,7 +14,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/charmbracelet/x/term"
 	sdkclient "github.com/jentic/jentic-one/cli/client"
 	"github.com/jentic/jentic-one/cli/internal/agentops"
 	"github.com/jentic/jentic-one/cli/internal/cli/clictx"
@@ -253,7 +252,23 @@ func (a *app) executeE(cmd *cobra.Command, opts *executeOptions, target string) 
 		}
 		body = mpBody
 		multipartContentType = ct
-	case opts.data == "-" || (opts.data == "" && opts.dataFile == "" && !term.IsTerminal(os.Stdin.Fd())):
+	case opts.data == "-":
+		// Explicit stdin body (`-d -`): the caller opted in, so block until EOF.
+		data, readErr := io.ReadAll(os.Stdin)
+		if readErr != nil {
+			return fmt.Errorf("read stdin: %w", readErr)
+		}
+		if len(data) > 0 {
+			body = bytes.NewReader(data)
+		}
+	case opts.data == "" && opts.dataFile == "" && stdinHasPipedBody(os.Stdin):
+		// #1354: implicit stdin fallback (no body flag). Only taken when stdin is a
+		// pipe or regular file that actually carries data — a real `echo … |
+		// execute`. A non-TTY stdin that is merely INHERITED and idle (a
+		// backgrounded process, or an interactive agent/harness whose stdin is a
+		// socket/pty with no EOF) must NOT be read: io.ReadAll would block forever
+		// there, hanging every body-less execute. That was the bug — the old
+		// heuristic keyed on !IsTerminal alone, which is true for those idle fds.
 		data, readErr := io.ReadAll(os.Stdin)
 		if readErr != nil {
 			return fmt.Errorf("read stdin: %w", readErr)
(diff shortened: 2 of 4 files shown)


ren-jentic:
## Deep review — findings fixed in 65e9271…abf271c

The core fix is correct and well-diagnosed: gating the implicit stdin read on the fd's shape (`stdinHasPipedBody`) is the right call, and the explicit/implicit split preserves both legitimate body inputs. Three issues found and fixed on the branch:

### 1. The end-to-end regression test was a no-op (critical, test)

`TestResolveAPIBody_IdleNonTTYStdinDoesNotHang` used `/dev/null` as the "idle non-TTY stdin" — but a char device **EOFs instantly**, so `io.ReadAll(/dev/null)` returns immediately. The pre-fix `!term.IsTerminal` + `io.ReadAll` code also passes that test; it could never detect a reintroduction of the #1354 hang. Verified empirically: with the `stdinHasPipedBody` gate removed, the `/dev/null` test still passes.

**Fix:** unix-tagged socketpair tests (`stdinbody_unix_test.go`) whose stdin is open, dataless, and never EOFs — the exact shape an agent/harness or backgrounded process hands the CLI. With the gate removed they fail bounded at the 5s timeout guard (verified). The `/dev/null` test is kept, renamed to what it actually pins (`…IdleCharDeviceStdinHasNoBody`), as the portable no-body check.

### 2. Body-resolution logic duplicated 4× (moderate, maintainability)

The explicit + implicit read blocks were copy-pasted in both `execute.go` and `api.go` — the same duplication that let the first commit miss `jentic api` entirely (Copilot's critical finding). **Fix:** shared `stdinHasPipedBody` + `readStdinBody` in `stdinbody.go`, and execute's inline switch extracted into `resolveExecuteBody`, which also makes execute's body path drivable end to end by tests (addresses Copilot's remaining coverage comment — both resolvers now have blocking-stdin and piped-stdin tests).

### 3. `docs/development/fix-1354-execute-streaming-hang.md` doesn't belong in the repo (convention)

`docs/development/` holds evergreen developer docs (local-setup, releasing, …). A per-issue fix narrative — with branch names, review-round narration, and an `execute.go:258` line pin that rots — belongs in the issue/PR record, where this diagnosis is already fully captured. Removed; the comments in the code carry the present-tense invariant with the `#1354` pointer. Change-narrating comments ("the old `!IsTerminal` heuristic read those too") were also rewritten as present-tense invariants.

### Noted, not changed

- A **socket stdin that genuinely carries a body** (e.g. `echo body | ssh host jentic execute …`, where sshd hands the remote command a socketpair) is no longer read implicitly — indistinguishable from the idle harness fd. `-d -` is the explicit opt-in for that case. Right trade-off vs. hanging forever, just worth knowing.
- Behavior notes preserved: `-d -` with empty stdin still yields no body; `producer | execute` with a never-writing producer still blocks (correct for an explicit pipe).

### Verification

`go build ./...`, `go vet ./...`, `gofmt`, `golangci-lint run` (0 issues), and the full `go test ./...` suite (30 packages) pass; the socketpair tests fail bounded when the gate is reverted.

PR #1368, opened 2026-09-11
[NOT_YET] feat!: drop the toolkit tables behind acknowledgement gates (theme 5 phase 6b)

ren-jentic:
## Theme 5, Phase 6b — the acknowledged drops

Stacked on #1362 (phase 6a). The final phase of the toolkit removal (plan: jentic-one-plans `themes/theme-5-remove-toolkits.md`).

> ⚠️ **Merge gate (process, not just code)**: the plan gates 6b on operators running the 6a job against production data — `flatten-toolkits --verify --acknowledge` — plus the staging rehearsal against a production-shaped dataset. The migrations enforce this at run time (below), but this PR should also **wait for that acknowledgement before merging**, and per the release-sequence it belongs to the release AFTER the main cut (the deprecation window).

### Drop migrations — guard-and-raise, never guard-and-skip (hard problem 5)
- **Control**: drops `toolkit_permission_rules` → `toolkit_keys` → `toolkit_credential_bindings` → `toolkits` (reverse-dependency order, mirroring the `f7a8b9c0d1e2` downgrade). Gate: all four empty (fresh installs, the CI case) OR a `toolkit_flattening_acks` sentinel row; otherwise raises with the operator runbook. **Postgres ordering guard**: raises if any enterprise FK still references `control.toolkits` — the enterprise migration (enterprise#82) applies first.
- **Admin**: drops `agent_toolkit_bindings` (gate: legacy rows with zero direct bindings → raise) and sweeps retired `toolkits:*` scopes from stored grants (scalar + JSON columns, both dialects).
- Downgrades recreate empty tables; **rollback = downgrade + `export-toolkits --import`** — drilled in integration on both dialects.

### Execution history survives (hard problem 4 leftover)
`execution_records.toolkit_name` added (additive migration) and backfilled by the 6a flattening job (`--verify` fails on unbackfilled rows); `resolve_toolkit_names` deleted — the `GET /executions` post-drop 500 hazard is closed. `GroupBy.TOOLKIT` keeps grouping the frozen `toolkit_id` column.

### Broker
`direct_bindings_enabled` (flag + entire flag-off derivation path), `Jentic-Toolkit-Id` (headers + spec), `ambiguous_toolkit`/`no_toolkit_binding` directives, `TOOLKIT_*` events (no live emitters), and **`jntc_live_` plaintext acceptance** all deleted — the phase-4 deprecation window closes here (401; holders use their `sak_` key). Key-retirement job/CLI/startup die with their source table.

### ORM vs the rollback path
All 5 toolkit models + 6 repos deleted; `export-toolkits` / `flatten-toolkits` converted to migration-independent raw SQL so they still work against tables recreated by a downgrade.

### UI
The 5d deprecation-window remnants deleted on schedule (`/app/toolkits*` aliases + toast, Monitor `?toolkit_id=` scrub, redirect e2e spec); API client regenerated for the reshaped control spec.

### Deliberate `Toolkit` survivors
Historical migrations, the 6a flattening/export/ack machinery (rollback path), `execution_records.toolkit_id/name`, `GroupBy.TOOLKIT`, `ActorType`/`AuditTargetType` legacy enum members (S-03: history stays queryable), `RETIRED_SCOPES`.

### Validation
- ruff + mypy clean · unit/arch/web **3925 ✓** · integration SQLite **799 ✓** · Postgres **1131 ✓** · Go CLI build/test ✓ · UI tsc/lint/vitest **1129 ✓**
- New `test_phase6b_drop_migrations.py`: empty-DB migrate, block-without-ack, proceed-with-ack, admin gate, and the downgrade + re-import drill.
- Pre-existing note: two unmarked `test_verify.py` failures exist on the base commit too (outside CI's `-m integration` gate) — untouched.


Manuel-Jentic:
@ren-jentic, could we agree an order between this PR (theme-5 6b) and **theme-8 Phase 4** (epic #1379, dropping the service-account tables)? The two overlap in a few places.

**Where things stand.** Theme-8 Phase 1 (#1386) is merged into the integration branch `feat/credentials-and-approvals-overhaul`, on top of theme 7. It migrates every service account to a successor agent and copies the key digest. Phase 2 (#1403, removing the SA surface) and an L2 boot-ordering fix (#1402) are open against that branch. Phase 4 hasn't started. Like 6b, it waits on a production run followed by `--verify --acknowledge`.

**Overlaps:**
1. **`toolkit_flattening_repo.py:169,216-217`** reads `service_accounts` (`_LIST_SERVICE_ACCOUNT_IDS`) when building the known-actor set. Once theme-8 Phase 4 drops that table, this query must be retargeted or removed. That's simpler if 6b lands first and Phase 4 edits the survivor.
2. **Key retirement.** 6b removes the key-retirement job, its CLI and its startup hook. On the integration branch, theme-8 Phase 1 changed that job to mint **agents** instead of SAs. #1402 also runs it in sequence with the SA migration at boot, in a single `_start_boot_migrations` task in `app_factory.py`. So 6b will conflict with #1402 and needs to keep the SA-migration step when it removes the key-retirement step.
3. **The `jntc_live_` deprecation message.** 6b tells holders to "use their `sak_` key". After theme-8 Phase 1, a `sak_` plaintext still authenticates, but as the successor agent. No new `sak_` keys are issued after Phase 2, so the long-term advice is the successor agent's `jak_` key.
4. **Base branch.** This PR targets `feat/theme-5-phase-6a-flattening`, which was branched from `main`, while theme 7 and theme 8 are on the integration branch. Whichever of the two reaches `main` second will need a rebase across `app_factory.py`, `__main__.py`, `RETIRED_SCOPES` and the regenerated OpenAPI and client artifacts.

**Proposal (the order the theme-8 plan prefers):**
- 6b lands first, once its acknowledgement gate is met.
- Theme-8 Phase 4 then rebases onto it: it retargets item 1 and drops SA tolerance from what 6b leaves behind.

If 6b is going to wait longer than theme 8, the alternative is that Phase 4 neutralises the flattening query itself, and 6b rebases over it. Which order works for you? I'm happy to take the rebase on the theme-8 side either way.

ren-jentic:
Heads-up from [#1411](https://github.com/jentic/jentic-one/pull/1411), which adds post-migration upgrade steps to `python -m jentic_one.migrations.run` (toolkit-key retirement, plus a one-shot toolkit flatten recorded in a new control `upgrade_steps` table).

When this PR rebases onto it:

- **Delete the two theme-5 steps** in `src/jentic_one/control/services/upgrade_steps.py` (`theme5_retire_toolkit_keys`, `theme5_flatten_toolkits`). They import `KeyRetirementService` / `ToolkitFlatteningService`, which this PR removes. Keep the `UpgradeStepService` / `upgrade_steps` ledger mechanism itself; the next release's data steps will use it.
- Also remove their tests in `tests/unit/test_migration_upgrade_steps.py` and `tests/integration/control/test_upgrade_steps.py`, and the `KEY_RETIREMENT_LOCK_KEY` use in `control/services/run_lock.py`.
- No ordering hazard in the meantime: #1411's steps report `skipped` (`toolkit_tables_absent`) once your drop migrations have run, so a deployment that picks up both never errors.
- #1411 also adds admin revision `5c7e2a9d4f16` on top of `b9d0e1f2a3b4`. Your admin drop migrations need to re-parent onto it.

ren-jentic:
@Manuel-Jentic agreed: **6b lands first**, and theme-8 Phase 4 rebases onto it. Thanks for offering to take the rebase on that side. Point by point:

1. **`_LIST_SERVICE_ACCOUNT_IDS` in the flattening repo.** It stays in 6b as-is. Phase 4 can retarget or remove it on the survivor. After 6b, the flattening job is rollback-path machinery only, run against tables recreated by a downgrade, so Phase 4 can simplify it freely.
2. **Key retirement.** 6b now deletes the job, its CLI, the boot task and the #1411 post-migration step, and keeps the upgrade-step mechanism with an empty registry. When #1402 reaches `main`, the rebase keeps the SA-migration step in `_start_boot_migrations` and drops only the key-retirement half. I'm happy to do that rebase on the 6b side if 6b is still open then.
3. **`jntc_live_` advice.** 6b now **keeps** `jntc_live_` acceptance until the published date (no earlier than 2026-12-01), instead of closing the window. The deprecation warning now says to rotate to a key minted for the migrated account, without naming the `sak_` prefix. Once Phase 2 stops issuing `sak_` keys, please update that string to the successor agent's `jak_` key. It's one line in `shared/auth/api_key_resolver.py`, plus the `jntc_live_` row in `openapi_meta.py` and the broker spec.
4. **Base branch.** 6b is rebased onto #1411 (`fix/0.40-upgrade-safety`) and now targets it. It will go to `main` once #1411 merges, so it'll be on `main` before the integration branch. The integration branch then rebases over `app_factory.py`, `__main__.py`, `RETIRED_SCOPES` and the regenerated specs, as you proposed.

PR #1371, opened 2026-09-14
feat(credentials): agent-driven OAuth 2.0 connect flows (device_authorization + authorization_code)

DavidAtJentic:
## Summary

Introduces agent-driven OAuth 2.0 connect flows. Agents (or users) call `POST /integrations:connect` against a config-seeded vendor registry, the platform runs a state machine through either `RFC 8628` device authorization or `RFC 6749` authorization code, and the associated API is auto-imported into the catalog. This change would support a curated set of public API clients shipping with the service as well as eventually integrations that support Dynamic Client Registration or any other dynamic/procedural client format.

## Changes

- **control**: new `/integrations:connect`, `/connect-sessions/{id}[:confirm|:cancel|/status]`, `/vendors`, `/vendors/{key}/auth-capabilities` endpoints. New `ConnectSessionService` state machine + `AuthFlowHandler` protocol with `DeviceAuthorizationHandler` and `AuthCodeFlowHandler` implementations. Server-driven polling via a new `ConnectPollScanner` background job. Shared `/credentials/oauth/callback` route routes to the standalone-credential or connect-session path via a `sid` state-JWT claim; replay protection lives in a shared prologue used by both entrypoints.
- **shared/config**: new `vendors.entries.<slug>` registry surface — per-vendor flows (device_authorization / authorization_code), scope catalog with read/write classification, identity probe.
- **migrations**: new `connect_sessions`, `device_authorization_credentials`, `agent_credential_permissions` tables; `credentials.state` column added with backfill.
- **ui**: `VendorConnectFlow` (self + approve modes) + `DeviceCodeConnectDialog`, wired into the credentials dialog. Agent's `reason` free-text surfaces on the approve page.
- **security hardening (baked in)**: callback nonce replay guard, SSRF `validate_upstream_url` on five vendor-facing httpx calls, https-only guard on six vendor-URL navigation sites, session-id enumeration oracle closed on `/status` + `:cancel`, per-actor rate limit on `:connect`, `navigator.sendBeacon` for tab-close cancel, boot-time vendor-shape validator.
- **observability**: three metrics on the connect flow — `sessions_created_total`, `sessions_terminal_total` (vendor + flow + outcome labels), `time_to_connected_seconds`.
- **out of scope**: agent-credential bindings replacing toolkit membership (theme-5), managed-vendor variants (Pipedream-style).

## Risk & rollback

- **Migration adds tables + a NOT-NULL column with backfill** on `credentials`. Reversible via `alembic downgrade`; empty tables + column drop.
- **New config surface** (`vendors.entries`) is off-by-default — an empty registry disables the flow entirely. No changes required to existing deployments.
- **New scope** `credentials:connect` — narrower than `credentials:write`, additive to the permission catalog.
- **Rollback**: revert the branch and downgrade one migration; no data written to the new tables outside connect sessions.

## Test plan

- `make check` passes (lint, mypy, arch tests, secrets audit).
- Backend: ~440 unit + 13 integration tests across the connect subsystem; new suites for SSRF guards, vendor validator, rate limit, safe-navigation, callback replay refusal.
- UI: 129 credentials tests (component + hook + safe-navigation).
- `make test-arch`: 150/152 pass; the two remaining failures are the vendored-orm-facts drift pending on a companion rules-repo PR (`chore/orm-facts-add-device-auth-credentials`).
- Manual: end-to-end device-flow connect against a real GitHub OAuth App verified locally.

## Review guide

Start with `src/jentic_one/control/services/integrations/connect_session_service.py` (state machine + finalise sequence) and `flow_handlers/base.py` (`AuthFlowHandler` protocol). Then the two handlers, then `web/routers/integrations.py`. UI review starts at `ui/src/shared/credentials/components/VendorConnectFlow.tsx`. Security and quality fixes surfaced during local agent reviews are landed as their own commits.

## Checklist


Manuel-Jentic:
Rebased this branch onto `main` (was 72 behind / `CONFLICTING` after theme 5 landed) and force-pushed — @DavidAtJentic please review the resolution rather than redoing it. **All 28 of your commits are preserved in order and untouched in message/structure**; the rebase adaptation lives in 3 trailing commits so it's easy to audit (and to drop/redo if you disagree):

- `2efb8d3a` `refactor(credentials): capture connect approvals as direct agent-credential bindings` — **the one judgment call that needs your eyes.** Your `AgentCredentialPermission` table was an explicit stand-in for theme 5; theme 5 landed with a different shape, so the table/repo/DDL are gone and `confirm()` now writes `AgentPermissionRuleRepository.replace_user_rules` (control DB) then `EffectsRepository.bind_agent_to_credential` (admin DB, idempotent). `_mark_terminal()` sweeps the binding via a new `unbind_agent_from_credential` mirror so no ghost binding outlives a deleted pending credential. Your approve-page wire contract (`{method, path-glob, effect}`) is untouched — globs are translated to canonical regex rules at persistence time only.
- `63f6d73b` — pure regen (OpenAPI, UI/CLI clients, endpoints docs, config schema, secrets baseline).
- `1ab323d5` — test alignment: permission-catalog enumerations + your migrations re-parented onto main's control head (`u2c3d4e5f6a7`; your two revision ids collided with theme-5 migrations that took them, so they're re-id'd `v3d4e5f6a7b8`/`w4e5f6a7b8c9`), plus a SQLite variant for the `requested_scopes` migration.

Verification post-rebase: 4427 unit/web/arch passed (82.7% cov), 179 control integration, lint/mypy clean, CLI `go build`+`go test` green, UI build + 1197 tests green, `cz check` conforms for all 31 messages.

Two notes:
1. **Companion PR**: jentic/jentic-one-rules#28 adds the `device_authorization_credentials` ksuid exemption to the mounted rules source (your vendored copy already has it) — without it the two arch tests fail on machines with the rules mount. Merge it with/before this PR.
2. **Known crash-window** (pre-existing posture, flagged not fixed): if the process dies between the control-DB rule write and the admin-DB bind at confirm, rules exist without a binding until manual reconcile — same posture the access-request flow accepted.

Context for the urgency: this PR is the head of the theme-7 chain (epic #1374, Friday target) — phase 1b (`request_connection` + `jentic connect`) starts from your `POST /integrations:connect` surface as soon as this merges.

Manuel-Jentic:
FYI for prioritization: theme 7's remaining work is staged behind this PR — Phase 1b (`request_connection` MCP tool + `jentic connect`) and the enterprise connect-sessions list (gates jentic-one-enterprise#84) both build on these connect flows. The rest of the theme-7 cascade (#1376 → #1378 → #1377 → #1373, epic #1374) is review-ready independently. No rush intended — just full information for sequencing.

Manuel-Jentic (changes requested):
Ran a deep review pass on the rebased head (`5fe4fbdb`): four focused review laps (security, backend lifecycle, frontend, integration/generated-artefacts) plus a hands-on tour of the running app (SQLite profile + Playwright). Overall the architecture is genuinely strong — signed one-shot actor-bound state, keyset encryption end to end including refresh, SSRF guards with their own test suite, RFC 8628 handled to the letter, and the parity-pinned rule matcher is the right design. The asks below are concentrated in the session lifecycle, the approval surface, and generated artefacts. Items marked **[verified live]** were reproduced on the running app.

## Blockers

**1. Device sessions stick in `polling` forever once the vendor code expires.** `ConnectPollScanner._due_credentials` filters to `device_code_expires_at > now` (`shared/jobs/connect_poll_scanner.py:139-143`), so a row whose vendor TTL lapses between ticks is never selected again — the handler's `expired` branch and the session-TTL guard are unreachable through the only driver. `/status` returns `pending` forever. Fix: drop the `> now` filter (let `advance` report `expired`) or add a second sweep for rows past expiry.

**2. `created` and abandoned auth-code sessions have no TTL at all.** Session TTL only lives inside `advance_polling_session`. Close the browser before `:confirm` (state `created`), or abandon the vendor popup on the auth-code path (state `polling`, no aux row), and the session plus its orphan `state="pending"` credential row leak forever. **[verified live]** — a handful of abandoned wizard attempts left six identical "GitHub" credential cards with no pending badge, plus duplicate catalog auto-imports (one dead-lettered) since every `:connect` enqueues one. Fix: a flow-agnostic sweep over `state IN ('created','polling') AND created_at < now() - TTL`, distinct rendering for pending cards, dedupe the auto-import.

**3. Unguarded `_mark_terminal` can delete a live connected credential.** `connect_session_service.py:996-999` never re-checks the row is non-terminal. Reachable via (a) callback replay — the state JWT nonce is not consumed on the error path, so replaying the callback URL with `error=access_denied` after a successful connect deletes the live credential, vaulted token, and binding; (b) a multi-pod scanner race (no poll mutual exclusion). Fix: compare-and-swap (`UPDATE … WHERE state IN ('created','polling')`, bail on rowcount 0) — one line of defence covers all callers.

**4. No ownership scoping on `GET /connect-sessions/{id}` / `:confirm`; unvalidated `agent_id`.** `routers/integrations.py:182-281` applies no access filters, initiator check, or poll_token gate — any actor with `credentials:write` can read any session's review data and confirm any session, binding an arbitrary `agent_id` (never validated for existence/ownership). Session IDs travel in approval URLs, so they aren't secrets. Fix: gate on initiator/owner identity per the query-scoping conventions and verify agent ownership before binding.

**5. SQLite backend broken by the new migration. [verified live]** `x5f6a7b8c9d0_add_requested_permission_rules_to_connect_sessions.py:29-36` uses raw `postgresql.JSONB` + `server_default=sa.text("'[]'::jsonb")`, and the ORM (`control/core/schema/connect_sessions.py:11,66,74`) imports `JSONB` directly — `make migrate-sqlite` dies with `UnsupportedCompilationError`, taking `config/local-sqlite.yaml` and the SQLite integration lane with it. Your own `v3d4e5f6a7b8` did this right with `.with_variant(sa.JSON(), "sqlite")`; `w4e5f6a7b8c9` even documents the SQLite support. Fix: mirror that precedent + use `json_variant()` in the model.

**6. Go CLI client regen missing → `check-cli-gen` should be red. [verified]** The regen commit skipped `cli/client/generated/control/`: vendored `spec.yaml` still contains `PermissionRuleModel` (3×) which `openapi/control/control.openapi.yaml` deleted (0×), and `client.go` still types `ConfirmSessionRequest.PermissionRules` as `*[]PermissionRuleModel`. Fix: `make -C cli generate-api`.

**7. Rules editor silently widens agent-requested grants.** `RuleListEditor.tsx:336-354` — the edit draft only carries `effect/methods/path/match_mode`; a requested rule with `operations: ["repos/delete"]` that the human merely touches is saved **without** the operations constraint, and `RulePreviewRow` never renders `operations`/`comment`, so the approver can't see what they're dropping. In an approval surface this is the failure class the page exists to prevent. Fix: round-trip and render `operations`/`comment`.

**8. The "no ops affected" preview is unseeded-random.** `template-matcher.ts:126-142` samples 20 unseeded `RandExp` strings (the docstring at `:67-69` claims seeding that isn't there) — the verdict flips between renders and frequently reports "grants nothing" for regex rules that grant plenty. Fix: structural segment-wise intersection, or at minimum seed + memoize per (rule, template).

**9. Wizard dead ends. [verified live]**
- Retry: `VendorConnectFlow.tsx:406-411` resets phase but never clears `connectFiredRef`, so `:connect` never re-fires and Continue stays disabled forever.
- Scope-less vendors: Continue is disabled at `selectedScopes.size === 0` (`:489,834`) while the empty state promises "the connection will use the vendor's defaults" — such vendors can never be connected.
- No agents: user-initiated connects hard-gate on an agent existing ("Create one first, then come back", no link, Continue disabled) even though `ConfirmSessionRequest.agent_id` is optional server-side.

**10. Vendor upstream 4xx escapes `:confirm` as a raw 500. [verified live]** A valid confirm whose device-authorization `begin` fails at the vendor raises `DeviceAuthorizationUpstreamError` unmapped (`routers/integrations.py:249` → `connect_session_service.py:488` → `device_authorization.py:105`); the human sees "Internal Server Error", and retry 500s again. Repro'd with a vendor 404. Fix: map to a 502/424-style problem envelope with a retryable hint.

## Security asks (pre-merge)

- **Rate limiter guards the wrong endpoint**: the `:connect` limiter's stated purpose (`routers/integrations.py:56-67`) is protecting the vendor, but the vendor call fires at `:confirm`, which is unlimited — and a failed `begin` leaves the session `created`, so `:confirm` is retryable in a tight loop during a vendor incident. Extend the same bucket to `:confirm`.
- **Vendor body in `DeviceAuthorizationUpstreamError.__str__`** (`device_authorization.py:34-37`): a vendor answering the poll with a form-encoded 200 (`access_token=gho_…`) fails `.json()` and puts a live token into the exception message — one logger away from tokens in logs. Carry the HTTP status only (the auth-code handler already does this deliberately, with a great comment).
- Fast-follows: PKCE (S256) on both auth-code paths; ReDoS guard for regex rules (1000 chars + stdlib `re` at broker eval time); the new `{placeholder}` compilation retroactively widens *stored* exact/prefix rules containing `{` — gate or migrate.

No HIGH-severity security findings otherwise — token custody, redirect-URI handling, vendor-secret handling, and the SPA's https-only navigation guards all checked out explicitly.

## Worth fixing in the same pass

- `confirm()` `created`-guard is a plain read → two concurrent confirms both fire vendor `begin` (conditional UPDATE + 409).
- `credentials.state` is documented as broker-enforced (`core/schema/credentials.py:62-66`) but `EffectsRepository.resolve_credentials_for_api` has no state filter — the broker can select the still-`pending`, token-less credential confirm just bound.
- Error responses across `routers/integrations.py`/`vendors.py` are ad-hoc `JSONResponse({"detail": …})` instead of problem-details; the 500 branch returns `str(exc)` verbatim.
- Device-flow Cancel doesn't cancel (`CredentialsPage.tsx:301` only clears local state; the poll loop runs to its 120 s timeout) — thread an `AbortSignal`.
- `poll_interval_seconds` is threaded through the types then ignored (hooks hardcode 3000/1500 ms).
- No poll mutual exclusion across workers (N pods → N× the RFC 8628 rate; `SKIP LOCKED` lease), scanner head-of-line blocking (serial 15 s vendor timeouts × 100 candidates), and no audit entries for connect lifecycle mutations.
- MSW has no handlers for `/integrations:connect` / `/connect-sessions/*`, and the two e2e specs in the PR validate previous PRs' behaviour — nothing automated has ever driven the new wizard end to end, which is how the dead ends in (9) shipped. **[verified live]**: had to tour it by hand because the mocked lane can't render it.

Happy to pair on any of these — the sequencing that seems cheapest: (5)+(6) are mechanical and unblock two CI lanes; (3)+(4)+the confirm TOCTOU are one small service-layer pass; (1)+(2) are one scanner sweep; then the rules-editor/wizard pass with MSW fixtures so it's testable in the fast lane.

Manuel-Jentic (approved):

PR #1387, opened 2026-09-17
fix(auth): admit first-time Google Workspace users on OAuth-client SSO

DavidAtJentic:
## Summary

First-time Google Workspace users signing in via a registered third-party OAuth client hit `access_denied` after approving on the consent page, and no user account was provisioned. The `hd` (hosted-domain) hard gate was rejecting them because `IdpClaims.hosted_domain` was `None`- Google delivers `hd` on the ID token (which we discarded) rather than on `/userinfo`, and the consent-mediated flow additionally dropped `hosted_domain` when it serialized the claims onto the consent handle. The consent-handle round-trip is what made the failure visible only on the OAuth-client path - direct webapp SSO passes the live `IdpClaims` object straight into provisioning, so nothing there depends on the serialized handle carrying the field.

With this PR, first-time Google Workspace users will now be able complete SSO through a registered OAuth client and get a user account provisioned, instead of hitting an error on the consent page.

## Changes

- `auth/core/idp/oidc.py`, `auth/core/idp/adapter.py` - `exchange_code` now returns a `TokenExchangeResult(userinfo, id_token_claims)`. `GoogleOidcAdapter` verifies the ID token against Google's published JWKS (RS256, iss/aud/exp required, keys cached in a process-shared `PyJWKClient`) and reads `hd` from the verified payload, with userinfo as fallback. The generic OIDC adapter has no trust anchor and returns `{}` from the verification hook - no vendor-specific logic in the base.
- `auth/web/flow.py`, `auth/web/routers/authorize.py` - preserve `hosted_domain` on the consent-handle payload and restore it when reconstructing `IdpClaims` at POST `/oauth/consent`, so the OAuth-client flow presents the same claim to the admission policy as the direct-webapp path.

## Risk & rollback

Low risk. Adds a login-time dependency on Google's JWKS endpoint (`https://www.googleapis.com/oauth2/v3/certs`); keys are cached in-process with rotation-aware refetch. Revert = revert the commit.

## Test plan

- `make test-unit` - all 600 auth unit tests pass, including new `test_google_map_claims_reads_hd_from_id_token` and `test_google_map_claims_falls_back_to_userinfo_hd` in `tests/unit/auth/test_idp_adapter.py`.
- `uv run mypy` and `uv run ruff check` clean on all touched files.
- Manual verification pending on QA1: first-time `@jentic.com` SSO via the registered OAuth client should now provision and complete.

## Checklist


ManuelFontRodriguez (dismissed):
(removed — posted from the wrong account)

src/jentic_one/auth/web/routers/authorize.py
```diff
@@ -2206,12 +2206,14 @@ async def consent_submit(
         logger.warning("oauth_consent_missing_claims", client_id=client_id)
         return RedirectResponse(url="/error?error=invalid_consent", status_code=302)
     else:
+        hosted_domain_raw = claims_data.get("hosted_domain")
```
Manuel-Jentic, line 2209: **Blocker:** this fixes the approve arm, but `_claims_from_params` (line 1654) builds the same `IdpClaims` without `hosted_domain`, and the agent-create submit provisions from it (`provision_from_claims(claims)` around line 2468). A first-time Workspace user on an agent-consent-model client who submits the "create your first agent" form still gets `hosted_domain=None` → `hd_hard_gate` → `access_denied`.

Suggest adding `hosted_domain` to `_claims_from_params` and reusing it here instead of the inline copy, so there's one place that rebuilds claims from the handle.

src/jentic_one/auth/web/flow.py
```diff
@@ -508,6 +508,11 @@ async def write_idp_consent_handle(
             "email_verified": claims.email_verified,
             "first_name": claims.first_name,
             "last_name": claims.last_name,
+            # Preserved across the consent round-trip so admission-policy hard
+            # gates (e.g. Google Workspace `hd`) see the same claim they'd see
+            # on the direct callback path — dropping it here rebuilds the
+            # claims with hosted_domain=None and rejects every Workspace user.
+            "hosted_domain": claims.hosted_domain,
```
Manuel-Jentic, line 515: Could we add a test that writes a handle with `hosted_domain="jentic.com"` and asserts it reaches `provision_from_claims` on both the consent-approve and agent-create paths? That round trip is the regression that shipped here.

src/jentic_one/auth/core/idp/oidc.py
```diff
@@ -116,6 +141,16 @@ class GoogleOidcAdapter(OidcAdapter):
...
+    JWKS_URI = "https://www.googleapis.com/oauth2/v3/certs"
+    #: Google accepts both issuers on ID tokens.
+    ACCEPTED_ISSUERS = ("https://accounts.google.com", "accounts.google.com")
+    #: Google signs ID tokens with RS256.
+    SIGNING_ALGORITHMS = ("RS256",)
+
+    #: Shared JWKS client — caches keys in-process, refetches on unknown ``kid``
+    #: (bounded by ``lifespan``) so a Google key rotation is picked up without a
+    #: restart. One instance per process is enough; PyJWKClient is thread-safe.
+    _jwks_client: PyJWKClient = PyJWKClient(JWKS_URI, cache_keys=True, lifespan=300)
```
Manuel-Jentic, line 153: A few concerns with this:
- **Blocking I/O:** `PyJWKClient` uses blocking `urllib.request.urlopen` (default `timeout=30`), and this runs inside async `exchange_code`. On a cold cache, a key rotation, or a slow Google response, the whole event loop stalls, not just this login. Consider `await asyncio.to_thread(self._verify_id_token, id_token)` and/or a short `timeout=`.
- **Class-level singleton:** it's created at import and shared across every adapter instance, so tests can't inject it, and it ignores the endpoint overrides the adapter supports (with a mock IdP, every login logs `google_id_token_verification_failed`). An instance attribute built in `__init__` would avoid both problems.

src/jentic_one/auth/core/idp/oidc.py
```diff
@@ -126,12 +161,38 @@ def _default_token_endpoint(self) -> str:
...
+        """
+        try:
+            signing_key = self._jwks_client.get_signing_key_from_jwt(id_token)
+            claims: dict[str, object] = jwt.decode(
+                id_token,
+                signing_key.key,
+                algorithms=list(self.SIGNING_ALGORITHMS),
+                audience=self._config.client_id,
+                issuer=list(self.ACCEPTED_ISSUERS),
+                options={"require": ["exp", "iss", "aud"]},
```
Manuel-Jentic, line 180: This duplicates `shared/auth/jwt_verification.py::TrustedIssuerVerifier`, which documents itself as the single source for asymmetric JWT verification and additionally does alg-none/HS rejection before key lookup, clock-skew `leeway`, and catches the `OverflowError`/`ValueError` PyJWT raises on absurd numeric claims (not caught here, so that would surface as a 500). Could we build a `TrustedIssuerVerifier` for Google's issuer(s) and reuse it?

Either way, this path needs tests: wrong `aud`, wrong `iss`, expired, bad signature, and a JWKS fetch failure (a locally generated RSA key plus a stubbed JWKS client is enough).

src/jentic_one/auth/core/idp/oidc.py
```diff
@@ -89,10 +100,24 @@ async def exchange_code(self, code: str, *, redirect_uri: str) -> dict[str, obje
                 headers={"Authorization": f"Bearer {access_token}"},
             )
             userinfo_resp.raise_for_status()
-            return userinfo_resp.json()  # type: ignore[no-any-return]
+            userinfo: dict[str, object] = userinfo_resp.json()
+
+        id_token = token_data.get("id_token")
```
Manuel-Jentic, line 105: Nits, now that we parse the ID token: (a) its `nonce` isn't checked against the one sent in `authorize_url`, and (b) there's no check that `id_token_claims["sub"] == userinfo["sub"]`. Both are cheap and low-risk hardening.

src/jentic_one/auth/core/idp/oidc.py
```diff
@@ -126,12 +161,38 @@ def _default_token_endpoint(self) -> str:
...
+        except (jwt.InvalidTokenError, jwt.PyJWKClientError) as exc:
+            logger.warning("google_id_token_verification_failed", error=str(exc))
+            return {}
+
+    def map_claims(self, exchange: TokenExchangeResult) -> IdpClaims:
         """Map Google claims, surfacing the `hd` (hosted-domain) claim.
 
-        `hd` is present only for Google Workspace accounts; it's None for
-        consumer Google accounts. Admission policies use it as a hard gate.
+        Google delivers ``hd`` on the ID token, not (reliably) on userinfo, so
```
Manuel-Jentic, line 190: Two things here:
1. Is this claim confirmed? Before this PR, `hd` was only read from userinfo. If userinfo never had it, first-time Workspace users on the *direct* webapp SSO path would also fail the `hd` gate on qa1/main/eu-main. As far as I know, Google's OIDC userinfo does include `hd` for Workspace accounts. If the consent-handle fix alone resolves QA1, the ID-token part might not be needed in this PR.
2. The userinfo fallback is more than "backwards compatibility with test fixtures": it's the real fallback whenever ID-token verification fails. The docstring should say that.

Manuel-Jentic (changes requested):
Thanks. The root cause (the consent handle serializing claims without `hosted_domain`) is correctly identified, and failing closed on unverifiable ID tokens is the right call. The `TokenExchangeResult` interface change is clean (I found no other `IdpAdapter` implementations in OSS or enterprise), and older handles without the field still load safely through `.get`.

Requesting changes for one blocker plus some test and design follow-ups (details inline):

**Blocker**
1. `_claims_from_params` (`authorize.py:1654`) still rebuilds `IdpClaims` without `hosted_domain`, and the agent-create submit (`:2452` → `provision_from_claims` at `:2468`) provisions from it. A first-time Workspace user on an agent-consent-model client who creates their first agent still hits the enterprise `hd_hard_gate` and gets `access_denied`: the same bug on a second path.

**Should fix**
2. There's no regression test for the handle round trip (`write_idp_consent_handle` → approve / agent-create → `provision_from_claims` carrying `hosted_domain`). That's the bug that shipped, and a test would have caught #1.
3. `_verify_id_token` is untested: nothing checks wrong `aud`/`iss`, an expired token, a bad signature, or a JWKS fetch failure.
4. This re-implements what `shared/auth/jwt_verification.py::TrustedIssuerVerifier` already does, and that module documents itself as the single source for asymmetric JWT verification. It also covers alg-none/HS rejection, leeway, and the `OverflowError`/`ValueError` PyJWT raises on absurd numeric claims. Could we build one for Google's issuer instead?
5. `PyJWKClient` fetches with blocking `urllib` (30s default timeout) inside async `exchange_code`, so a cold cache or a slow Google response stalls the whole worker. Consider `asyncio.to_thread` and/or a short `timeout=`.

**Question: is the ID-token part needed?** Before this PR, `hd` was only read from userinfo. If Google's userinfo never carried `hd`, first-time Workspace users on the *direct* webapp SSO path would also be rejected on every deployment with `hosted_domain: jentic.com`. As far as I know, Google's OIDC userinfo does return `hd` for Workspace accounts. It's worth confirming on QA1 whether the consent-handle fix alone resolves it. If it does, the ID-token verification could be split out (or dropped) and this PR gets much smaller. If we keep it, the PR description's root-cause statement should be adjusted.

**Nits**
- No `sub` cross-check between the ID token and userinfo (cheap, low risk).
- The ID token's `nonce` isn't checked against the one sent in `authorize_url`. Not a regression, but now that we parse the ID token this is a good time to add it.

